PR #7, opened 2015-09-04
add graphics overlay as class member

mbcoder:


mbcoder:
@yferrera a minor fix as it needs to be final or a class member for this to work.

mbcoder:
Closing request as this is no longer needed

PR #11, opened 2015-09-25
MapImageLayerSublayerVisibility

yferrera:
@doneill please review


doneill:
@yferrera The issue on my side is compatibility with ControlsFX and JavaFX 8u60 where private API's that ControlsFX depends on changed in 8u60.  I am getting the sample to work with Java 8u45 so we may need to ensure users do not update to 8u60 to use this dependency.  ControlsFX team says they wil have a new version to support 8u60 in a couple of months.

mara-dolan:
Thought I'd loop in @kathleen1 because this sounds like it could have sys req or doc implications.

doneill:
@mara-dolan I don't think it requires updates to the sys-req doc as this library is not part of the SDK only part of the sample, and potentially future samples, so the only requirement would be on devs using this sample and should be documented with this sample.  

Now I look at this from an individual sample perspective whereby you can use different versions of the JDK for different samples, I am not clear how this would effect a sample viewer app as that will depend on how devs build/compile/run that app.

yferrera:
Thanks @doneill

PR #13, opened 2015-09-29
setting the initial sample to SetInitialMapLocation

doneill:


doneill:
@eric906 Please take a quick review

PR #14, opened 2015-10-09
Editing samples improvement

yferrera:
@mbcoder could you review?


mbcoder:
Tested and this is good to merge.

PR #46, opened 2016-07-06
Dictionary Renderer Sample

john0005:
@tyle8552  Can you give this a look over?

Issue: https://devtopia.esri.com/runtime/java-se-sdk/issues/556
Sample I went off:  https://github.com/Esri/arcgis-runtime-samples-qt/tree/master/ArcGISRuntimeSDKQt_CppSamples/DisplayInformation/GraphicsOverlayDictionaryRenderer


john0005:
I need to do something about this name. There is going to be a dictionary renderer for graphics overlay and feature layer. So I could name it dictionary-renderer-graphicsoverlay.

Other option continuing from my  refeactoring the samples speech. There is already a feature layer category so we can put a dictionary renderer there and if we had a graphics overlay category we could put the other one there.

ghost:
@john0005 I would follow the pattern set by the Qt team. Put this in `display-information`.

ghost:
I noticed, when you zoom out the graphics look bad stacked on top of each other. Is there a setting to change this?

ghost:
@john0005 I added a dependency which helps simplify xml parsing. Even if it's not a standard, I think anything that simplifies a part of the sample which is not directly related to our product can be used. 

I also simplified some other code. We should try to have less `void` methods. For example, you would assume `createGraphic` would create and return a graphic. We shouldn't have to look in that method to see if gets added to the graphics overlay. Making these methods functional allows us to take advantage of streams and pipeline feature/graphic processing.

ghost:
Only things left is to figure out why `mapView.setViewpointGeometryAsync(graphicsOverlay.getExtent()` is not working, find out if labels & graphics can be size-reduced when zoomed out, and move this to `display-information`.

john0005:
For zooming out I just made it so that when you zoom out passed a certain level the graphics disappear.

For extent I found out that you have to wait till the graphics are done drawing to the view.

``` java
mapView.addLayerViewStateChangedListener(layer -> {
      if(layer.getLayerViewStatus().iterator().next() == LayerViewStatus.ACTIVE){
        mapView.setViewpointGeometryAsync(graphicsOverlay.getExtent());
      }
    });
```

PR #58, opened 2016-08-17
Tyle8552/rename

ghost:
Fixed compilation errors from method renames. Updated build version dependency.


vGandhi:
Looks good @tyle8552

PR #63, opened 2016-08-31
added tick marks and label for rotation angle

doneill:
@john0005 @tyle8552 I added tick marks and a label to show the rotation angle. Please review.

### DIFF
--- a/map-view/map-rotation/src/main/java/com/esri/samples/mapview/MapRotation.java
+++ b/map-view/map-rotation/src/main/java/com/esri/samples/mapview/MapRotation.java
@@ -20,7 +20,6 @@ import javafx.application.Application;
 import javafx.geometry.Insets;
 import javafx.geometry.Pos;
 import javafx.scene.Scene;
-import javafx.scene.control.Label;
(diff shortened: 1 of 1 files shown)


john0005:
I feel like the label is not need since it just repeats what the tick marks are displaying. If the label is kept I also feel like it should display 0 when the application first starts because it doesn't display for me until the rotation is changed.

Other than that I really like the addition of the tick marks.

ghost:
@doneill I agree, the label is unnecessary. If you're worried about the labels/ticks not being easily visible, I suggest adding a semi-transparent black background like many of our other samples. See the Update Attributes (Editing) sample's CSS as an example. Other than that, looks good. Thanks.

doneill:
@tyle8552 @john0005 I removed the label.  Only added because all other SDK samples label, the tick marks are a nice effect on the JavaFX slider control, agree that it is redundant with the label.

ghost:
@doneill good to merge.

john0005:
Going to merge this so I can update the dev site today.

PR #74, opened 2016-09-30
Calculate Distance 3D

john0005:
@tyle8552 Can you look this over as well?

issue: https://devtopia.esri.com/runtime/java-se-sdk/issues/855

PR #82, opened 2016-10-12
Calculating Distance 3D for new setup

john0005:
@tyle8552 update this sample for the new build, works on beta-2 but not the build the new setup is set to. 

Original PR: https://github.com/Esri/arcgis-runtime-samples-java/pull/74


ghost:
@john0005 Is this ready for review with the new structure?

john0005:
Yea, one problem that I have come across is that you need to pan or zoom for the graphics to move.

src/main/java/com/esri/samples/scene/calculate_distance_3d/CalculateDistance3dController.java
```diff
@@ -0,0 +1,236 @@
...
+    return convertToCartesianPoint(point1).distance(convertToCartesianPoint(point2));
+  }
+
+  /**
+   * Converts a Point to the Cartesian coordinate system.
+   * 
+   * @param Point point to convert
+   * @return a 3D point in Cartesian coordinates
+   */
+  private Point3D convertToCartesianPoint(Point point) {
```
ghost, line 198: Are we really promoting manual 3d distance calculation? This will be very inaccurate. I think it might make more sense to use the `GeometryEngine` to get the linear distance (should be doing geodetic distance for points on other sides of the US). I believe it gives you the distance in the spatial reference you're using. Then you could `getUnit` and do some unit conversion to get whatever distance metric you want.

PR #87, opened 2016-11-07
added [] to the expression.

mbcoder:
Fixing so the extrusion works again

@tyle8552 or @vGandhi can you review and merge if it's good?


ghost:
@mbcoder We also need to update the readme. I will create a new PR with these changes.

PR #97, opened 2016-11-29
Tyle8552/local server stop

ghost:
Changed local server samples to stop local server with jvm shutdown hook instead of application stop method. 

#### Reasoning
In the context of the sample viewer, when samples stop and switch to the next sample, the timing of the local server stopping was unpredictable, leading to unresponsive samples and exceptions. For example, a sample would start the local server, but the previous sample's async stop operation would immediately stop the local server. 

With the jvm shutdown hook, standalone samples would behave the same way, since the jvm shuts down shortly after the application closes. But for the sample viewer, the local server would not stop between samples--only when the sample viewer closes.

#### Alternative
The alternative is to maintain a separate branch of the samples for the sample viewer in which the local server samples do not contain the stop method. Local server stopping would be handled by the sample viewer.

### DIFF
--- a/src/main/java/com/esri/samples/localserver/local_server_geoprocessing/LocalServerGeoprocessingController.java
+++ b/src/main/java/com/esri/samples/localserver/local_server_geoprocessing/LocalServerGeoprocessingController.java
@@ -166,12 +166,5 @@ public class LocalServerGeoprocessingController {
     if (mapView != null) {
       mapView.dispose();
     }
-
-    // stop local server when jvm shuts down
-    Runtime.getRuntime().addShutdownHook(new Thread(() -> {
-      if (server.getStatus() == LocalServerStatus.STARTED) {
-        server.stopAsync();
-      }
-    }));
   }
 }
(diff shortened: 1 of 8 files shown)


ghost:
@john0005 Can you review? Let me know if you want to discuss.

ghost:
@vGandhi Mentioning you so you can provide some input if you want.

src/main/java/com/esri/samples/localserver/local_server_geoprocessing/LocalServerGeoprocessingController.java
```diff
@@ -162,11 +160,18 @@ protected void handleClearResults() {
...
-   * Stops the animation and disposes of application resources.
+   * Stops and releases all resources used in application.
    */
   void terminate() {
     if (mapView != null) {
       mapView.dispose();
     }
+
+    // stop local server when jvm shuts down
+    Runtime.getRuntime().addShutdownHook(new Thread(() -> {
```
colinanderson: The local server is watching the Java process and should shutdown when the JVM stops so doing this in a shutdown hook shouldn't be necessary.

ghost: @mbcoder Based on this, I think it makes sense to just remove the stop methods from the samples. If local server shuts down when the application closes automatically, we shouldn't tell users to add this redundant code. We have the `LocalServerServices` sample to demonstrate stopping it manually with a button. I think that's probably good enough.

ghost:
After discussing with @vGandhi about Colin's comment, it makes sense to merge this.

PR #104, opened 2016-12-07
Tile Cache Sample

ghost:
Demonstrates loading a tiled layer from a local tpk.


ghost:
@john0005 Could you please review? Not sure about the common design for this one. Mobile platforms are allowing sideloading users' own tpks. They also have a menu for switching between tpks. I'm of the opinion that this should be kept simple. What do you think?

john0005:
I like it the way that you have it. If they know how to load a tpk using a tiled cache, then setting up their own and switching shouldn't be a problem. 

Sample is also good to merge.

PR #122, opened 2017-03-09
Local server Sample to allow packages to be changed

vGandhi:
https://devtopia.esri.com/runtime/java-se-sdk/issues/993

@john0005 Please review.

* Added a button to select a package file from the system.

### DIFF
--- a/src/main/java/com/esri/samples/localserver/local_server_services/LocalServerServicesController.java
+++ b/src/main/java/com/esri/samples/localserver/local_server_services/LocalServerServicesController.java
@@ -41,10 +41,10 @@ public class LocalServerServicesController {
 
   @FXML private Button btnStartServer;
   @FXML private ComboBox<String> serviceOptions;
-  @FXML private Label lblSelectedPackage;
   @FXML private Button btnSelectPackage;
+  @FXML private Label lblSelectedPackage;
   @FXML private TextArea statusLog;
-  @FXML private ListView<String> runningServices;  
+  @FXML private ListView<String> runningServices;
 
   private ListenableList<LocalService> services;
   private HostServices hostServices;
--- a/src/main/java/com/esri/samples/localserver/local_server_services/LocalServerServicesController.java
+++ b/src/main/java/com/esri/samples/localserver/local_server_services/LocalServerServicesController.java
@@ -29,20 +29,18 @@ import com.esri.arcgisruntime.util.ListenableList;
 import javafx.application.HostServices;
 import javafx.application.Platform;
 import javafx.fxml.FXML;
-import javafx.scene.control.Alert;
-import javafx.scene.control.Alert.AlertType;
 import javafx.scene.control.Button;
 import javafx.scene.control.ComboBox;
 import javafx.scene.control.Label;
 import javafx.scene.control.ListView;
 import javafx.scene.control.TextArea;
+import javafx.stage.FileChooser;
 
 public class LocalServerServicesController {
 
   @FXML private Button btnStartServer;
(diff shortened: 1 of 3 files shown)


src/main/java/com/esri/samples/localserver/local_server_services/LocalServerServicesController.java
```diff
@@ -123,6 +135,32 @@ private void handleStartSelectedService() {
...
+        lblSelectedPackage.setText(pathStart +  "MessageInABottle.gpk");
+        break;
+      default:
+        break;
+    }
+  }
+  
+  @FXML
+  private void handleSelectPackage() {
+    File selectedFile = LocalServerServicesSample.openFileDialog();
```
ghost: You can get the stage from the scene like so: 
```java
File selectedFile = new FileChooser().showOpenDialog(btnStartServer.getScene().getWindow());
```
No need for the static method and field on the main class.

src/main/resources/fxml/local_server_services.fxml
```diff
@@ -27,20 +28,24 @@
...
+            <items>
+              <FXCollections fx:factory="observableArrayList">
+                <String fx:value="Map Service"/>
+                <String fx:value="Feature Service"/>
+                <String fx:value="Geoprocessing Service"/>
+              </FXCollections>
+            </items>
+          </ComboBox>
+          <Button fx:id="btnSelectPackage" text="Select package" disable="${!btnStartServer.disable}" onAction="#handleSelectPackage"/>
+          <Label fx:id="lblSelectedPackage" text=""/>
```
ghost: text attribute not needed as it is set to the default value

src/main/java/com/esri/samples/localserver/local_server_services/LocalServerServicesController.java
```diff
@@ -18,19 +18,31 @@
...
+import javafx.scene.control.ComboBox;
+import javafx.scene.control.Label;
+import javafx.scene.control.ListView;
+import javafx.scene.control.TextArea;
 
 public class LocalServerServicesController {
 
   @FXML private Button btnStartServer;
   @FXML private ComboBox<String> serviceOptions;
+  @FXML private Button btnSelectPackage;
```
ghost: I prefer the naming convention: `selectPackageButton`. Most of our samples use this convention.

ghost: Also, you can remove this line for `btnSelectPackage` as it is not used in the code.

ghost (commented):
I left a few suggestions. Also, I know you didn't write this part. But I don't think we need the check to see if the service has already started. Local Server seems fine making copies of the same service and I think we should allow/demonstrate that.

src/main/java/com/esri/samples/localserver/local_server_services/LocalServerServicesController.java
vGandhi: Updated name.

src/main/java/com/esri/samples/localserver/local_server_services/LocalServerServicesController.java
vGandhi: Yep, that's better.

src/main/resources/fxml/local_server_services.fxml
vGandhi: Removed.

vGandhi:
@tyle8552 I have addressed your feedback.

john0005:
Looks good 👍

PR #124, opened 2017-03-22
removed basemap thumbnail resources and changed ui to listview

ghost:
This sample was breaking every time a new default basemap option was introduced. We also have so many options that having a button for each covered most of the map.

I removed the thumbnail resources and changed the UI to use a simple listview.

No need to change this on master since the last release doesn't have all those basemap types.


ghost:
@john0005 Can you review?

john0005:
Looks good. Will also need to update the Picture and how to use section in the README.

PR #128, opened 2017-03-30
changed sample to use services dropdown as filter for selecting packages

ghost:


ghost:
@john0005 can you review this once more? There was some confusion as to what kathleen wanted. Hardik says these changes will work for her.

john0005:
@tyle8552 Image in README is slightly off. 
lssimage

I believe that the open and start button should be switched.

Everything else looks really good.

ghost:
Made the changes. Thanks.

PR #139, opened 2017-05-11
cone symbol position fixes.

mbcoder:
Aligning the code to middle and changing composite distances.

I've also made the cone more pointy!


mbcoder:
@tyle8552 can you review for me?

ghost (approved):
Changes look good. Although, it's kind of hard to tell because it's near impossible to zoom back in on the thing after zooming out. Is this a bug? Seems like the zoom speed has increased.

mbcoder:
@tyle8552 thanks for this. Zooming in is fixed in the latest build so I'll merge this.

PR #145, opened 2017-05-26
Tyle8552/follow graphic camera controller

ghost:
Updated to use orbit camera controller. Removed camera position sliders plus some other small improvements.


ghost:
@mbcoder Assigning you to review since JC will be out for vacation.

mbcoder:
@tyle8552 I'll do a proper code review tomorrow.  Initially this is looking good, but the initial angle of the orbit camera controller would be better if you could see the horizon.  So angled so it is looking more at the back of the aircraft rather than at the top.

src/main/java/com/esri/samples/scene/animate_3d_graphic/Animate3dGraphicController.java
```diff
@@ -115,42 +105,45 @@ public void initialize() {
...
+      // create a graphic with a ModelSceneSymbol of a plane to add to the scene
+      String modelURI = new File("./samples-data/bristol/Collada/Bristol.dae").getAbsolutePath();
+      ModelSceneSymbol plane3DSymbol = new ModelSceneSymbol(modelURI, 1.0);
+      plane3DSymbol.loadAsync();
+      plane3D =  new Graphic(new Point(0, 0, 0, WGS84), plane3DSymbol);
       sceneOverlay.getGraphics().add(plane3D);
 
+      // create an orbit camera controller to follow the plane
+      orbitCameraController = new OrbitGeoElementCameraController(plane3D, 20.0);
+      sceneView.setCameraController(orbitCameraController);
```
mbcoder: Not sure if this is the best code, but the following line makes the horizon visible when the aircraft starts and I think its a better initial view point:

```
orbitCameraController.setCameraPitchOffset(75);
```

build.gradle
```diff
@@ -14,7 +14,7 @@ buildscript {
     }
 }
 
-arcgis.version = '100.1.0-1690'
+arcgis.version = '100.1.0-1715'
```
mbcoder: a moving target, but the latest is 1725

src/main/java/com/esri/samples/scene/animate_3d_graphic/README.md
```diff
@@ -1,25 +1,25 @@
-<h1>Animate 3d Symbols</h1>
+<h1>Animate 3d Graphic</h1>
 
-<p>Demonstrates how to animate a graphic's position and rotation and follow it with a camera. Also shows how to combine
-  a SceneView and MapView in an MVC application with property binding.</p>
+<p>Demonstrates how to animate a graphic's position and rotation and follow using an OrbitGeoElementCameraController. 
+Also shows how to combine a SceneView and MapView in an MVC application with property binding.</p>
```
mbcoder: Could change this to be worded:
```
This application demonstrates the following:
  - How to animate a graphic's position and orientation
  - Following a graphic using an OrbitGeoElementCameraController
  - Synchronising a SceneView and Mapview in an MVC application with a property binding
```

Or something like that which reads better then a sentence with "and and and".

mbcoder (dismissed):
@tyle8552 I like how the camera controller has really simplified this app.  

A couple of minor comments to consider, but once addressed this is good to go.

PR #148, opened 2017-06-09
Alter leader position to TOP for callout.

mbcoder:
Updates needed following callout bug fixes.


colinanderson (commented):
@mbcoder looks ok.

PR #149, opened 2017-06-12
Updating to use public service and moving extent to San Diego 

mbcoder:
Updating to match service used in iOS.


colinanderson:
@mbcoder looks ok.

PR #161, opened 2017-06-28
Rename Open Existing Map -> Open Map URL

john0005:
issue: https://devtopia.esri.com/runtime/java-se-sdk/issues/1071


src/main/java/com/esri/samples/map/open_map_url/OpenMapURLSample.java
```diff
@@ -14,7 +14,7 @@
  * the License.
```
tschie, line 14: Need to update stage title

src/main/java/com/esri/samples/map/open_map_url/README.md
```diff
@@ -2,7 +2,7 @@
```
tschie, line 2: Need to update `h1` tag title

tschie (changes requested):
See comments. Also, I checked the sample in the sample viewer. Either we need to change the sample name to OpenMapUrlSample or I need to make a fix to the sample viewer. It currently formats the name by splitting the class name where it changes from lowercase to uppercase so it's not splitting URL and Sample.

tschie:
I have a fix for allowing for URL in sample viewer: https://devtopia.esri.com/runtime/java-se-sample-viewer/pull/18. This needs to be merged before the sample.

PR #164, opened 2017-07-06
Update sample for best loading practice

john0005:

issue: https://devtopia.esri.com/runtime/java-se-sdk/issues/1039


tschie:
@john0005 Looks good but I'd like to bring this up in samples scrum tomorrow to see if other teams are doing this level of checks (I think it's good) before merging. This also gets into the question of if we should be catching general Exceptions etc.

tschie (approved):

tschie:
@john0005 This is good to go once conflicts are addressed

PR #178, opened 2017-08-04
Offline Route: Listen for task to load before accessing

john0005:
Sample was not working for me and had to wait for task to load to fix problem.


tschie:
Thanks for finding that @john0005. I just added 1 more commit just moving around some of the view logic. Could you review and make sure it works for you now?

tschie (approved):

PR #183, opened 2017-09-12
Tschie/update renderer percentclip

tschie:
PercentClip for both Stretch and RGB renderer samples has a condition that min + max <= 100. Also min/max RGB values from 0-255 should have min < max always. Updated logic and UI based on these changes.

Update the controlsfx version to use a `RangeSlider`. The sample viewer will have to be updated in another PR to allow this.


tschie:
@john0005 Can you please review?

src/main/resources/fxml/rgb_renderer.fxml
```diff
@@ -15,42 +15,46 @@
...
   -->
 
-<?import javafx.scene.control.*?>
+<?import javafx.beans.property.SimpleBooleanProperty?>
+<?import javafx.scene.control.Button?>
+<?import javafx.scene.control.ComboBox?>
+<?import javafx.scene.control.Label?>
+<?import javafx.scene.control.Slider?>
+<?import javafx.scene.control.Spinner?>
 <?import javafx.scene.layout.*?>
```
john0005: Since you got rid of `<?import javafx.scene.control.*?>` want to git rid of `<?import javafx.scene.layout.*?>` and add the following.
```XML
<?import javafx.scene.layout.StackPane?>
<?import javafx.scene.layout.VBox?>
<?import javafx.scene.layout.HBox?>
<?import javafx.scene.layout.GridPane?>
```

src/main/resources/fxml/rgb_renderer.fxml
```diff
@@ -15,42 +15,46 @@
   ~ the License.
   -->
 
-<?import javafx.scene.control.*?>
+<?import javafx.beans.property.SimpleBooleanProperty?>
```
john0005: This is not used.

john0005 (approved):
Some minor things. Looks good.

PR #197, opened 2017-12-01
Update find place to use MVC pattern

tschie:
- Updated to use FXML, more readable without all the UI logic in the code
- Changing to wrong extent issue fixed


tschie:
@john0005 Can you please review?

PR #204, opened 2017-12-20
Initial implementation of WMS layer URL sample

tschie:
WMS layer sample


tschie:
@john0005 Can you please review if you have time?

PR #213, opened 2018-02-06
list transformations by suitability sample

tschie:
Can be pulled into 100.2.1 and 100.3.0 after merging.


tschie:
@john0005 Please review when you have time. Don't worry if it doesn't make it into the next release. We can add it after

john0005 (approved):
Looks good. I added the `How to use sample` part.

PR #216, opened 2018-02-15
Update copyright

john0005:

### DIFF
--- a/src/main/java/com/esri/samples/ogc/wms_layer_url/WmsLayerUrlSample.java
+++ b/src/main/java/com/esri/samples/ogc/wms_layer_url/WmsLayerUrlSample.java
@@ -26,15 +26,11 @@ import javafx.scene.layout.StackPane;
 import javafx.stage.Stage;
 
 import com.esri.arcgisruntime.layers.WmsLayer;
-import com.esri.arcgisruntime.layers.WmtsLayer;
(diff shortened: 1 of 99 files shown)


src/main/java/com/esri/samples/ogc/wms_layer_url/WmsLayerUrlSample.java
```diff
@@ -26,15 +26,11 @@
 import javafx.stage.Stage;
 
 import com.esri.arcgisruntime.layers.WmsLayer;
-import com.esri.arcgisruntime.layers.WmtsLayer;
 import com.esri.arcgisruntime.loadable.LoadStatus;
 import com.esri.arcgisruntime.mapping.ArcGISMap;
 import com.esri.arcgisruntime.mapping.Basemap;
 import com.esri.arcgisruntime.mapping.Viewpoint;
 import com.esri.arcgisruntime.mapping.view.MapView;
-import com.esri.arcgisruntime.ogc.wmts.WmtsLayerInfo;
```
john0005, line 35: These are not needed

tschie (changes requested):
@john0005 We don't need to update all the copyrights. We only need to do that for new samples and ones that have been significantly changed. The formatting fixes are fine.

PR #226, opened 2018-03-16
change sublayer renderer sample

tschie:
shows how to change a sublayers renderer. no offline data required


tschie:
@john0005 please review when you get a chance

john0005 (approved):
This is good.

PR #228, opened 2018-03-20
clip geometry sample

tschie:
simple sample demonstrating clipping a geometry with envelopes. no offline data required


john0005 (approved):
Looks good.

PR #232, opened 2018-04-04
vector tiled layer url sample

tschie:
Shows how to add vector tiled layers as basemaps. Uses 5 different cool vector tiled layers.


john0005:
I like these `Nova` new favorite.

src/main/java/com/esri/samples/tiledlayers/vector_tiled_layer_url/VectorTiledLayerURLSample.java
```diff
@@ -0,0 +1,127 @@
...
+          new ArcGISVectorTiledLayer("http://www.arcgis.com/home/item.html?id=86f556a2d1fd468181855a35e344567f")
+      );
+
+      // load all of the layers so their names can be shown in the list and so they appear as soon as they are selected
+      layerList.getItems().forEach(Layer::loadAsync);
+
+      // set the layer as the map's basemap when selected
+      layerList.getSelectionModel().selectedIndexProperty().addListener(e -> {
+        ArcGISVectorTiledLayer layer = layerList.getSelectionModel().getSelectedItem();
+        map.setBasemap(new Basemap(layer));
```
john0005: I get an `Object already owned` exception here when trying to add a layer that has already been added. May need to create an issue for this. I would believe that setting the basemap would clear out the layer that was previously set. 

Same thing happens with current build.

john0005 (changes requested):
Found an issue. Everything else looks good.

src/main/java/com/esri/samples/tiledlayers/vector_tiled_layer_url/VectorTiledLayerURLSample.java
tschie: @john0005 Good catch. I think the problem is in creating a new basemap with the layer, not setting it; which I think is the correct behavior. A solution is just to copy the selected layer before creating a new basemap. I will make that change.

PR #233, opened 2018-04-07
web tiled layer sample

tschie:
shows how to show non-arcgis service tiled with a web tiled layer. no local data required.


john0005 (approved):
Looks good.

PR #234, opened 2018-04-25
distance measurement analysis sample

tschie:
demonstrates the new distance measurement analysis type


src/main/java/com/esri/samples/analysis/distance_measurement_analysis/DistanceMeasurementAnalysisController.java
```diff
@@ -0,0 +1,169 @@
...
+    sceneView.getAnalysisOverlays().add(analysisOverlay);
+
+    // initialize a distance measurement and add it to the analysis overlay
+    Point start = new Point(-4.494677, 48.384472, 24.772694, SpatialReferences.getWgs84());
+    Point end = new Point(-4.495646, 48.384377, 58.501115, SpatialReferences.getWgs84());
+    distanceMeasurement = new LocationDistanceMeasurement(start, end);
+    analysisOverlay.getAnalyses().add(distanceMeasurement);
+
+    // zoom to the initial measurement
+    sceneView.setViewpointCameraAsync(new Camera(start, 200.0, 0.0, 45.0, 0.0));
```
john0005: i would just set this to `setViewpointCamera` so that when the application starts it already at the location and doesn't do that slow zoom in.

john0005 (approved):
Looks good. Minor suggestion.

PR #238, opened 2018-05-09
Map rotation compass

tschie:
Adds the toolkit as a dependency. Shows the Compass component in the Map Rotation sample.


src/main/java/com/esri/samples/mapview/map_rotation/README.md
```diff
@@ -1,27 +1,29 @@
...
 <h2>How it works</h2>
 
-<p>To rotate an <code>ArcGISMap</code>:</p>
+<p>To rotate a map:</p>
 
 <ol>
-    <li>Create an ArcGIS map. </li>
-    <li>Add the map to the <code>MapView</code>, <code>MapView.setMap()</code>. </li>
+    <li>Create an <code>ArcGISMap</code>.</li>
+    <li>Add the map to a <code>MapView</code></code>.</li>
```
john0005: Extra `<code>`.

john0005 (changes requested):
Code in place looks good! One thing though if you click the compass it will set the MapView back to north. This however does set the slider back to zero.

PR #259, opened 2018-08-01
change selection UI to a combo box

tschie:
use a combo box instead of two buttons to more clearly show which expression is currently being used


tschie:
@john0005 please review

john0005 (approved):
Good!

PR #276, opened 2018-09-24
display kml network links sample

tschie:
shows how to display KML that automatically updates and shows network messages


tschie:
I've asked why we only see two messages for this data. Will update when I find out

colinanderson:
I think you should only see one message with that data.

tschie:
Just re-ran this with the latest build and now I only see one message

src/main/java/com/esri/samples/ogc/display_kml_network_links/DisplayKMLNetworkLinksSample.java
```diff
@@ -0,0 +1,112 @@
...
+
+      // show an alert when any network link messages are received
+      kmlDataset.addKmlNetworkLinkMessageReceivedListener(kmlNetworkLinkMessageReceivedEvent -> {
+        Alert alert = new Alert(Alert.AlertType.INFORMATION, kmlNetworkLinkMessageReceivedEvent.getMessage());
+        alert.setHeaderText("KML Network Link Message");
+        alert.initOwner(sceneView.getScene().getWindow());
+        alert.show();
+      });
+
+      // add the KML layer as an operational layer and zoom to it after it is loaded
```
john0005: Would say, // add the KML layer as an operational layer and check if loaded correctly
This part has nothing to do with setting zoom

src/main/java/com/esri/samples/ogc/display_kml_network_links/README.md
```diff
@@ -0,0 +1,35 @@
...
+<h2>How to use the sample</h2>
+
+<p>The data shown should refresh automatically every few seconds. A network message will be displayed when the dataset is loaded.</p>
+
+<h2>How it works</h2>
+
+<p>To show KML with network links and display network messages:</p>
+
+<ol>
+    <li>Create a <code>KmlNetworkDataset</code> from a KML source which has network links.</li>
```
john0005: This should just be KmlDataset.

john0005 (changes requested):
A minor change.

PR #277, opened 2018-09-24
identify kml features sample

tschie:
shows how to identify clicked kml features and show their balloon content in a callout


john0005 (approved):
Looks good!

PR #278, opened 2018-09-24
list kml contents sample

tschie:
demonstrates how to populate a tree view showing the hierarchy of kml nodes.

I've added the local kml file to samples-data


john0005 (approved):
Looks good.

PR #283, opened 2018-11-05
Update ScaleBarSample

john0005:
Updating the format spacing of this sample.


john0005:
@tschie Can you look this over? Need change the spacing format to be like the other samples so that it will display properly on dev site.

tschie (approved):

PR #289, opened 2018-11-26
Java 11

tschie:
Updates to run samples with Java 11 / OpenJFX 11


tschie:
@Rachael-E Please review and check if you can run the samples with Java 11. You shouldn't have to download the JavaFX SDK. These buildscript changes should handle all of that.

tschie:
Merging to fix the sample viewer build

PR #292, opened 2018-12-07
Generate offline map overrides

tschie:
Shows how to override the default offline map parameters.


src/main/java/com/esri/samples/map/generate_offline_map_overrides/README.md
```diff
@@ -0,0 +1,129 @@
...
+<li>take features with no geometry offline.</li>
+</ul>
+
+<p><img src="GenerateOfflineMapOverrides.png"/></p>
+
+<h2>How to use the sample</h2>
+
+<p>Sign in with a free developer account when prompted for credentials (taking web maps offline requires an account).</p>
+
+<p>Use the min/max scale spinners to adjust the levelIds to be taken offline for the Streets basemap.</p>
```
Rachael-E: Can we write 'levelIds' as 'level IDs'? On markdown it looks like a typo and isn't clear at first it means level IDs. Also wondered if there is an API link that could be included to link to what level IDs are, or if a short description is available, could be included here?

src/main/java/com/esri/samples/map/generate_offline_map_overrides/GenerateOfflineMapOverridesController.java
```diff
@@ -0,0 +1,282 @@
...
+            try {
+              GenerateOfflineMapParameterOverrides overrides = parameterOverridesFuture.get();
+
+              // get the export tile cache parameters for the base layer
+              OfflineMapParametersKey basemapParamKey = new OfflineMapParametersKey(
+                  mapView.getMap().getBasemap().getBaseLayers().get(0));
+              ExportTileCacheParameters exportTileCacheParameters =
+                  overrides.getExportTileCacheParameters().get(basemapParamKey);
+
+              // create a new sublist of LODs in the range requested by the user
```
Rachael-E: What are LODs - should this be level IDs?

src/main/java/com/esri/samples/map/generate_offline_map_overrides/README.md
```diff
@@ -0,0 +1,129 @@
...
+
+<h1>How it works</h1>
+
+<ol>
+<li>Load a web map from a <code>PortalItem</code>. Authenticate with the portal if required.</li>
+
+<li>Create an <code>OfflineMapTask</code> with the map.</li>
+
+<li>Generate default task parameters using the extent area you want to download with <code>offlineMapTask.
+.createDefaultGenerateOfflineMapParametersAsync(extent)</code>.</li>
```
Rachael-E: Typo: remove duplicate period here

src/main/java/com/esri/samples/map/generate_offline_map_overrides/README.md
```diff
@@ -0,0 +1,129 @@
...
+
+<li>Generate additional "override" parameters using the default parameters with <code>offlineMapTask
+.createGenerateOfflineMapParameterOverridesAsync(parameters)</code>.</li>
+
+<li>For the basemap:
+
+
+<ul>
+<li>Get the parameters <code>OfflineMapParametersKey</code> for the basemap layer`.</li>
```
Rachael-E: Typo: remove stray back tick at the end of this sentence

src/main/java/com/esri/samples/map/generate_offline_map_overrides/README.md
```diff
@@ -0,0 +1,129 @@
...
+</li>
+
+<li>To filter the features downloaded in an operational layer:
+
+
+<ul>
+<li>Get the layer options for the operational layer using the directions in step 6.</li>
+
+<li>Loop through the layer options. If the option layerID matches the layer's ID, set the filter clause with 
+<code>layerOption.setWherClause(sqlQueryString)</code> and set the query option with <code>layerOption.setQueryOption
```
Rachael-E: typo: should be "setWhereClause" not "setWherClause"

also there is an extra space after .setQueryOption which can be removed.

src/main/java/com/esri/samples/map/generate_offline_map_overrides/GenerateOfflineMapOverridesController.java
```diff
@@ -0,0 +1,282 @@
...
+  @FXML
+  private void initialize() {
+    // handle authentication with the portal
+    AuthenticationManager.setAuthenticationChallengeHandler(new DefaultAuthenticationChallengeHandler());
+
+    // create a portal item with the itemId of the web map
+    Portal portal = new Portal("https://www.arcgis.com", true);
+    PortalItem portalItem = new PortalItem(portal, "acc027394bc84c2fb04d1ed317aac674");
+
+    // create a graphics overlay for display the download area
```
Rachael-E: typo: should be "create a graphics overlay for displaying the download area"

src/main/java/com/esri/samples/map/generate_offline_map_overrides/GenerateOfflineMapOverridesController.java
```diff
@@ -0,0 +1,282 @@
...
+                  // replace the current map with the result offline map when the job finishes
+                  GenerateOfflineMapResult result = job.getResult();
+                  mapView.setMap(result.getOfflineMap());
+                  graphicsOverlay.getGraphics().clear();
+                  generateOfflineMapButton.setDisable(true);
+                } else if (!"Job canceled.".equals(job.getError().getAdditionalMessage())) {
+                  new Alert(Alert.AlertType.ERROR, job.getError().getAdditionalMessage()).show();
+                }
+                Platform.runLater(() -> progressBar.setVisible(false));
+                generateOfflineMapButton.setDisable(false);
```
Rachael-E: The first time Generate offline map is clicked, the progress bar loads and the saved offline data are displayed on the screen. As a user, I see the Generate offline map button is enabled again and assume it would create a subset of offline data based on the current view. However, clicking the button results in a null pointer exception at line 153. 

Could disable the Generate Offline Button at this point so that the user isn't tempted to generate a secondary offline map as the UI infers, or refactor the code to allow a re-generation of offline data?

src/main/java/com/esri/samples/map/generate_offline_map_overrides/GenerateOfflineMapOverridesController.java
```diff
@@ -0,0 +1,282 @@
...
+    Portal portal = new Portal("https://www.arcgis.com", true);
+    PortalItem portalItem = new PortalItem(portal, "acc027394bc84c2fb04d1ed317aac674");
+
+    // create a graphics overlay for display the download area
+    graphicsOverlay = new GraphicsOverlay();
+    mapView.getGraphicsOverlays().add(graphicsOverlay);
+
+    // show a red border around the download area
+    downloadArea = new Graphic();
+    graphicsOverlay.getGraphics().add(downloadArea);
```
Rachael-E: The download area graphic doesn't always draw immediately after user logs in: leading to error when clicking on generate offline map button. If user pans the map slightly the graphic draws fine though.

src/main/java/com/esri/samples/map/generate_offline_map_overrides/GenerateOfflineMapOverridesController.java
```diff
@@ -0,0 +1,282 @@
...
+                  overrides.getExportTileCacheParameters().get(basemapParamKey);
+
+              // create a new sublist of LODs in the range requested by the user
+              exportTileCacheParameters.getLevelIDs().clear();
+              for (int i = minScaleLevelSpinner.getValue(); i < maxScaleLevelSpinner.getValue(); i++) {
+                exportTileCacheParameters.getLevelIDs().add(i);
+              }
+              // set the area of interest to the original download area plus a buffer
+              exportTileCacheParameters.setAreaOfInterest(GeometryEngine.buffer(downloadArea.getGeometry(),
+                  extentBufferDistanceSpinner.getValue()));
```
Rachael-E, line 160: When I set the spinner to max (500 as specified in the fxml file), the offline map created still matches exactly the geometry of the download area. I thought the buffer would give an extra 500m of basemap coverage/offline data outside of the download area: might be misunderstanding what this functionality is for, or could it be that it isn't handling as expected?

src/main/java/com/esri/samples/map/generate_offline_map_overrides/GenerateOfflineMapOverridesController.java
```diff
@@ -0,0 +1,282 @@
...
+import com.esri.arcgisruntime.tasks.offlinemap.OfflineMapTask;
+import com.esri.arcgisruntime.tasks.tilecache.ExportTileCacheParameters;
+
+public class GenerateOfflineMapOverridesController {
+
+  @FXML private MapView mapView;
+  @FXML private Spinner<Integer> minScaleLevelSpinner;
+  @FXML private Spinner<Integer> maxScaleLevelSpinner;
+  @FXML private Spinner<Integer> extentBufferDistanceSpinner;
+  @FXML private Spinner<Integer> minHydrantFlowRateSpinner;
```
Rachael-E, line 68: For these spinners, it would be useful from a UX perspective to warn user if the value they've inputted is outside the max value range as defined in the FXML

Rachael-E (commented):
@tschie Looking good! I encountered errors/exceptions when interacting with the sample which I've commented on, and made general typo corrections/suggestions.

src/main/java/com/esri/samples/map/generate_offline_map_overrides/GenerateOfflineMapOverridesController.java
tschie: @Rachael-E so JavaFX spinners are a little weird because by default, you have to hit Enter after typing in a value to commit it. Not sure if you were doing that. I will add a note to the readme to warn users.

src/main/java/com/esri/samples/map/generate_offline_map_overrides/GenerateOfflineMapOverridesController.java
tschie: By default, when you commit the value or hit the up or down arrows, it will clamp the value to the allowed range.

src/main/java/com/esri/samples/map/generate_offline_map_overrides/GenerateOfflineMapOverridesController.java
tschie: Level of detail, synonymous with level IDs. But you're right, they should be consistent.

src/main/java/com/esri/samples/map/generate_offline_map_overrides/GenerateOfflineMapOverridesController.java
tschie: We could use a slider instead. But I felt like many users would want to directly input the value. What do you think?

tschie:
@Rachael-E I think it should all be good now. Can you please give it another look?

src/main/java/com/esri/samples/map/generate_offline_map_overrides/GenerateOfflineMapOverridesController.java
Rachael-E: @tschie I think the spinner is ok, now that there is the note in the readme to hit enter after the value has been entered. I would make that note bold though so it catches the readers eye when looking through it to understand the sample.

src/main/java/com/esri/samples/map/generate_offline_map_overrides/GenerateOfflineMapOverridesController.java
```diff
@@ -110,6 +107,9 @@ private void initialize() {
       updateDownloadArea();
```
Rachael-E: The download area graphic is still not always appearing after log in: I've had a play around with the code too but can't figure out a way to make it always appear. Happens once every 3-4 attempts.

src/main/java/com/esri/samples/map/generate_offline_map_overrides/README.md
```diff
@@ -8,22 +8,21 @@ sample. For more fine-grained control over the data you are taking offline, use
...
+    <li>reduce the amount of data (such as tiles) downloaded for one or more layers in the map.</li>
+    <li>filter features to be taken offline.</li>
+    <li>take features with no geometry offline.</li>
 </ul>
 
 <p><img src="GenerateOfflineMapOverrides.png"/></p>
 
 <h2>How to use the sample</h2>
 
+<p>Note: When manually inputting a value in the spinner controls, make sure to hit Enter to commit the value.</p>
```
Rachael-E: This is really helpful: would suggest making it bold to catch the readers eye.

Rachael-E (approved):
@tschie just a note about the readme and the download area issue to address and this is good to go!

src/main/java/com/esri/samples/map/generate_offline_map_overrides/GenerateOfflineMapOverridesController.java
tschie: Yeah, it was very hard to repro for me. Seemed like a timing issue. I only saw it happen a couple times. I think I have it sorted out now.

PR #305, opened 2019-01-21
Choose camera controller

Rachael-E:
@tschie here's the refactored sample which simply shows the difference between the `GlobeCameraController` and the `OrbitGeoElementCameraController`. Please can you review? Thanks!


Rachael-E:
@tschie I've updated this sample: it now orbits an aeroplane/location with the globe camera option. This is the latest version I'm happy with after changing the orange diamond: please can you review? Thanks!

src/main/java/com/esri/samples/scene/choose_camera_controller/ChooseCameraControllerSample.java
```diff
@@ -0,0 +1,200 @@
...
+
+    try {
+
+      // create stack pane and JavaFX app scene
+      StackPane stackPane = new StackPane();
+      Scene fxScene = new Scene(stackPane);
+      fxScene.getStylesheets().add(getClass().getResource("/css/style.css").toExternalForm());
+
+      // set title, size, and add JavaFX scene to stage
+      stage.setTitle("Change Atmosphere Effect Sample");
```
tschie: ;)

src/main/java/com/esri/samples/scene/display_scene/DisplaySceneSample.java
```diff
@@ -67,6 +67,11 @@ public void start(Stage stage) {
       // add a camera and initial camera position
       Camera camera = new Camera(28.4, 83.9, 10010.0, 10.0, 80.0, 0.0);
       sceneView.setViewpointCamera(camera);
+
```
tschie, line 70: This should be removed

src/main/java/com/esri/samples/scene/choose_camera_controller/ChooseCameraControllerSample.java
```diff
@@ -0,0 +1,200 @@
...
+      orbitCameraController.setCameraHeadingOffset(150);
+
+      // instantiate a new camera controller which orbits a target location
+      Point locationPoint = new Point(-109.929589, 38.437304, 3500, SpatialReferences.getWgs84());
+      orbitLocationCameraController = new OrbitLocationCameraController(locationPoint, 10);
+      orbitLocationCameraController.setCameraPitchOffset(3);
+      orbitLocationCameraController.setCameraHeadingOffset(150);
+
+      // instantiate a new label to display which camera controller is active
+      Label cameraModeLabel = new Label("MOVE FREELY ROUND THE SCENE");
```
tschie: I don't think this is necessary. The globe camera controller button could just say "Freely pan around the globe" or something like that.

src/main/java/com/esri/samples/scene/choose_camera_controller/ChooseCameraControllerSample.java
```diff
@@ -0,0 +1,200 @@
...
+      cameraModeLabel.setMaxWidth(Double.MAX_VALUE);
+      cameraModeLabel.setAlignment(Pos.CENTER);
+
+      // instantiate control buttons to choose what camera controller is active
+      Button orbitCameraControllerButton = new Button("Orbit Aeroplane Camera Controller");
```
tschie: I think this can just be Orbit Aeroplane. I prefer gearing the UI towards the end-user rather than the developer. A user wouldn't need to know they are using a camera controller. The developer has the source code, so they can see how in order to make a button with orbiting behavior, they have to create a camera controller.

src/main/java/com/esri/samples/scene/choose_camera_controller/ChooseCameraControllerSample.java
tschie: I also think this might suggest using a different UI. Maybe a radio button group or a combobox so the currently selected mode is obvious.

src/main/java/com/esri/samples/scene/choose_camera_controller/ChooseCameraControllerSample.java
```diff
@@ -0,0 +1,200 @@
...
+import javafx.scene.layout.StackPane;
+import javafx.scene.layout.VBox;
+import javafx.scene.paint.Paint;
+import javafx.stage.Stage;
+
+import java.io.File;
+
+public class ChooseCameraControllerSample extends Application {
+
+  private OrbitGeoElementCameraController orbitCameraController;
```
tschie: Do these need to be fields?

tschie (changes requested):
Looks good! But I do have a few suggestions.

src/main/java/com/esri/samples/scene/choose_camera_controller/ChooseCameraControllerSample.java
Rachael-E: oops...

src/main/java/com/esri/samples/scene/display_scene/DisplaySceneSample.java
Rachael-E: good catch: didn't spot I'd accidentally done that to another file :(

src/main/java/com/esri/samples/scene/choose_camera_controller/ChooseCameraControllerSample.java
Rachael-E: I agree....will update the UI with radio button or combobox.

Rachael-E:
@tschie thanks for the review and picking up some mistakes! I've updated the UI with radio buttons so it's a bit cleaner. Can you take another look please? Updated the screenshot and README as well.

src/main/java/com/esri/samples/scene/choose_camera_controller/ChooseCameraControllerSample.java
```diff
@@ -0,0 +1,192 @@
...
+      sceneView.setViewpointCamera(camera);
+
+      // instantiate a new camera controller which orbits the aeroplane at a set distance
+      OrbitGeoElementCameraController orbitAeroplaneCameraController = new OrbitGeoElementCameraController(plane3D, 100.0);
+      orbitAeroplaneCameraController.setCameraPitchOffset(30);
+      orbitAeroplaneCameraController.setCameraHeadingOffset(150);
+
+      // instantiate a new camera controller which orbits a target location
+      Point locationPoint = new Point(-109.929589, 38.437304, 3500, SpatialReferences.getWgs84());
+      OrbitLocationCameraController orbitLocationCameraController = new OrbitLocationCameraController(locationPoint, 10);
```
tschie: This is too close. The location should be on the ground and the camera should be much farther away if the idea is to orbit around that basin.

src/main/java/com/esri/samples/scene/choose_camera_controller/ChooseCameraControllerSample.java
```diff
@@ -0,0 +1,192 @@
...
+      globeCamera.setSelected(true);
+
+      // set the buttons to a toggle group
+      ToggleGroup toggleGroup = new ToggleGroup();
+      orbitAeroplane.setToggleGroup(toggleGroup);
+      orbitLocation.setToggleGroup(toggleGroup);
+      globeCamera.setToggleGroup(toggleGroup);
+
+      // set the radio buttons to choose which camera controller is active
+      orbitAeroplane.setOnAction(event -> {
```
tschie: you can omit the curly braces

tschie (approved):
Looks good! Just a couple more things I missed before. Good to merge after those are fixed.

PR #306, opened 2019-01-23
Add an integrated mesh layer

tschie:
Demonstrates the new `IntegratedMeshLayer` type.


src/main/java/com/esri/samples/scene/add_an_integrated_mesh_layer/README.md
```diff
@@ -0,0 +1,32 @@
+<h1>Add an integrated mesh layer</h1>
+
+<p>View an integrated mesh layer from a scene service.</p>
+
+<p><img src="AddAnIntegratedMeshLayer.png"/></p>
+
+<h2>Use Case</h2>
+
+<p>An integrated mesh can represent artificial and natural 3D features such as
```
Rachael-E, line 9: Having now seen the sample, I think it would be good to explain a little more what the integrated mesh actually is when compared with the elevation source layer.  It's not instantly obvious to a new user that it contains more detail than the standard elevation source, and the difference becomes clearer whilst panning to the margins of the integrated mesh layer. Suggest explicitly pointing out that difference, maybe in a "How to use this sample" section to encourage the user to compare and contrast the mesh with elevation source layer. 

"An `IntegratedMeshLayer` typically provides more detail than an `ArcGISElevationSourceLayer`. For example, it may have higher resolution imagery/textures (seen in this sample on the cliff face in Yosemite) and detailed elevation data. To appreciate the latter, navigate to the edge of the integrated mesh layer in this sample and compare tree elevations with the standard elevation source layer."

Rachael-E (commented):
This all looks great @tschie! Code is good, no changes required there. I've suggested an expansion to the README to make the difference between an integrated mesh and elevation source clearer. If you're happy with this then we'll need to update the design for this sample too.

src/main/java/com/esri/samples/scene/add_an_integrated_mesh_layer/README.md
tschie: @Rachael-E Could you open a PR for a design readme addition so others on the team can add their feedback?

I have heard that some products (not Runtime yet) can use the data of an integrated mesh _as_ an elevation source for a surface. So I'm not sure they are meant to compete. I'm also wondering if pointing out the contrast in elevation detail here only highlights how low density the world elevation source is (as it should be at world scale). A local elevation source might have comparable detail (lacking trees, since ground elevation sources should ignore those). Lastly, the elevation accuracy is only as good as the imagery, radar/lidar data, and stitching algorithm that formed the mesh. You could have an integrated mesh with only 4 images, whose elevation data would be worse than the world elevation source. You're right that you will typically find meshes with higher resolution elevation, but I don't know if that is an intrinsic property of meshes.

PR #315, opened 2019-02-11
View point cloud data offline

tschie:
shows how to view a point cloud layer from an offline slpk


src/main/java/com/esri/samples/scene/view_point_cloud_data_offline/README.md
```diff
@@ -0,0 +1,33 @@
+<h1>View point cloud data offline</h1>
+
+<p>Display local 3D point cloud data.</p>
+
+<p><img src="ViewPointCloudDataOffline.png"/></p>
+
+<h2>Use case</h2>
+
+<p>Point clouds are often used to visualize massive sets of sensor data such as lidar. The point locations indicate where the sensor data was measured spatially, and the color or size of the points indicate the measured/derived value of the sensor reading. In the case of lidar, the color of the visualized point could be the color of the reflected light, so that the point cloud forms a true color 3D image of the area.</p>
```
Rachael-E, line 9: There should be a mention of what the colours are in the app: ideally this would be in a How to use the Sample Section, though in the absence of that (since all the user has to do is open it), I'd describe the colours in Use case/ or About the data section. 

e.g. after the section about true colour, something like this: "The color of the visualized point could also represent the elevation value of the data. In this sample, the points are colored for elevation on a spectrum from red (high elevation) to blue (low elevation).

src/main/java/com/esri/samples/scene/view_point_cloud_data_offline/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+<p>Point clouds can be loaded offline from scene layer packages (.slpk).</p>
```
Rachael-E: I think this could be better in additional information section: doesn't fit well in the use case section.

src/main/java/com/esri/samples/scene/view_point_cloud_data_offline/ViewPointCloudDataOfflineSample.java
```diff
@@ -0,0 +1,111 @@
...
+      ArcGISScene scene = new ArcGISScene();
+      scene.setBasemap(Basemap.createImagery());
+      sceneView.setArcGISScene(scene);
+
+      // set the base surface with world elevation
+      Surface surface = new Surface();
+      surface.getElevationSources().add(new ArcGISTiledElevationSource("http://elevation3d.arcgis.com/arcgis/rest/services/WorldElevation3D/Terrain3D/ImageServer"));
+      scene.setBaseSurface(surface);
+
+      // add a a point cloud layer with data from Balboa Park in San Diego
```
Rachael-E: typo: remove additional 'a'.

Would add in that this data is a scene layer package: 

"add a point cloud layer with scene layer package data from Balboa Park in San Diego"

Rachael-E (commented):
This looks good @tschie! Just a few comments and typo to address.

PR #319, opened 2019-02-20
Map reference scale

Rachael-E:
@tschie this is for the map reference scale sample: please can you take you look? Thanks!


src/main/java/com/esri/samples/map/map_reference_scale/MapReferenceScaleController.java
```diff
@@ -0,0 +1,139 @@
...
+    map.setReferenceScale(250000);
+    scaleLabel.setText("LOADING...");
+    loadingLabel.setText("LOADING...");
+
+    // set up the combobox to have 1:50k, 100k, 250k and 500k reference scales
+    scaleComboBox.getItems().addAll("1:50000", "1:100000", "1:250000", "1:500000");
+    scaleComboBox.getSelectionModel().select(2);
+
+    // create a label to display current scale of the map
+    mapView.addMapScaleChangedListener(event -> {
```
tschie: can be simplified to
```java
mapView.addMapScaleChangedListener(event -> 
    scaleLabel.setText("Current Map Scale 1:" + Math.round(mapView.getMapScale()))
);
```

src/main/java/com/esri/samples/map/map_reference_scale/MapReferenceScaleController.java
```diff
@@ -0,0 +1,139 @@
...
+      setScaleSymbolsOnSelectedLayers();
+      loadingLabel.setText("Apply Reference Scale");
+    });
+  }
+
+  /**
+   * Sets the scale symbols (i.e. if the layer will honor the reference scale or not)
+   * for each layer within the map
+   */
+  private void setScaleSymbolsOnSelectedLayers() {
```
tschie: This method should only be called once. You could just move the logic into the `addDoneLoadingListener` lambda above.

src/main/java/com/esri/samples/map/map_reference_scale/MapReferenceScaleController.java
```diff
@@ -0,0 +1,139 @@
...
+    Portal portal = new Portal("http://runtime.maps.arcgis.com");
+    PortalItem portalItem = new PortalItem(portal, "3953413f3bd34e53a42bf70f2937a408");
+
+    // create a map with the portal item
+    map = new ArcGISMap(portalItem);
+
+    // set the map to the map view
+    mapView.setMap(map);
+    map.setReferenceScale(250000);
+    scaleLabel.setText("LOADING...");
```
tschie: You could alternatively hide the vboxes at the start and show them when the map is done loading. I would also put a spinner at the center of the map to indicate it is loading. With "loading" labels, it's unclear what is loading. With a spinner over the map, it's clearer.

src/main/java/com/esri/samples/map/map_reference_scale/MapReferenceScaleController.java
```diff
@@ -0,0 +1,139 @@
...
+    loadingLabel.setText("LOADING...");
+
+    // set up the combobox to have 1:50k, 100k, 250k and 500k reference scales
+    scaleComboBox.getItems().addAll("1:50000", "1:100000", "1:250000", "1:500000");
```
tschie: You can do this in the FXML:
```xml
<ComboBox fx:id="scaleComboBox" maxWidth="Infinity" value="1:250000">
   <items>
        <FXCollections fx:factory="observableArrayList">
             <String fx:value="1:50000"/>
             <String fx:value="1:100000"/>
             <String fx:value="1:250000"/>
             <String fx:value="1:500000"/>
        </FXCollections>
   </items>
</ComboBox>
```

src/main/java/com/esri/samples/map/map_reference_scale/MapReferenceScaleController.java
```diff
@@ -0,0 +1,139 @@
...
+  private void setScaleSymbolsOnSelectedLayers() {
+
+    LayerList operationalLayers = map.getOperationalLayers();
+
+    for (Layer layer : operationalLayers) {
+      // create a checkbox per operational layer name
+      CheckBox checkBox = new CheckBox(layer.getName());
+      checkBox.setSelected(true);
+      // add the checkboxes to the VBox
+      vBox.getChildren().add(checkBox);
```
tschie: This is fine for this sample. For reference though, you could also use a list view with a check box list cell type like in this sample: https://github.com/Esri/arcgis-runtime-samples-java/blob/master/src/main/java/com/esri/samples/featurelayers/statistical_query_group_and_sort/StatisticalQueryGroupAndSortController.java#L98.

That would come in handy if you ever wanted to add/remove layers or if you had a lot of layers and needed to scroll.

src/main/java/com/esri/samples/map/map_reference_scale/MapReferenceScaleController.java
```diff
@@ -0,0 +1,139 @@
...
+
+      FeatureLayer featureLayer = (FeatureLayer) layer;
```
tschie: Should probably check if `layer instanceof FeatureLayer` before creating a check box.

src/main/java/com/esri/samples/map/map_reference_scale/MapReferenceScaleController.java
```diff
@@ -0,0 +1,139 @@
...
+        featureLayer.setScaleSymbols(checkBox.isSelected());
+      });
+    }
+  }
+
+  /**
+   * Takes the reference scale from the combobox, and sets it as the map's reference scale.
+   */
+  @FXML
+  private void handleScaleButtonClicked() {
```
tschie, line 128: I don't seem to be able to select a new reference scale without also setting the map scale. Seems like they should be independent.

I would think there would be two handlers: one when the combo box selected item changes to set the reference scale, and another when the button is clicked to set the map scale to the reference scale.

Is that not how the other SDKs are doing it?

src/main/java/com/esri/samples/map/map_reference_scale/MapReferenceScaleController.java
```diff
@@ -0,0 +1,139 @@
...
+
+    // get string value from the combobox, convert to double
+    String selectedString = scaleComboBox.getSelectionModel().getSelectedItem();
+    String[] splitString = selectedString.split(":");
```
tschie: Instead of doing this, you could give the combo box a `Double` type and use a string converter for the display:

```java
@FXML
private ComboBox<Double> scaleComboBox;

//...

scaleComboBox.setConverter(new StringConverter<Double>() {
      @Override
      public String toString(Double value) {
        return "1:" + Math.round(value);
      }

      @Override
      public Double fromString(String string) {
        // not needed
        return null;
      }
});
```

If you used my suggestion above about the FXML, you'd need to update that too:
```xml
<ComboBox fx:id="scaleComboBox" maxWidth="Infinity">
      <value>
           <Double fx:value="250000"/>
      </value>
      <items>
            <FXCollections fx:factory="observableArrayList">
                    <Double fx:value="50000"/>
                    <Double fx:value="100000"/>
                    <Double fx:value="250000"/>
                    <Double fx:value="500000"/>
           </FXCollections>
      </items>
</ComboBox>
```

src/main/java/com/esri/samples/map/map_reference_scale/README.md
```diff
@@ -0,0 +1,45 @@
...
+
+<h2>Relevant API</h2>
+
+<ul>
+<li>ArcGISMap</li>
+
+<li>FeatureLayer</li>
+</ul>
+
+<h2 id="additionalinformation">Additional Information</h2>
```
tschie: You can remove the id attribute

src/main/java/com/esri/samples/map/map_reference_scale/README.md
```diff
@@ -0,0 +1,45 @@
...
+
+<p>The map reference scale should normally be set by the map's author and not exposed to the end user like it is in this sample. </p>
+
+<h4>Tags</h4>
+
+<p>Maps &amp; Scenes, reference scale</p>
```
tschie: I think that symbol would be fine as `&`

tschie (approved):
Looks good! Just a few optional suggestions

src/main/java/com/esri/samples/map/map_reference_scale/MapReferenceScaleController.java
tschie: Seem my suggestion below if you want to use `Double`s

src/main/java/com/esri/samples/map/map_reference_scale/MapReferenceScaleController.java
Rachael-E: Good catch, I'd missed that from the earlier design: just checked with Trevor and the behaviour should be separated. Working on it!

Rachael-E:
@tschie thanks for the review, lots of really helpful pointers there! I've addressed all of your comments: please could you take another look?

src/main/java/com/esri/samples/map/map_reference_scale/MapReferenceScaleController.java
```diff
@@ -0,0 +1,162 @@
...
+import javafx.util.StringConverter;
+
+public class MapReferenceScaleController {
+
+  @FXML
+  private MapView mapView;
+  @FXML
+  private Label scaleLabel;
+  @FXML
+  private Label loadingLabel;
```
tschie: Might want to give this another name now. Also, since the text is only set once, it can be done in the FXML.

src/main/java/com/esri/samples/map/map_reference_scale/MapReferenceScaleController.java
```diff
@@ -0,0 +1,162 @@
...
+
+      scaleVBox.setVisible(true);
+      layerVBox.setVisible(true);
+
+      loadingLabel.setText("Apply Reference Scale");
+    });
+  }
+
+  /**
+   * Set the map reference scale when a selection is made using the combobox
```
tschie: Javadoc should have punctuation

src/main/java/com/esri/samples/map/map_reference_scale/MapReferenceScaleController.java
```diff
@@ -0,0 +1,162 @@
...
+  @FXML
+  private void handleComboBoxSelection() {
+    setMapReferenceScale(scaleComboBox.getSelectionModel().getSelectedItem());
+  }
+
+  /**
+   * Takes the reference scale from the combobox, and sets it as the map's reference scale.
+   */
+  @FXML
+  private void handleScaleButtonClicked() {
```
tschie, line 128: I was thinking it would be like this:

```java
   /**
   * Set the map's reference scale to the scale selected in the combo box.
   */
  @FXML
  private void handleComboBoxSelection() {
    
    map.setReferenceScale(scaleComboBox.getSelectionModel().getSelectedItem());
  }

  /**
   * Set the current viewpoint scale to the map's reference scale.
   */
  @FXML
  private void handleScaleButtonClicked() {
    
    // get the center of the current viewpoint extent
    Point centerPoint = mapView.getCurrentViewpoint(Viewpoint.Type.CENTER_AND_SCALE).getTargetGeometry().getExtent().getCenter();
    // get the map's current reference scale
    double currentReferenceScale = mapView.getMap().getReferenceScale();
    // set a viewpoint with the scale at the map's reference scale
    Viewpoint newViewPoint = new Viewpoint(centerPoint, currentReferenceScale);
    mapView.setViewpointAsync(newViewPoint);
  }
```

The first handler already sets the reference scale when a selection is made. No need to set it again in the second handler (just get it from the map).

src/main/java/com/esri/samples/map/map_reference_scale/MapReferenceScaleController.java
```diff
@@ -0,0 +1,162 @@
...
+      }
+    });
+
+    // create a label to display current scale of the map
+    mapView.addMapScaleChangedListener(event -> {
+      scaleLabel.setText("Current Map Scale 1:" + Math.round(mapView.getMapScale()));
+    });
+
+    map.addDoneLoadingListener(() -> {
+
```
tschie: should check `map.getLoadStatus() == LoadStatus.LOADED` and show alert otherwise

src/main/java/com/esri/samples/map/map_reference_scale/MapReferenceScaleController.java
```diff
@@ -0,0 +1,162 @@
...
+      progressIndicator.setVisible(false);
+
+      LayerList operationalLayers = map.getOperationalLayers();
+
+      for (Layer layer : operationalLayers) {
+        // create a checkbox per operational layer name
+        CheckBox checkBox = new CheckBox(layer.getName());
+        checkBox.setSelected(true);
+        // add the checkboxes to the VBox
+        layerVBox.getChildren().add(checkBox);
```
tschie: Move the check box logic inside `layer instanceof FeatureLayer`. You don't want to create check boxes for non feature layers. Basically we just ignore any non feature layers.

src/main/java/com/esri/samples/map/map_reference_scale/MapReferenceScaleController.java
```diff
@@ -0,0 +1,162 @@
...
+
+        if (layer instanceof FeatureLayer) {
+
+          FeatureLayer featureLayer = (FeatureLayer) layer;
+          // set if the feature layer will honor the reference scale
+          checkBox.setOnAction(event -> {
```
tschie: any time you have a single line expression in a lambda you can omit the braces.

tschie (changes requested):
Good changes. Almost there!

Rachael-E:
@tschie thanks for catching those! All addressed now and the README updated. Could you take another look? Thanks!

tschie (approved):

PR #323, opened 2019-03-12
point scene layer sample

tschie:
displays a point scene layer from a scene layer service


tschie:
@Rachael-E Please review when you get a chance

src/main/java/com/esri/samples/scene/add_a_point_scene_layer/AddAPointSceneLayerSample.java
```diff
@@ -0,0 +1,100 @@
...
+public class AddAPointSceneLayerSample extends Application {
+
+  private SceneView sceneView;
+
+  @Override
+  public void start(Stage stage) {
+
+    try {
+
+      // set the title and size of the stage and show it
```
Rachael-E, line 44: for consistency with other samples, move the two lines of code with StackPane and Scene initiation above the stage methods, so that all 5 stage methods are grouped together.

src/main/java/com/esri/samples/scene/add_a_point_scene_layer/AddAPointSceneLayerSample.java
```diff
@@ -0,0 +1,100 @@
...
+      stage.setTitle("Add A Point Scene Layer Sample");
```
Rachael-E: "A" can be lower case

src/main/java/com/esri/samples/scene/add_a_point_scene_layer/README.md
```diff
@@ -0,0 +1,41 @@
+<h1>Add a point scene layer</h1>
+
+<p>View a point scene layer from a scene service.</p>
+
+<p><img src="AddAPointSceneLayer.png"/></p>
+
+<h2>Use case</h2>
+
+<p>Point scene layers can efficiently display large amounts of point features. While point cloud layers can only display simple symbols, point scene layers can display any type of billboard symbol or even 3D models, as long as the location of the symbol can be described by a point. Points are cached and automatically thinned when zoomed out to improve performance.</p>
```
Rachael-E, line 9: To really emphasise this point...could each point be represented by our 3D plane model instead of a simple white circle? It would mean having to change the base data in Pro/AGOL though. Just a thought.

src/main/java/com/esri/samples/scene/add_a_point_scene_layer/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+<h2>How to use the sample</h2>
+
+<p>Pan around the scene and zoom in. Notice how many thousands of additional features appear at each successive zoom scale.</p>
```
Rachael-E, line 13: it also shows features 'behind' the active face of the globe (e.g. looking on Africa airports from Australia are visible behind): is this intentional?

Rachael-E (commented):
@tschie  Looking good! Just some refactoring / wordsmithing suggestions. And thoughts on the data.

src/main/java/com/esri/samples/scene/add_a_point_scene_layer/README.md
tschie: I think this is a good idea but not sure we have control over the service. We can look at it again after the release.

src/main/java/com/esri/samples/scene/add_a_point_scene_layer/README.md
tschie: That definitely sounds like a bug.

tschie: I reproduced it with 2328 (super trippy!) and then saw it looks fixed with the latest build 2358. Could you confirm it looks okay to you with that build?

Rachael-E: Oh whoops, I must have been checking it with the out of date build. With 2358 it looks good now!

Rachael-E (approved):
This is good to merge now @tschie :)

PR #325, opened 2019-03-18
View content beneath terrain surface

Rachael-E:
@tschie as we discussed, here's a PR for the view content beneath terrain surface sample. Once the core issues are addressed we should be able to remove the layer view state changed status. Thanks!


src/main/java/com/esri/samples/scene/view_content_beneath_terrain_surface/ViewContentBeneathTerrainSurfaceSample.java
```diff
@@ -0,0 +1,82 @@
...
+    // create a scene from a web scene Url and set it to the scene view
+    ArcGISScene scene = new ArcGISScene("https://www.arcgis.com/home/item.html?id=91a4fafd747a47c7bab7797066cb9272");
+    sceneView = new SceneView();
+    sceneView.setArcGISScene(scene);
+
+    // add a progress indicator to show the scene is loading
+    ProgressIndicator progressIndicator = new ProgressIndicator(ProgressIndicator.INDETERMINATE_PROGRESS);
+
+    // when the scene has loaded, set navigation constraint and opacity to see below the surface
+    scene.addDoneLoadingListener(() -> {
```
tschie, line 40: I think these could go outside the load listener:

```java
scene.getBaseSurface().setNavigationConstraint(NavigationConstraint.NONE);
// set opacity to view content beneath the base surface
scene.getBaseSurface().setOpacity(0.5f);
```

tschie: Also it may be better to use a draw status changed listener for the progress indicator instead of a done loading listener:
```java
sceneView.addDrawStatusChangedListener(e -> {
   if (e.getDrawStatus() == DrawStatus.COMPLETED) {
   }
});
```

src/main/java/com/esri/samples/scene/view_content_beneath_terrain_surface/README.md
```diff
@@ -0,0 +1,35 @@
+<h1>View content beneath terrain surface</h1>
+
+<p>See through terrain in a scene and move the camera underground.</p>
+
+<p><img src="ViewContentBeneathTerrainSurface.png" alt="" /></p>
```
tschie: don't need the `alt` attribute

tschie (approved):
Looks good once the bug is fixed!

colinanderson:
>we should be able to remove the layer view state changed status.

This should no longer be needed now.

src/main/java/com/esri/samples/scene/view_content_beneath_terrain_surface/ViewContentBeneathTerrainSurfaceSample.java
Rachael-E: for the first point, the navigation constraint and setting the opacity will only update when they are contained within the scene loading listener, so I've left that in there.

Rachael-E:
@tschie the bug is fixed and sample working as expected! When you get a moment please can you take another look.

tschie:
@Rachael-E Looks good!

PR #327, opened 2019-03-22
remove online elevation data source

Rachael-E:
Now that the scene package has valid elevation included, I've removed the online elevation source to better demonstrate the offline use case. @tschie can you approve when you have the time, many thanks.


tschie (approved):
I think the two lines creating and setting a scene above aren't need either. Right? Not sure how I didn't catch that before.

Rachael-E:
oh whoops, well spotted, thanks!

PR #330, opened 2019-03-25
Generate offline map with local basemap

Rachael-E:
Another sample for review please when you have the chance @tschie: this one for generating an offline map using a local basemap. Thanks!


src/main/java/com/esri/samples/map/generate_offline_map_with_local_basemap/GenerateOfflineMapWithLocalBasemap.java
```diff
@@ -0,0 +1,331 @@
...
+      // create a map with the portal item
+      map = new ArcGISMap(portalItem);
+      addMapLoadedListener();
+      // set the map to the map view
+      mapView.setMap(map);
+      // update the download area box whenever the viewpoint changes
+      mapView.addViewpointChangedListener(viewpointChangedEvent -> updateDownloadArea());
+
+      // create a progress bar to show download progress
+      createProgressIndicator();
```
tschie: Generally, you shouldn't use methods to initialize UI. You can either move the code inline here, or use FXML.

src/main/java/com/esri/samples/map/generate_offline_map_with_local_basemap/GenerateOfflineMapWithLocalBasemap.java
```diff
@@ -0,0 +1,331 @@
...
+      stage.setWidth(800);
+      stage.setHeight(700);
+      stage.setScene(scene);
+      stage.show();
+
+      // create a map view
+      mapView = new MapView();
+
+      // create a button to take the map offline
+      createTakeMapOfflineButton();
```
tschie: Same comment here. Inline or convert to FXML.

src/main/java/com/esri/samples/map/generate_offline_map_with_local_basemap/GenerateOfflineMapWithLocalBasemap.java
```diff
@@ -0,0 +1,331 @@
...
+          if (minScale <= maxScale) {
+            minScale = maxScale + 1;
+          }
+
+          // create an offline map task with the map
+          task = new OfflineMapTask(map);
+          // create default generate offline map parameters
+          ListenableFuture<GenerateOfflineMapParameters> generateOfflineMapParametersListenableFuture = task.createDefaultGenerateOfflineMapParametersAsync(downloadArea.getGeometry(), minScale, maxScale);
+          // get the offline map parameters from the offline map task
+          generateOfflineMapParameters = generateOfflineMapParametersListenableFuture.get();
```
tschie: Move this into the done listener callback. Otherwise it will just block the thread until done. This will also move the try/catch inside the done listener. See other samples with listenable futures for reference.

src/main/java/com/esri/samples/map/generate_offline_map_with_local_basemap/GenerateOfflineMapWithLocalBasemap.java
```diff
@@ -0,0 +1,331 @@
...
+          generateOfflineMapParametersListenableFuture.addDoneListener(() -> {
+
+            // define the directory in which the downloaded .tpk file relevant to this offline map is saved
+            downloadedBasemapSavedLocation = new File("./samples-data/naperville/").getAbsolutePath();
+            // get the name of the basemap file from the offline map as supplied by the map's author (in this instance naperville_imagery.tpk)
+            String referenceBasemapFileName = generateOfflineMapParameters.getReferenceBasemapFilename();
+
+            // if the basemap file name isn't empty, search for that file name within local computer
+            if (!referenceBasemapFileName.isEmpty()) {
```
tschie: Is it likely this could be empty? Or could it be null (if so, you'd want to check for that)?

src/main/java/com/esri/samples/map/generate_offline_map_with_local_basemap/GenerateOfflineMapWithLocalBasemap.java
```diff
@@ -0,0 +1,331 @@
...
+                alert.show();
+                progressBar.setVisible(false);
+              }
+            } else {
+              Alert alert = new Alert(Alert.AlertType.ERROR, "The map's author has not specified a local basemap");
+              alert.show();
+            }
+          });
+
+        } catch (InterruptedException e1) {
```
tschie: If you do the same thing for difference exceptions, you can combine them:
```java
catch (InterrupatedException | ExecutionException ex) {
  ex.printStackTrace();
}
```

You should probably show an alert to the user though. Something like "Failed to create generate offline map parameters."

src/main/java/com/esri/samples/map/generate_offline_map_with_local_basemap/GenerateOfflineMapWithLocalBasemap.java
```diff
@@ -0,0 +1,331 @@
...
+        downloadArea.setGeometry(envelope);
+      }
+    }
+  }
+
+  /**
+   * Waits for the map to load before enabling interaction with the take map offline
+   * button, and adding a red square on to the map which defines the download area.
+   */
+  private void addMapLoadedListener(){
```
tschie: No need for a separate method.

src/main/java/com/esri/samples/map/generate_offline_map_with_local_basemap/GenerateOfflineMapWithLocalBasemap.java
```diff
@@ -0,0 +1,331 @@
...
+      // update the download area box whenever the viewpoint changes
+      mapView.addViewpointChangedListener(viewpointChangedEvent -> updateDownloadArea());
+
+      // create a progress bar to show download progress
+      createProgressIndicator();
+
+      // when the take map offline button is clicked, start the offline map task job
+      offlineMapButton.setOnAction(e -> {
+        try {
+          offlineMapButton.setVisible(false);
```
tschie: Better to disable buttons than make them invisible

src/main/java/com/esri/samples/map/generate_offline_map_with_local_basemap/GenerateOfflineMapWithLocalBasemap.java
```diff
@@ -0,0 +1,331 @@
...
+      // on any error, display the stack trace.
+      e.printStackTrace();
+    }
+  }
+
+  /**
+   * Called when the take map offline button is clicked. Builds parameters for the offline map task from the UI
+   * inputs and executes the task.
+   */
+  private void generateOfflineMap() throws IOException {
```
tschie: I assume this method can only be called once since you can only take a map offline once. If that's the case, you don't need a separate method.

src/main/java/com/esri/samples/map/generate_offline_map_with_local_basemap/GenerateOfflineMapWithLocalBasemap.java
```diff
@@ -0,0 +1,331 @@
...
+              if (localBasemapFile.exists()) {
+                // get the file's directory
+                localBasemapDirectory = localBasemapFile.getParent();
+
+                try {
+                  // open a new dialog box to alert user to the existence of the .tpk saved already on their machine
+                  chooseWhichBasemapToLoad();
+
+                } catch (IOException e1) {
+                  e1.printStackTrace();
```
tschie: Only the top most try/catch prints the stack trace (because if an exception bubbles up that far, we likely don't know of it's cause well enough to communicate that in an alert). All other exception handling should show an alert to the user.

tschie (changes requested):
I left some comments about code structure. I know some other SDKs break out their code into lots of single-use methods, but I like to avoid that with the Java samples. Let me know if you'd like to chat and go over these.

tschie:
@Rachael-E I've opened a PR with my proposed edits as we discussed.

tschie (approved):
Looks good, but the README needs to be converted to HTML. Good to merge after that

PR #338, opened 2019-04-18
remove use of arcgis gradle plugin

tschie:
handle native dependencies without the ArcGIS Runtime Gradle plugin

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -21,6 +21,9 @@ repositories {
     maven {
         url 'https://esri.bintray.com/arcgis'
     }
+    maven {
+        url 'http://olympus.esri.com/artifactory/arcgisruntime-repo'
+    }
 }
 
 configurations {
--- a/build.gradle
+++ b/build.gradle
@@ -1,24 +1,13 @@
-buildscript {
-    repositories {
-        maven {
-            url 'https://esri.bintray.com/arcgis'
-        }
-    }
-    dependencies {
-        classpath 'com.esri.arcgisruntime:gradle-arcgis-java-plugin:1.0.0'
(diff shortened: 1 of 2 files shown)


Rachael-E (commented):
I've tested this out, and the samples download and run fine. Though to continue using right click to run the samples, as Tyler suggested I added the environment variable into the run configuration. `ARCGISRUNTIMESDKJAVA_100_5_0=build/natives`. This then seems to run fine. @mbcoder is this then behaving as you'd expect?

build.gradle
```diff
@@ -82,8 +81,19 @@ task downloadData {
...
+    configurations.natives.asFileTree.each {
+        from(zipTree(it))
+    }
+    into "build/natives"
+}
+
 run {
+    dependsOn copyNatives
     mainClassName = project.hasProperty("mainClass") ? mainClass : 'com.esri.samples.map.display_map.DisplayMapSample'
+    environment "ARCGISRUNTIMESDKJAVA_100_5_0", "build/natives"
```
Rachael-E: @mbcoder and I were wondering if the environment variable set up should come before applying the mainClassName: if there are any timing issues then the main class name might be found before the runtime environment variable is. If timing could be an issue within the run task, then the environment variable set up should come before class name.

Rachael-E (approved):
@tschie just a query from @mbcoder and myself about ordering in the run task. Otherwise good to merge!

build.gradle
tschie: The order of these properties does not matter. All task configurations just set things up. Gradle determines the correct order to do everything.

PR #342, opened 2019-04-29
Add a point scene layer sample

JonLavi:
Removed some unused imports


JonLavi:
Got into a bit of a mixup with my branches, feel free to delete.
Should have just been the removal of unused imports in AddAPointSceneLayer, but I accidentally combined it with the commits of another sample I am working on.

PR #344, opened 2019-05-08
Edit and sync features

JonLavi:
Hi @tschie, @Rachael-E, I've got a PR for Edit And Sync Features (inspired by the [Android](https://github.com/Esri/arcgis-runtime-samples-android/tree/master/java/edit-and-sync-features) sample). When you have some time, I'd be happy to take some feedback. This one is a little more complex so don't feel rushed.


src/main/java/com/esri/samples/editing/edit_and_sync_features/README.md
```diff
@@ -0,0 +1,56 @@
+<h1>Edit and sync features</h1>
+
+<p>Synchronize offline edits with a feature service.</p>
+
+<p><img src="EditAndSyncFeatures.png/></p>
```
Rachael-E: this image is missing

src/main/java/com/esri/samples/editing/edit_and_sync_features/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+<h2>Use Case</h2>
```
Rachael-E: sentence case for titles

src/main/java/com/esri/samples/editing/edit_and_sync_features/README.md
```diff
@@ -0,0 +1,56 @@
...
+  <li>GenerateGeodatabaseJob</li>
+  <li>GenerateGeodatabaseParameters</li>
+  <li>SyncGeodatabaseJob</li>
+  <li>SyncGeodatabaseParameters</li>
+  <li>SyncLayerOption</li>
+</ul>
+
+<h2>About the data</h2>
+
+<p>The basemap for this sample is  a San Francisco offline tile package, provided by ESRI to support ArcGIS Runtime SDK Samples. The <i>WildfireSync</i> feature service elements illustrate a collection schema for wildlfire information.</p>
```
Rachael-E: additional space on the first line between "is" and "a"

Rachael-E: also a typo in the last use of wildfire at the end of the last sentence

src/main/java/com/esri/samples/editing/edit_and_sync_features/README.md
```diff
@@ -0,0 +1,56 @@
...
+    </ol>
+  </li>
+</ol>
+
+<h2>Relevant API</h2>
+
+<ul>
+  <li>FeatureLayer</li>
+  <li>FeatureTable</li>
+  <li>GeodatabaseSyncTask</li>
```
Rachael-E: alphabetical: GeodatabaseSyncTask should come after the last GenerateGeodatabase list

src/main/java/com/esri/samples/editing/edit_and_sync_features/README.md
Rachael-E: and the last inverted comma to close the file reference is missing: this is causing the .md not to render correctly.

src/main/java/com/esri/samples/editing/edit_and_sync_features/README.md
```diff
@@ -0,0 +1,56 @@
...
+  <li>Click on the Generate Geodatabase button to make an offline database of the area. Once the job completes successfully, the available features within this area will be displayed.</li>
+  <li>A feature can be selected by tapping on it. The selected feature can be moved to a new location by tapping anywhere on the map. </li>
+  <li>Once a successful edit has been made to a feature, the Sync Geodatabase button is enabled. Press this button to synchronize the edits made to the local geodatabase with the remote feature service.</li>
+</ol>
+
+<h2>How it works</h2>
+
+<ol>
+  <li>Create a <code>GeodatabaseSyncTask</code> from a URL.</li>
+  <li>Use <code>createDefaultGenerateGeodatabaseParametersAsync(...)</code> to create <code>GenerateGeodatabaseParameters</code> from the <code>GeodatabaseSyncTask</code>, passing in an <code>Envelope</code> argument.</li>
```
Rachael-E: remove the full stops between brackets at the end of the method call e.g. <code>createDefaultGenerateGeodatabaseParametersAsync()</code>. I haven't seen the syntax (...) before in our ReadMes so best to keep it as () until we can check with @tschie.

The descriptions here can be made clearer. 

"Use <code>createDefaultGenerateGeodatabaseParametersAsync()</code> on the geodatabase sync task to create <code>GenerateGeodatabaseParameters</code>, passing in an <code>Envelope</code> extent as the parameter.

src/main/java/com/esri/samples/editing/edit_and_sync_features/README.md
```diff
@@ -0,0 +1,56 @@
...
+  <li>Create a <code>GenerateGeodatabaseJob</code> from the <code>GeodatabaseSyncTask</code> using <code>generateGeodatabaseAsync(...)</code> passing in parameters and a path to the local geodatabase.</li>
+  <li>Start the <code>GenerateGeodatabaseJob</code> and, on success, load the <code>Geodatabase</code>.</li>
+  <li>On successful loading, call <code>getGeodatabaseFeatureTables()</code> on the <code>Geodatabase</code> and add it to the <code>ArcGISMap</code>'s operational layers.</li>
+  <li>To sync changes between the local and web geodatabases:
+    <ol>
+      <li>Define <code>SyncGeodatabaseParameters</code> including setting the SyncGeodatabaseParameters.SyncDirection`.</li>
```
Rachael-E: "Create a <code>SyncGeodatabaseParameters</code> object, and set its sync direction with <code>syncGeodatabaseParameters.setSyncDirection()</code>"

src/main/java/com/esri/samples/editing/edit_and_sync_features/EditAndSyncFeaturesSample.java
```diff
@@ -0,0 +1,403 @@
...
+import com.esri.arcgisruntime.loadable.LoadStatus;
+import com.esri.arcgisruntime.mapping.ArcGISMap;
+import com.esri.arcgisruntime.mapping.Basemap;
+import com.esri.arcgisruntime.mapping.GeoElement;
+import com.esri.arcgisruntime.mapping.view.Graphic;
+import com.esri.arcgisruntime.mapping.view.GraphicsOverlay;
+import com.esri.arcgisruntime.mapping.view.IdentifyLayerResult;
+import com.esri.arcgisruntime.mapping.view.MapView;
+import com.esri.arcgisruntime.symbology.SimpleLineSymbol;
+import com.esri.arcgisruntime.tasks.geodatabase.*;
```
Rachael-E: imports should always be expanded, not condensed with the * all import notation (there's a setting in IntelliJ to stop this automatically happening)

src/main/java/com/esri/samples/editing/edit_and_sync_features/EditAndSyncFeaturesSample.java
```diff
@@ -0,0 +1,403 @@
...
+
+public class EditAndSyncFeaturesSample extends Application {
+
+  private MapView mapView;
+  private final AtomicInteger replica = new AtomicInteger();
```
Rachael-E: this doesn't need to be a member variable. Instead, instantiate this as a local variable where it is first used (ln 256)

src/main/java/com/esri/samples/editing/edit_and_sync_features/EditAndSyncFeaturesSample.java
```diff
@@ -0,0 +1,403 @@
...
+   * Opens and runs application.
+   *
+   * @param args arguments passed to this application
+   */
+  public static void main(String[] args) {
+
+    Application.launch(args);
+  }
+
+  private enum EditState { // enumeration to track editing of points
```
Rachael-E: in keeping with most other samples using Enums in the Java directory, move this up to after the member variables (above the start method). The comment for this should be above not to the side :

`// enumeration to track which phase of the workflow the sample is in.`

src/main/java/com/esri/samples/editing/edit_and_sync_features/EditAndSyncFeaturesSample.java
```diff
@@ -0,0 +1,403 @@
...
+import com.esri.arcgisruntime.mapping.view.MapView;
+import com.esri.arcgisruntime.symbology.SimpleLineSymbol;
+import com.esri.arcgisruntime.tasks.geodatabase.*;
+
+public class EditAndSyncFeaturesSample extends Application {
+
+  private MapView mapView;
+  private final AtomicInteger replica = new AtomicInteger();
+  private EditAndSyncFeaturesSample.EditState currentEditState;
+  private GraphicsOverlay graphicsOverlay;
```
Rachael-E: same as above for GraphicsOverlay: this can be a local variable, doesn't need to be a member variable

src/main/java/com/esri/samples/editing/edit_and_sync_features/EditAndSyncFeaturesSample.java
```diff
@@ -0,0 +1,403 @@
...
+      stage.setTitle("Edit and Sync Features Sample");
+      stage.setWidth(800);
+      stage.setHeight(700);
+      stage.setScene(scene);
+      stage.show();
+
+      // set edit state to not ready until geodatabase generation has completed successfully
+      currentEditState = EditState.NOTREADY;
+
+      // create a map view and add a map
```
Rachael-E: the map isn't being added at this point in the code. If you want, you could instantiate the ArcGISMap first, and then create the map view beneath that code.

src/main/java/com/esri/samples/editing/edit_and_sync_features/EditAndSyncFeaturesSample.java
```diff
@@ -0,0 +1,403 @@
...
+      mapView = new MapView();
+
+      // create a graphics overlay and symbol to mark the extent
```
Rachael-E: where's the symbol to mark the extent? At this point in the code, the only thing happening is instantiating the graphics overlay and adding it to the map view (it is empty at this point).

src/main/java/com/esri/samples/editing/edit_and_sync_features/EditAndSyncFeaturesSample.java
```diff
@@ -0,0 +1,403 @@
...
+      Basemap basemap = new Basemap(tiledLayer);
+      ArcGISMap map = new ArcGISMap(basemap);
+      mapView.setMap(map);
+
+      // create a control panel
+      VBox controlsVBox = new VBox(6);
+      controlsVBox.setBackground(new Background(new BackgroundFill(Paint.valueOf("rgba(0,0,0,0.3)"), CornerRadii.EMPTY, Insets.EMPTY)));
+      controlsVBox.setPadding(new Insets(10.0));
+      controlsVBox.setMaxSize(180, 20);
+      controlsVBox.getStyleClass().add("panel-region");
```
Rachael-E, line 133: for this line of code to be "activated", it requires the source style.css file to be added to the JavaFx Scene object (this can be added after it is instantiated on line 87).

src/main/java/com/esri/samples/editing/edit_and_sync_features/EditAndSyncFeaturesSample.java
```diff
@@ -0,0 +1,403 @@
...
+  private ProgressIndicator progressIndicator;
+
+  @Override
+  public void start(Stage stage) {
+
+    try {
+      // create stack pane and application scene
+      StackPane stackPane = new StackPane();
+      Scene scene = new Scene(stackPane);
+
```
Rachael-E, line 107: add the reference to the style.css file:
      `scene.getStylesheets().add(getClass().getResource("/css/style.css").toExternalForm());`

src/main/java/com/esri/samples/editing/edit_and_sync_features/EditAndSyncFeaturesSample.java
```diff
@@ -0,0 +1,403 @@
...
+      // create progress indicator
+      progressIndicator = new ProgressIndicator();
+      progressIndicator.setVisible(false);
+
+      // add listener to handle generate/sync geodatabase button
+      geodatabaseButton.setOnAction(e -> {
+        if (currentEditState == EditState.NOTREADY) {
+          // update button text and disable button
+          geodatabaseButton.setDisable(true);
+          geodatabaseButton.setText("Generating Geodatabase...");
```
Rachael-E, line 191: nice, very helpful to the user!

src/main/java/com/esri/samples/editing/edit_and_sync_features/README.md
```diff
@@ -0,0 +1,56 @@
...
+<p><img src="EditAndSyncFeatures.png/></p>
+
+<h2>Use Case</h2>
+
+<p>By generating a local geodatabase, a user can take an offline copy of a feature service, make changes to it while still offline, and later synchronize their edits to the online feature service. This is useful in cases where a network connection is not available (e.g. working in remote areas), but users still need to be able to make changes to features.</p>
+
+<h2>How to use the sample</h2>
+
+<ol>
+  <li>Pan and zoom into the desired area, making sure the area you want to take offline is within the current extent of the view. </li>
```
Rachael-E: in keeping with Java's other samples generating geodatabase parameters, this sample should follow their design of using an extent which is smaller than the whole extent of the screen (e.g. like the .NET implementation of this sample). It makes sense for Android to use the full extent because of their smaller screens but for Windows/MacOS/Linux it's more user intuitive to have a visual guide for what data they'd want to download. 

With these updates to the code, this step would read:

"Pan and zoom to the area you would like to download point features for, ensuring that all desired features are within the red rectangle".

src/main/java/com/esri/samples/editing/edit_and_sync_features/EditAndSyncFeaturesSample.java
```diff
@@ -0,0 +1,403 @@
...
+
+          // show the extent of the geodatabase to be generated
```
Rachael-E: // create a graphic (using a red line) to mark the extent of the geodatabase to be generated

src/main/java/com/esri/samples/editing/edit_and_sync_features/EditAndSyncFeaturesSample.java
```diff
@@ -0,0 +1,403 @@
...
+          final SimpleLineSymbol boundarySymbol = new SimpleLineSymbol(SimpleLineSymbol.Style.SOLID, 0xFF0000FF, 5);
```
Rachael-E: this should be red not blue to be in keeping with Java's other samples: 0xFFFF0000

src/main/java/com/esri/samples/editing/edit_and_sync_features/EditAndSyncFeaturesSample.java
```diff
@@ -0,0 +1,403 @@
...
+      progressIndicator = new ProgressIndicator();
+      progressIndicator.setVisible(false);
+
+      // add listener to handle generate/sync geodatabase button
+      geodatabaseButton.setOnAction(e -> {
+        if (currentEditState == EditState.NOTREADY) {
+          // update button text and disable button
+          geodatabaseButton.setDisable(true);
+          geodatabaseButton.setText("Generating Geodatabase...");
+
```
Rachael-E, line 192: the extent of the geodatabase to be drawn should be shown from when the app starts: not after the button is clicked. See [GenerateOfflineMapWithLocalBasemapSample](https://github.com/Esri/arcgis-runtime-samples-java/blob/master/src/main/java/com/esri/samples/map/generate_offline_map_with_local_basemap/GenerateOfflineMapWithLocalBasemapSample.java#L71) for an example of this, and check out the private method in there for `updateDownloadArea`.

src/main/java/com/esri/samples/editing/edit_and_sync_features/EditAndSyncFeaturesSample.java
Rachael-E: this doesn't need to be final

Rachael-E: doesn't need to be final

src/main/java/com/esri/samples/editing/edit_and_sync_features/EditAndSyncFeaturesSample.java
```diff
@@ -0,0 +1,403 @@
...
+        if (event.isStillSincePress() && event.getButton() == MouseButton.PRIMARY) {
+          // get screen point where user clicked
+          Point2D screenPoint = new Point2D(event.getX(), event.getY());
+
+          // get map location corresponding to screen point
+          Point mapPoint = mapView.screenToLocation(screenPoint);
+
+          // iterate through all feature layers in the map
+          for (Layer layer : mapView.getMap().getOperationalLayers()) {
+            final FeatureLayer featureLayer = (FeatureLayer) layer;
```
Rachael-E: doesn't have to be final

src/main/java/com/esri/samples/editing/edit_and_sync_features/EditAndSyncFeaturesSample.java
```diff
@@ -0,0 +1,403 @@
...
+            ListenableFuture<IdentifyLayerResult> results = mapView.identifyLayerAsync(featureLayer, screenPoint, 1, false, 1);
+            results.addDoneListener(() -> {
+              try {
+                // get selected feature, if it is a point
+                List<GeoElement> elementList = results.get().getElements();
+                if (elementList.size() > 0 && elementList.get(0) instanceof ArcGISFeature && elementList.get(0).getGeometry().getGeometryType() == GeometryType.POINT) {
+
+                  // clicked on a feature, select it
+                  ArcGISFeature selectedFeature = (ArcGISFeature) elementList.get(0);
+                  // TODO: this doesn't clear the selection in all FL, hence some elements stay selected?
```
Rachael-E: all TODOs should be cleared before merging to master: or is this a bug you are highlighting?

src/main/java/com/esri/samples/editing/edit_and_sync_features/EditAndSyncFeaturesSample.java
```diff
@@ -0,0 +1,403 @@
...
+
+            // identify any clicked feature
+            ListenableFuture<IdentifyLayerResult> results = mapView.identifyLayerAsync(featureLayer, screenPoint, 1, false, 1);
+            results.addDoneListener(() -> {
+              try {
+                // get selected feature, if it is a point
+                List<GeoElement> elementList = results.get().getElements();
+                if (elementList.size() > 0 && elementList.get(0) instanceof ArcGISFeature && elementList.get(0).getGeometry().getGeometryType() == GeometryType.POINT) {
+
+                  // clicked on a feature, select it
```
Rachael-E: this comment could be clearer

src/main/java/com/esri/samples/editing/edit_and_sync_features/EditAndSyncFeaturesSample.java
```diff
@@ -0,0 +1,403 @@
...
+                  ArcGISFeature selectedFeature = (ArcGISFeature) elementList.get(0);
+                  // TODO: this doesn't clear the selection in all FL, hence some elements stay selected?
+                  featureLayer.clearSelection(); // clear previous selection
```
Rachael-E: I don't think you need this line? to clear the selection, the user can right click.

src/main/java/com/esri/samples/editing/edit_and_sync_features/EditAndSyncFeaturesSample.java
```diff
@@ -0,0 +1,403 @@
...
+                  featureLayer.selectFeature(selectedFeature);
+
+                  // change the state to editing
+                  currentEditState = EditState.EDITING;
+
+                } else {
+                  // didn't click on a feature
```
Rachael-E: I'm a bit confused by this section. I think some clearer commenting would help. If the user didn't click on a feature, why is the selected feature having its geometry updated in this code block? Is this referring more to if the user didn't click on a point? Though the user can't select anything other than a point.

src/main/java/com/esri/samples/editing/edit_and_sync_features/EditAndSyncFeaturesSample.java
```diff
@@ -0,0 +1,403 @@
...
+                  ListenableFuture<FeatureQueryResult> selectedQuery = featureLayer.getSelectedFeaturesAsync();
+                  selectedQuery.addDoneListener(() -> {
+                    try {
+                      // check if a feature is currently selected
+                      FeatureQueryResult selectedQueryResult = selectedQuery.get();
+                      Iterator<Feature> features = selectedQueryResult.iterator();
+                      if (features.hasNext()) {
```
Rachael-E: comment as to what this means

src/main/java/com/esri/samples/editing/edit_and_sync_features/EditAndSyncFeaturesSample.java
```diff
@@ -0,0 +1,403 @@
...
+                            selectedFeature.setGeometry(mapPoint);
+                            selectedFeature.getFeatureTable().updateFeatureAsync(selectedFeature);
+
+                            // refresh ui to enable syncing
+                            currentEditState = EditState.READY;
+                            geodatabaseButton.setText("Sync Geodatabase");
+                            geodatabaseButton.setDisable(false);
+                          }
+                        });
+                      } // else nothing currently selected, do nothing
```
Rachael-E: I don't think a comment is needed here, unless it is referring to the code on line 213?

src/main/java/com/esri/samples/editing/edit_and_sync_features/EditAndSyncFeaturesSample.java
```diff
@@ -0,0 +1,403 @@
...
+                }
+              } catch (InterruptedException | ExecutionException e) {
+                displayMessage("Exception getting clicked feature", e.getCause().getMessage());
+              }
+            });
+          }
+        } else if (event.isStillSincePress() && event.getButton() == MouseButton.SECONDARY) {
+          // on secondary mouse click, clear feature selection
+          for (Layer layer : mapView.getMap().getOperationalLayers()) {
+            final FeatureLayer featureLayer = (FeatureLayer) layer;
```
Rachael-E: doesn't need to be final

src/main/java/com/esri/samples/editing/edit_and_sync_features/EditAndSyncFeaturesSample.java
```diff
@@ -0,0 +1,403 @@
...
+          // get geodatabase when done
+          geodatabaseJob.addJobDoneListener(() -> {
+            if (geodatabaseJob.getStatus() == Job.Status.SUCCEEDED) {
+              geodatabase = geodatabaseJob.getResult();
+              geodatabase.loadAsync();
+              geodatabase.addDoneLoadingListener(() -> {
+                if (geodatabase.getLoadStatus() == LoadStatus.LOADED) {
+                  // add the geodatabase FeatureTables to the map as a FeatureLayer
+                  geodatabase.getGeodatabaseFeatureTables().forEach(geodatabaseFeatureTable -> {
+                    // TODO: why is the adding of the FT to the map not within a doneListener?
```
Rachael-E: another TODO to be removed. The adding of the FT to the map is within a done loading listener for the geodatabase?

src/main/java/com/esri/samples/editing/edit_and_sync_features/EditAndSyncFeaturesSample.java
```diff
@@ -0,0 +1,403 @@
...
+                  displayMessage("Error loading geodatabase", geodatabase.getLoadError().getMessage());
+                }
+                // set edit state to ready, update button text
+                currentEditState = EditState.READY;
+                geodatabaseButton.setText("Sync Geodatabase");
+              });
+            } else if (geodatabaseJob.getError() != null) {
+              displayMessage("Error generating geodatabase", geodatabaseJob.getError().getMessage());
+            } else {
+              displayMessage("Unknon Error generating geodatabase", null);
```
Rachael-E: typo in unknown

Rachael-E (changes requested):
good start! There are some changes to be made though, the largest one stemming from differences in the Android implementation to Java. When the Java app loads it should display a red box for the extent of the geodatabase to be generated, not the whole extent.  General code tidy ups to address as well.

src/main/java/com/esri/samples/editing/edit_and_sync_features/EditAndSyncFeaturesSample.java
JonLavi: Oh, this one will have snuck in when I made some final changes. I found the intellij settings now so shouldn't happen again. Thanks!

src/main/java/com/esri/samples/editing/edit_and_sync_features/EditAndSyncFeaturesSample.java
JonLavi: Very good point, I rearranged it so it's structured more logically

src/main/java/com/esri/samples/editing/edit_and_sync_features/EditAndSyncFeaturesSample.java
JonLavi: Same as above, I must have moved things around and forgot to update the comments to reflect. Everything is done at the point now where the graphics are added, so all the relevant code is together

src/main/java/com/esri/samples/editing/edit_and_sync_features/EditAndSyncFeaturesSample.java
JonLavi: I've rearranged this and improved on the comments. It's necessary to clear the selection before assigning the new feature to the selection, because otherwise the user can select multiple features.
Also, if the user somehow ends up selecting multiple features, only the most recently selected one will be moved on successive clicks.

src/main/java/com/esri/samples/editing/edit_and_sync_features/EditAndSyncFeaturesSample.java
JonLavi: It's quite a tricky one, yeah. Maybe the best way to clarify this would be to only use the term feature? (since features can represent points, but not all points are necessarily features). 

In short, we need to find out whether we clicked on a feature or not. If we did, we want to select it (and de-select and previously selected ones, so only one feature is ever selected at any given time). If we didn't click on a feature we want to check whether a feature has already been selected (which means it will now be moved). If that's the case, then we move it. If no feature was selected, we just clicked anywhere else on the map, and we do nothing (which was the other comment you highlighted)

I've edited the comments to try and outline this decision tree a bit better.

JonLavi:
Thanks @Rachael-E , implemented the improvements you suggested and made the comments more clear. Let me know what you think.

src/main/java/com/esri/samples/editing/edit_and_sync_features/EditAndSyncFeaturesSample.java
```diff
@@ -156,41 +198,47 @@ public void start(Stage stage) {
...
-                        // move selected feature to clicked location
+                        // retrieve the currently selected feature
                         ArcGISFeature selectedFeature = (ArcGISFeature) features.next();
                         selectedFeature.loadAsync();
                         selectedFeature.addDoneLoadingListener(() -> {
                           if (selectedFeature.canUpdateGeometry()) {
-                            // apply the edits
+                            // move the selected feature to the clicked map point
                             selectedFeature.setGeometry(mapPoint);
+                            // update the feature table
```
Rachael-E: the text has already been set to "Sync Geodatabase" once the Generate Geodatabase has been completed (line 332). I think it would be better to disable the button after the generate geodatabase is completed, and once the feature has been selected and moved, then enable the button again.

So here, remove the code setting the text and after 334 add code to disable the button.

src/main/java/com/esri/samples/editing/edit_and_sync_features/EditAndSyncFeaturesSample.java
```diff
@@ -246,14 +292,15 @@ public void generateGeodatabase(Envelope extent) {
...
-              .createDefaultGenerateGeodatabaseParametersAsync(extent);
+              .createDefaultGenerateGeodatabaseParametersAsync(geodatabaseExtentGraphics.getGeometry());
       defaultParameters.addDoneListener(() -> {
         try {
           // set parameters and don't include attachments
           GenerateGeodatabaseParameters parameters = defaultParameters.get();
           parameters.setReturnAttachments(false);
 
           // create a temporary file for the geodatabase
+          final AtomicInteger replica = new AtomicInteger();
```
Rachael-E: after line 332, after setting the text to "Sync", add in code to disable the button (until a feature is clicked and moved, see comment above on this)

`geodatabaseButton.setDisable(true);`

src/main/java/com/esri/samples/editing/edit_and_sync_features/EditAndSyncFeaturesSample.java
```diff
@@ -156,41 +198,47 @@ public void start(Stage stage) {
...
                 List<GeoElement> elementList = results.get().getElements();
+                // determine whether the clicked feature represents a point that can be moved
                 if (elementList.size() > 0 && elementList.get(0) instanceof ArcGISFeature && elementList.get(0).getGeometry().getGeometryType() == GeometryType.POINT) {
 
-                  // clicked on a feature, select it
+                  // clear the previous selection (to avoid selecting multiple features at the same time)
+                  featureLayer.clearSelection();
+
+                  // grab the new feature and make it selected
                   ArcGISFeature selectedFeature = (ArcGISFeature) elementList.get(0);
```
Rachael-E: rename to newSelectedFeature to avoid confusion with selectedFeature on line 235

src/main/java/com/esri/samples/editing/edit_and_sync_features/EditAndSyncFeaturesSample.java
```diff
@@ -156,41 +198,47 @@ public void start(Stage stage) {
...
-                  // TODO: this doesn't clear the selection in all FL, hence some elements stay selected?
-                  featureLayer.clearSelection(); // clear previous selection
                   featureLayer.selectFeature(selectedFeature);
 
                   // change the state to editing
                   currentEditState = EditState.EDITING;
 
                 } else {
-                  // didn't click on a feature
+                  // the user did not click on a feature (implying no selection was consciously made)
```
Rachael-E: Still confused by this comment. At this point, the user has clicked a feature and wants to move it: so the selection is still active at this point and a selection was consciously made?

Rachael-E (changes requested):
nice changes! this is looking good, just a few more things to address.

src/main/java/com/esri/samples/editing/edit_and_sync_features/EditAndSyncFeaturesSample.java
JonLavi: If the user has previously clicked on a feature and selected it, then we can assume the intention with a click on the map is to move it. However there's still the possibility that we don't have a feature selected at this point, and the user is simply clicking around the map. I made a diagram to hopefully make this section of the code a bit clearer. Happy to take suggestions on how to structure this differently. Nested if loops don't really convey the intentions too well here, and neither do my comments, unfortunately!

Untitled Diagram (2)

src/main/java/com/esri/samples/editing/edit_and_sync_features/EditAndSyncFeaturesSample.java
JonLavi: Good call, no point allowing to sync it if we haven't made any changes yet.

src/main/java/com/esri/samples/editing/edit_and_sync_features/EditAndSyncFeaturesSample.java
JonLavi: Done. Don't need to disable it again since it's already disabled right after clicking it, before `generateGeodatabase()` it's called. So I've changed this function to not reactivate the button. It gets reactivated once a feature is moved, like you suggested. Looks like I had it activate twice before that, actually (once after generating the geoDB and once after moving a feature). 

Edit2: looks like I've been enabling/disabling the button in the `showProgress()` method! I've removed it from there so it now only does what it says on the name. Disabling the button happens at the appropriate spots.

src/main/java/com/esri/samples/editing/edit_and_sync_features/EditAndSyncFeaturesSample.java
Rachael-E: ok, in that case, I'd remove the first comment and leave only the "retrieve the selected features...." comment, IMO it's clearer that way.

src/main/java/com/esri/samples/editing/edit_and_sync_features/EditAndSyncFeaturesSample.java
```diff
@@ -0,0 +1,465 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package com.esri.samples.editing.edit_and_sync_features;
```
colinanderson, line 17: I don't think it is that common to have underscores in package names - maybe it is how samples are done. Underscores seem to only be suggested in special cases - https://docs.oracle.com/javase/tutorial/java/package/namingpkgs.html

src/main/java/com/esri/samples/editing/edit_and_sync_features/EditAndSyncFeaturesSample.java
```diff
@@ -0,0 +1,465 @@
...
+
+      // set title, size and add scene to stage
+      stage.setTitle("Edit and Sync Features Sample");
+      stage.setWidth(800);
+      stage.setHeight(700);
+      stage.setScene(scene);
+      stage.show();
+
+      // set edit state to not ready until geodatabase generation has completed successfully
+      currentEditState = EditState.NOTREADY;
```
colinanderson: You could just initialize this at line 82.

src/main/java/com/esri/samples/editing/edit_and_sync_features/EditAndSyncFeaturesSample.java
```diff
@@ -0,0 +1,465 @@
...
+
+      // create a basemap from the tiled layer, and add it to a map
+      Basemap basemap = new Basemap(tiledLayer);
+      map = new ArcGISMap(basemap);
+
+      // create a map view and add the map
+      mapView = new MapView();
+      mapView.setMap(map);
+
+      // create a graphics overlay to mark the extent
```
colinanderson: I think you should move this to be nearer where it is used. It seems odd to create it then do some UI stuff before using it 20 or so lines later.

src/main/java/com/esri/samples/editing/edit_and_sync_features/EditAndSyncFeaturesSample.java
```diff
@@ -0,0 +1,465 @@
...
+      GraphicsOverlay graphicsOverlay;
+      graphicsOverlay = new GraphicsOverlay();
+      mapView.getGraphicsOverlays().add(graphicsOverlay);
+
+      // create a control panel
+      VBox controlsVBox = new VBox(6);
+      controlsVBox.setBackground(new Background(new BackgroundFill(Paint.valueOf("rgba(0,0,0,0.3)"), CornerRadii.EMPTY, Insets.EMPTY)));
```
colinanderson, line 129: Can't you just use `a `Color` e.g. Color.red` or define a color e.g. `Color.color(0, 0, 0, 0.3);`?

src/main/java/com/esri/samples/editing/edit_and_sync_features/EditAndSyncFeaturesSample.java
```diff
@@ -0,0 +1,465 @@
...
+   */
+  private void generateGeodatabase() {
+
+    // define geodatabase sync task
+    geodatabaseSyncTask = new GeodatabaseSyncTask("https://sampleserver6.arcgisonline.com/arcgis/rest/services/Sync/WildfireSync/FeatureServer");
+    geodatabaseSyncTask.loadAsync();
+    geodatabaseSyncTask.addDoneLoadingListener(() -> {
+
+      // create generate geodatabase parameters for the current extent
+      final ListenableFuture<GenerateGeodatabaseParameters> defaultParameters = geodatabaseSyncTask
```
colinanderson: I would name this `defaultParametersFuture`.

src/main/java/com/esri/samples/editing/edit_and_sync_features/EditAndSyncFeaturesSample.java
```diff
@@ -0,0 +1,465 @@
...
+              .createDefaultGenerateGeodatabaseParametersAsync(geodatabaseExtentGraphics.getGeometry());
+      defaultParameters.addDoneListener(() -> {
+        try {
+          // set parameters and don't include attachments
+          GenerateGeodatabaseParameters parameters = defaultParameters.get();
+          parameters.setReturnAttachments(false);
+
+          // create a temporary file for the geodatabase
+          final AtomicInteger replica = new AtomicInteger();
+          File tempFile = File.createTempFile("gdb" + replica.getAndIncrement(), ".geodatabase");
```
colinanderson: I think just creating a temp file without the replica value is probably enough.

src/main/java/com/esri/samples/editing/edit_and_sync_features/EditAndSyncFeaturesSample.java
```diff
@@ -0,0 +1,465 @@
...
+          // get geodatabase when done
+          geodatabaseJob.addJobDoneListener(() -> {
+            if (geodatabaseJob.getStatus() == Job.Status.SUCCEEDED) {
+              geodatabase = geodatabaseJob.getResult();
+              geodatabase.loadAsync();
+              geodatabase.addDoneLoadingListener(() -> {
+                if (geodatabase.getLoadStatus() == LoadStatus.LOADED) {
+                  // add the geodatabase FeatureTables to the map as a FeatureLayer
+                  geodatabase.getGeodatabaseFeatureTables().forEach(geodatabaseFeatureTable -> {
+                    geodatabaseFeatureTable.loadAsync();
```
colinanderson: They should load automatically when added to the map so no need for this line.

src/main/java/com/esri/samples/editing/edit_and_sync_features/EditAndSyncFeaturesSample.java
```diff
@@ -0,0 +1,465 @@
...
+        if (event.isStillSincePress() && event.getButton() == MouseButton.PRIMARY) {
+          // get screen point where user clicked
+          Point2D screenPoint = new Point2D(event.getX(), event.getY());
+
+          // get map location corresponding to screen point
+          Point mapPoint = mapView.screenToLocation(screenPoint);
+
+          // iterate through all feature layers in the map
+          for (Layer layer : mapView.getMap().getOperationalLayers()) {
+            FeatureLayer featureLayer = (FeatureLayer) layer;
```
colinanderson: You should probably check that they are an `instanceof FeatureLayer` before casting - even if you know that they are all `FeatureLayer` users will cut and paste from this sample.

src/main/java/com/esri/samples/editing/edit_and_sync_features/EditAndSyncFeaturesSample.java
```diff
@@ -0,0 +1,465 @@
...
+
+            // start a process to identify any features at the clicked screen point
+            ListenableFuture<IdentifyLayerResult> results = mapView.identifyLayerAsync(featureLayer, screenPoint, 1, false, 1);
```
colinanderson: There's a method that can identify on all the layers in one go  - `public ListenableFuture<List<IdentifyLayerResult>> identifyLayersAsync(Point2D screenPoint, double tolerance, boolean returnPopupsOnly)`.

src/main/java/com/esri/samples/editing/edit_and_sync_features/EditAndSyncFeaturesSample.java
```diff
@@ -0,0 +1,465 @@
...
+            results.addDoneListener(() -> {
+              try {
+                // get the results of the identification process to determine whether a feature was clicked on
+                List<GeoElement> elementList = results.get().getElements();
+                // determine whether the clicked feature represents a point that can be moved
+                if (elementList.size() > 0 && elementList.get(0) instanceof ArcGISFeature && elementList.get(0).getGeometry().getGeometryType() == GeometryType.POINT) {
```
colinanderson: Instead of `elementList.size() > 0` you could have `!elementList.isEmpty()`.

colinanderson: I think you could arrange this to avoid calling `elementList.get(0)` multiple times e.g.
```
if (!elementList.isEmpty) {
  GeoElement element = elementList.get(0);
  if (element instanceof ArcGISFeature && element.getGeometry().getGeometryType() == GeometryType.POINT) {
  ...
  }
}
```

src/main/java/com/esri/samples/editing/edit_and_sync_features/EditAndSyncFeaturesSample.java
```diff
@@ -0,0 +1,465 @@
...
+          generateGeodatabase();
+        } else if (currentEditState == EditState.READY) {
+          // sync the geodatabase
+          syncGeodatabase();
+        }
+      });
+
+      // handle clicks on the map view to select and move features
+      mapView.setOnMouseClicked((event) -> {
+        if (event.isStillSincePress() && event.getButton() == MouseButton.PRIMARY) {
```
colinanderson: Based on there being a second call to `event.isStillSincePress()` below maybe it would be better to do something like the following here
```
if (!event.isStillSincePress()) {
  return;
}
```

src/main/java/com/esri/samples/editing/edit_and_sync_features/EditAndSyncFeaturesSample.java
```diff
@@ -0,0 +1,465 @@
...
+
+                            // refresh ui to enable syncing
+                            currentEditState = EditState.READY;
+                            geodatabaseButton.setText("Sync Geodatabase");
+                            geodatabaseButton.setDisable(false);
+                          }
+                        });
+                      }
+                    } catch (InterruptedException | ExecutionException e) {
+                      displayMessage("Exception getting selected feature", e.getCause().getMessage());
```
colinanderson: I think you want `e.getMessage()` here as there might not be a cause.

src/main/java/com/esri/samples/editing/edit_and_sync_features/EditAndSyncFeaturesSample.java
```diff
@@ -0,0 +1,465 @@
...
+                    }
+                  });
+                }
+              } catch (InterruptedException | ExecutionException e) {
+                displayMessage("Exception getting clicked feature", e.getCause().getMessage());
```
colinanderson: Same as above - `e.getMessage()`.

src/main/java/com/esri/samples/editing/edit_and_sync_features/EditAndSyncFeaturesSample.java
```diff
@@ -0,0 +1,465 @@
...
+      }
+    });
+  }
+
+  /**
+   * Updates the extent of the area marked with a red border to be used for geodatabase generation
+   */
+  private void updateGeodatabaseExtentEnvelope() {
+    if (map.getLoadStatus() == LoadStatus.LOADED) {
+      // get the upper left corner of the view
```
colinanderson: Comment is a bit misleading - maybe say "get upper left corner of extent"?

src/main/java/com/esri/samples/editing/edit_and_sync_features/EditAndSyncFeaturesSample.java
```diff
@@ -0,0 +1,465 @@
...
+   * @param job the job to show progress for
+   */
+  private void showProgress(Job job) {
+
+    // show progress
+    job.addProgressChangedListener(() -> progressIndicator.setVisible(true));
+
+    // hide progress indicator on complete
+    job.addJobDoneListener(() -> {
+      if (job.getStatus() == Job.Status.SUCCEEDED) {
```
colinanderson: Should we handle the case where the job fails - otherwise you'll be left with progress showing even though the job is finished.

src/main/java/com/esri/samples/editing/edit_and_sync_features/EditAndSyncFeaturesSample.java
```diff
@@ -0,0 +1,465 @@
...
+
+  /**
+   * Show a message in an alert dialog.
+   *
+   * @param title   title of alert
+   * @param message message to display
+   */
+  private void displayMessage(String title, String message) {
+
+    Platform.runLater(() -> {
```
colinanderson: I suspect all your uses of this method are running on the UI thread and if so there's no need for the `Platform.runLater`.

colinanderson (commented):
@JonLavi some comments for you. I think you should break up the body of this sample into some methods - you have a large block of chained listeners that might make it hard to follow.

src/main/java/com/esri/samples/editing/edit_and_sync_features/EditAndSyncFeaturesSample.java
JonLavi: The `BackgroundFill()` constructor seems to only take an instance of `Paint` (https://docs.oracle.com/javase/8/javafx/api/javafx/scene/layout/BackgroundFill.html)

colinanderson: `Color` is a `Paint`
image

src/main/java/com/esri/samples/editing/edit_and_sync_features/EditAndSyncFeaturesSample.java
JonLavi: Thanks for pointing that out, the samples convention is indeed to use underscores. Good to know the convention though.

src/main/java/com/esri/samples/editing/edit_and_sync_features/EditAndSyncFeaturesSample.java
JonLavi: Oh! And as if by magic I got it to work now - I must have had a typo somewhere. Thanks!

Rachael-E: That's good to know Colin, thanks! In all of the other samples, we use the Paint option. For consistency I'd therefore say stick to that in this sample. We can have a chat with Tyler and see if we want to refactor the other samples to make the code a bit neater with Colin's Color suggestion.

src/main/java/com/esri/samples/editing/edit_and_sync_features/EditAndSyncFeaturesSample.java
```diff
@@ -0,0 +1,470 @@
...
+          geodatabaseJob.addJobDoneListener(() -> {
+            if (geodatabaseJob.getStatus() == Job.Status.SUCCEEDED) {
+              geodatabase = geodatabaseJob.getResult();
+              geodatabase.loadAsync();
+              geodatabase.addDoneLoadingListener(() -> {
+                if (geodatabase.getLoadStatus() == LoadStatus.LOADED) {
+
+                  // add the geodatabase FeatureTables to the map as a FeatureLayer
+                  geodatabase.getGeodatabaseFeatureTables().forEach(geodatabaseFeatureTable -> {
+                    mapView.getMap().getOperationalLayers().add(new FeatureLayer(geodatabaseFeatureTable));
```
Rachael-E: Can remove the curly braces here to make an expression lambda (as discussed for ClosestFacilityStatic sample). Can also reduce the code slightly by just calling
`map.getOperationalLayers().add(new FeatureLayer(geodatabaseFeatureTable));`

Rachael-E (approved):

src/main/java/com/esri/samples/editing/edit_and_sync_features/EditAndSyncFeaturesSample.java
JonLavi: Another good spot, thanks!

tschie:
Should show the feature layer at the beginning so you know what features you are taking offline. You could easily select an extent which doesn't contain any features you can move.

tschie:
Jobs can report their progress. You should show this in a progress bar instead of showing a progress indicator.

src/main/java/com/esri/samples/editing/edit_and_sync_features/EditAndSyncFeaturesSample.java
```diff
@@ -0,0 +1,468 @@
...
+            ListenableFuture<IdentifyLayerResult> results = mapView.identifyLayerAsync(featureLayer, screenPoint, 1, false, 1);
+            results.addDoneListener(() -> {
+              try {
+                // get the results of the identification process to determine whether a feature was clicked on
+                List<GeoElement> elementList = results.get().getElements();
+
+                // determine whether the clicked feature represents a point that can be moved
+                if (!elementList.isEmpty()){
+                  GeoElement geoElement = elementList.get(0);
+                  if (geoElement instanceof ArcGISFeature && geoElement.getGeometry().getGeometryType() == GeometryType.POINT) {
```
tschie: Why only allow point features to be moved?

src/main/java/com/esri/samples/editing/edit_and_sync_features/EditAndSyncFeaturesSample.java
```diff
@@ -0,0 +1,468 @@
...
+          GenerateGeodatabaseParameters defaultParameters = defaultParametersFuture.get();
+          defaultParameters.setReturnAttachments(false);
+
+          // create a temporary file for the geodatabase
+          File tempFile = File.createTempFile("gdb",".geodatabase");
+          tempFile.deleteOnExit();
+
+          // create and start the job
+          GenerateGeodatabaseJob geodatabaseJob = geodatabaseSyncTask.generateGeodatabase(defaultParameters, tempFile.getAbsolutePath());
+          geodatabaseJob.start();
```
tschie, line 282: Jobs can report their progress. You should show this in a progress bar instead of a progress indicator.

src/main/java/com/esri/samples/editing/edit_and_sync_features/EditAndSyncFeaturesSample.java
```diff
@@ -0,0 +1,468 @@
...
+      mapView.addDrawStatusChangedListener(drawStatusChangedListener);
+
+      // create a listener used to update the extent of the geodatabase area when the viewpoint changes
+      ViewpointChangedListener viewpointChangedListener = (viewpointChangedEvent) -> updateGeodatabaseExtentEnvelope();
+
+      // add the listener to the map view, to update the extent whenever the viewpoint changes
+      mapView.addViewpointChangedListener(viewpointChangedListener);
+
+      // add listener to handle generate/sync geodatabase button
+      geodatabaseButton.setOnAction(e -> {
```
tschie, line 187: Instead, have two buttons, one for generating and one for syncing. The sync one should be disabled until edits are made. No need for a state enum.

src/main/java/com/esri/samples/editing/edit_and_sync_features/EditAndSyncFeaturesSample.java
tschie: Yes, use this method. You can get the FeatureLayer by casting the IdentifyLayerResult's LayerContent to a FeatureLayer (do an `instanceof` check first). Something like this (copied from another project I have):

```java
List<IdentifyLayerResult> layerResults = identifyFuture.get();
layerResults.forEach(identifyLayerResult -> {
    LayerContent layerContent = identifyLayerResult.getLayerContent();
    if (layerContent instanceof FeatureLayer) {
         List<GeoElement> geoElements = identifyLayerResult.getElements();
         geoElements.forEach(geoElement -> {
             if (geoElement instanceof Feature) {
                  ((FeatureLayer) layerContent).selectFeature((Feature) geoElement);
             }
         });
     }
});
```

src/main/java/com/esri/samples/editing/edit_and_sync_features/EditAndSyncFeaturesSample.java
```diff
@@ -0,0 +1,468 @@
...
+                    ArcGISFeature newSelectedFeature = (ArcGISFeature) elementList.get(0);
+                    featureLayer.selectFeature(newSelectedFeature);
+
+                    // change the state to editing
+                    currentEditState = EditState.EDITING;
+                  }
+
+                } else {
+                  // retrieve the selected features in the feature layer
+                  ListenableFuture<FeatureQueryResult> selectedQuery = featureLayer.getSelectedFeaturesAsync();
```
tschie: Simplify this by only allowing one feature to be selected. Multiple features  Since you know select the feature you identified, you already know which feature is selected, so you don't need to ask for it again.

src/main/java/com/esri/samples/editing/edit_and_sync_features/EditAndSyncFeaturesSample.java
```diff
@@ -0,0 +1,468 @@
...
+          // sync the geodatabase
+          syncGeodatabase();
+        }
+      });
+
+      // handle clicks on the map view to select and move features
+      mapView.setOnMouseClicked((event) -> {
+
+        // return on mouse movement during click (i.e. click and drag)
+        if (!event.isStillSincePress()){
```
tschie: Just change the two `if` branches below to have two conditions: mouse button is primary/secondary and event is still since press.

src/main/java/com/esri/samples/editing/edit_and_sync_features/EditAndSyncFeaturesSample.java
```diff
@@ -0,0 +1,468 @@
...
+    }
+  }
+
+  /**
+   * Generates a local geodatabase and sets it to the map.
+   */
+  private void generateGeodatabase() {
+
+    // define geodatabase sync task
+    geodatabaseSyncTask = new GeodatabaseSyncTask("https://sampleserver6.arcgisonline.com/arcgis/rest/services/Sync/WildfireSync/FeatureServer");
```
tschie: Display this feature layer at the beginning so you know what features you are downloading. At the moment, it's easy to download an extent that doesn't contain any point features to move.

src/main/java/com/esri/samples/editing/edit_and_sync_features/EditAndSyncFeaturesSample.java
tschie: Alternatively, only allow point features to be displayed/downloaded so you don't have to deal with moving more complex shapes.

tschie (changes requested):

tschie:
If you get a chance, download .NET's sample viewer and look at their implementation. Aside from a few UI differences, our implementation should be roughly the same.

src/main/java/com/esri/samples/editing/edit_and_sync_features/EditAndSyncFeaturesSample.java
JonLavi: I used the progress bar originally, but because the progress hovers at ca 5% for a long time and then jumps to 100%, I found it nicer to have the progress indicator to reassure the user the sample isn't frozen. Can I suggest showing both?

colinanderson: I think the jumping progress is fixed in 100.6 so it might be an idea to go with the progress bar knowing it will come good in the next release.

src/main/java/com/esri/samples/editing/edit_and_sync_features/EditAndSyncFeaturesSample.java
JonLavi: Changed so only point features are displayed (as in .NET sample)

src/main/java/com/esri/samples/editing/edit_and_sync_features/EditAndSyncFeaturesSample.java
JonLavi: Thanks - I've looked at the .NET sample, which uses the edit state to deciding what happens when clicks are registered, as opposed to checking whether something is selected. Not only does it make the code a lot more readable, but then I can use this way of querying all feature layers in one go.

src/main/java/com/esri/samples/editing/edit_and_sync_features/EditAndSyncFeaturesSample.java
JonLavi: Done. I didn't do this originally because I couldn't figure out how to de-select the feature without `featureLayer.clearSelection()`, i.e. knowing the feature layer it belongs to. Is there any other way to do this?

JonLavi:
Hi @tschie, thanks for the feedback, it was very helpful! As a result, significant revamp on this sample. Thanks for pointing out the .NET implementation, it's a lot clearer.
I've extracted quite a few bits of code into their own methods to help keep thing readable while doing the changes, but I suppose a few of them could go in-line again:
```
 moveSelectedFeature(mapPoint)
 findAndSelectFeature(screenPoint)
 clearSelectedFeature()
 monitorViewpointchanges()
```
Let me know which you prefer.

src/main/java/com/esri/samples/editing/edit_and_sync_features/EditAndSyncFeaturesSample.java
tschie: No, you need to know the feature layer to deselect one of its features. In this case, `clearSelection` would be just as good because you're only selecting one feature at a time. If you wanted to deselect a specific future, use `unselectFeature(feature)`.

In editing workflows, you'd probably right-click on features to deselect them. In that case, you'd be able to get the feature layer from the identify result's layer content, and then use `unselectFeature`.

PR #346, opened 2019-05-09
Closest facility static

JonLavi:
Hi @Rachael-E , another sample here. I've looked up alternative options to handling the async tasks (other than chaining `.addDoneLoadingListener`). However the current option reads top-to-bottom whereas other solutions implement one-off methods. So I would say it is better to have it like this for readability.
Looking forward to some feedback again!


src/main/java/com/esri/samples/na/closest_facility_static/README.md
```diff
@@ -0,0 +1,61 @@
+<h1>Closest Facility (Static)</h1>
+
+<p>Find routes from several locations to the respective closest facility.</p>
+
+<p><img src="ClosestFacilityStatic.png"/></p>
+
+<h2>Use Case</h2>
```
Rachael-E: sentence case

src/main/java/com/esri/samples/na/closest_facility_static/ClosestFacilityStaticSample.java
```diff
@@ -0,0 +1,344 @@
...
+
+      // create a ArcGISMap with a Basemap instance with an Imagery base layer
+      ArcGISMap map = new ArcGISMap(Basemap.createStreetsWithReliefVector());
+
+      // set the map to be displayed in this view
+      mapView = new MapView();
+      mapView.setMap(map);
+
+      // Add a graphics overlay to the map (will be used later to display routes)
+      mapView.getGraphicsOverlays().add(new GraphicsOverlay());
```
Rachael-E: instantiate a new GraphicsOverlay on the line above, in keeping with samples convention for splitting things up neatly:

`GraphicsOverlay graphicsOverlay = new GraphicsOverlay();
mapView.getGraphicsOverlays().add(graphicsOverlay);`

src/main/java/com/esri/samples/na/closest_facility_static/ClosestFacilityStaticSample.java
```diff
@@ -0,0 +1,344 @@
...
+                solveRoutes();
+              } catch (Exception ex) {
+                ex.printStackTrace();
+              }
+            });
+
+            // handle reset button press
+            resetButton.setOnAction(e -> {
+              try {
+                resetRoutes();
```
Rachael-E: this method should be replaced with the code used in the resetRoutes method: and that method then removed as it is only used once and is short.

src/main/java/com/esri/samples/na/closest_facility_static/ClosestFacilityStaticSample.java
```diff
@@ -0,0 +1,344 @@
...
+      stage.setHeight(700);
+      stage.setScene(scene);
+      stage.show();
+
+      VBox controlsVBox = new VBox(6);
+      controlsVBox.setBackground(new Background(new BackgroundFill(Paint.valueOf("rgba(0,0,0,0.3)"), CornerRadii.EMPTY,
+              Insets.EMPTY)));
+      controlsVBox.setPadding(new Insets(10.0));
+      controlsVBox.setMaxSize(150, 50);
+      controlsVBox.getStyleClass().add("panel-region");
```
Rachael-E: remember to add the link to the style.css file for this line of code to be effective. However, since this sample only includes buttons and no label/checkbox/radio button within the panel region, this line isn't needed in this sample anyway.

src/main/java/com/esri/samples/na/closest_facility_static/README.md
```diff
@@ -0,0 +1,61 @@
...
+  </li> 
+  <li>Display the route to <code>MapView</code>:
+    <ul>
+      <li>create <code>Graphic</code> from route geometry, <code>new Graphic(closestFacilityRoute.getRouteGeometry())</code></li>
+      <li>add graphic to <code>GraphicsOverlay</code> which is attached to the mapview</li>
+    </ul>
+  </li>
+</ol>
+
+<h2 id="relevantapi">Relevant API</h2>
```
Rachael-E: remove id tag

src/main/java/com/esri/samples/na/closest_facility_static/README.md
```diff
@@ -0,0 +1,61 @@
...
+  <li>ClosestFacilityResult</li>
+  <li>ClosestFacilityRoute</li>
+  <li>ClosestFacilityTask</li>
+  <li>Facility</li>
+  <li>Graphic</li>
+  <li>GraphicsOverlay</li>
+  <li>Incident</li>
+</ul>
+
+<h2 id="tags">Tags</h2>
```
Rachael-E: remove id tag

src/main/java/com/esri/samples/na/closest_facility_static/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+<p><img src="ClosestFacilityStatic.png"/></p>
+
+<h2>Use Case</h2>
+
+<p>Quickly and accurately determining the most efficient route between a location and a facility is a frequently encountered task (e.g. emergency services).</p>
+
+<h2>How to use the sample</h2>
+
+<p>Click the 'Solve Routes' button to determine and display the route from each incident (<i>'Fire'</i>-Icon) to the nearest facility (<i>'Fire Station'</i>-Icon). Click the reset button to remove the views.</p>
```
Rachael-E: remove mentions of icon and just refer to as: 

"Click the 'Solve Routes' button to determine and display the route from each incident (fire) to the nearest facility (fire station)"

src/main/java/com/esri/samples/na/closest_facility_static/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+<p>To display a <code>ClosestFacilityRoute</code> between several incidents and facilities:</p>
+
+<ol>
+  <li>Create a <code>ClosestFacilityTask</code> using a Url from an online service.</li>
+  <li>Get the default set of <code>ClosestFacilityParameters</code> from the task: <code>closestFacilityTask.createDefaultParametersAsync().get()</code></li>
+  <li>Build a list of all Facilities and Incidents:
+    <ul>
+      <li>Create a <code>FeatureTable</code> using <code> ServiceFeatureTable(Uri) </code> </li>
+      <li>Query the <code>FeatureTable</code> for all <code>Feature</code>s using <code>..queryFeaturesAsync(queryParameters)</code>
```
Rachael-E: remove additional . before queryFeaturesAsync method reference

src/main/java/com/esri/samples/na/closest_facility_static/README.md
```diff
@@ -0,0 +1,61 @@
...
+      <li>Iterate over the result and add each <code>Feature</code> to a <code>List</code>, instantiating the feature as a <code>Facility</code> or <code>Incident</code>, respectively.
```
Rachael-E: remove "respectively" from the end of the sentence, not needed here.

src/main/java/com/esri/samples/na/closest_facility_static/README.md
```diff
@@ -0,0 +1,61 @@
...
+  <li>Get <code>ClosestFacilityResult</code> from solving the task with the provided parameters: <code>closestFacilityTask.solveClosestFacilityAsync(closestFacilityParameters)</code></li>
+  <li>Find the closest facility for each incident by iterating over the previously created <code>incidentsList</code>:
+    <ul>
+      <li>Get index list of closet facilities to the incident, <code>closestFacilityResult.getRankedFacilityIndexes(indexOfIncident).get(0)</code></li>
+      <li>Find closest facility route, <code>closestFacilityResult.getRoute(closestFacilityIndex, indexOfIncident)</code></li>
+    </ul>
+  </li> 
+  <li>Display the route to <code>MapView</code>:
+    <ul>
+      <li>create <code>Graphic</code> from route geometry, <code>new Graphic(closestFacilityRoute.getRouteGeometry())</code></li>
```
Rachael-E: "create a <code>Graphic</code> from the route geometry, with <code>closestFacilityRoute.getRouteGeometry()"

src/main/java/com/esri/samples/na/closest_facility_static/README.md
```diff
@@ -0,0 +1,61 @@
...
+      <li>add graphic to <code>GraphicsOverlay</code> which is attached to the mapview</li>
```
Rachael-E: "add the graphic to the <code>GraphicsOverlay</code>n and set it to the map view."

src/main/java/com/esri/samples/na/closest_facility_static/ClosestFacilityStaticSample.java
```diff
@@ -0,0 +1,344 @@
...
+
+      // create a line symbol to mark the route
+      simpleLineSymbol = new SimpleLineSymbol(SimpleLineSymbol.Style.SOLID, 0x4D0000FF, 5.0f);
+
+      // create a ClosestFacilityTask
+      closestFacilityTask = new ClosestFacilityTask("https://sampleserver6.arcgisonline.com/arcgis/rest/services/NetworkAnalysis/SanDiego/NAServer/ClosestFacility");
+
+      // Create a table for facilities using the FeatureServer
+      FeatureTable facilitiesFeatureTable = new ServiceFeatureTable("https://services2.arcgis.com/ZQgQTuoyBrtmoGdP/ArcGIS/rest/services/San_Diego_Facilities/FeatureServer/0");
+
```
Rachael-E: Nit picky again but I think this is more readable if you remove this blank line: keeping all facilities set up code together, and the same for incidents code.

src/main/java/com/esri/samples/na/closest_facility_static/ClosestFacilityStaticSample.java
```diff
@@ -0,0 +1,344 @@
...
+      // Create a feature layer from the table, apply facilities icon
+      FeatureLayer facilitiesFeatureLayer = new FeatureLayer(facilitiesFeatureTable);
+      facilitiesFeatureLayer.setRenderer(new SimpleRenderer(facilitySymbol));
+
+      // Create a table for incidents using the FeatureServer
+      FeatureTable incidentsFeatureTable = new ServiceFeatureTable("https://services2.arcgis.com/ZQgQTuoyBrtmoGdP/ArcGIS/rest/services/San_Diego_Incidents/FeatureServer/0");
+
```
Rachael-E: remove this blank line

src/main/java/com/esri/samples/na/closest_facility_static/ClosestFacilityStaticSample.java
```diff
@@ -0,0 +1,344 @@
...
+              // use the task to solve for the closest facility
+              ListenableFuture<ClosestFacilityResult> closestFacilityTaskResult = closestFacilityTask.solveClosestFacilityAsync(closestFacilityParameters);
+              closestFacilityTaskResult.addDoneListener(() -> {
+                try {
+                  ClosestFacilityResult closestFacilityResult = closestFacilityTaskResult.get();
+
+                  // find the closest facility for each incident
+                  for (int indexOfIncident = 0; indexOfIncident < incidentsList.size(); indexOfIncident++) {
+
+                    // get the index of the closest facility to incident. (i) is the index of the incident, [0] is the index of the closest facility.
```
Rachael-E: keep this comment simply to

"get the index of the closest facility to incident."

It's clear from your code what is happening

src/main/java/com/esri/samples/na/closest_facility_static/ClosestFacilityStaticSample.java
```diff
@@ -0,0 +1,344 @@
...
+                    Integer closestFacilityIndex = closestFacilityResult.getRankedFacilityIndexes(indexOfIncident).get(0);
+
+                    // get the route to the closest facility.
```
Rachael-E: be consistent with comment style throughout, usually no full stop at the end of these comments

src/main/java/com/esri/samples/na/closest_facility_static/ClosestFacilityStaticSample.java
```diff
@@ -0,0 +1,344 @@
...
+   * @param uri the URI of the picture to be used for the symbol
+   */
+  private PictureMarkerSymbol createSymbol(String uri) {
+    PictureMarkerSymbol symbol = new PictureMarkerSymbol(uri);
+    symbol.setHeight(30);
+    symbol.setWidth(30);
+    return symbol;
+  }
+
+  private void resetRoutes() {
```
Rachael-E: this method is only used once and is fairly short so doesn't need to be separated out. Move the code out of this method directly into where it is used in the `start` method.

N.B. if we were keeping this in, it would need a Javadoc comment. Just a note for future apps :)

src/main/java/com/esri/samples/na/closest_facility_static/ClosestFacilityStaticSample.java
```diff
@@ -0,0 +1,344 @@
...
+        }
+      });
+
+    } catch (Exception e) {
+      e.printStackTrace();
+    }
+  }
+
+  /**
+   * Creates a PictureMarkerSymbol from a URI and sizes it appropriately
```
Rachael-E: JavaDoc requires punctuation, full stop at the end of the sentence here.

src/main/java/com/esri/samples/na/closest_facility_static/ClosestFacilityStaticSample.java
```diff
@@ -0,0 +1,344 @@
...
+    symbol.setHeight(30);
+    symbol.setWidth(30);
+    return symbol;
+  }
+
+  private void resetRoutes() {
+    // clear the route graphics.
+    mapView.getGraphicsOverlays().get(0).getGraphics().clear();
+
+    // Reset the buttons
```
Rachael-E: lower case "reset"

src/main/java/com/esri/samples/na/closest_facility_static/ClosestFacilityStaticSample.java
```diff
@@ -0,0 +1,344 @@
...
+   */
+  private PictureMarkerSymbol createSymbol(String uri) {
+    PictureMarkerSymbol symbol = new PictureMarkerSymbol(uri);
+    symbol.setHeight(30);
+    symbol.setWidth(30);
+    return symbol;
+  }
+
+  private void resetRoutes() {
+    // clear the route graphics.
```
Rachael-E: for consistency, no full stop at end of inline comment

Rachael-E (changes requested):
looking good! A few changes, mostly typos/code convention catches.

JonLavi:
Brilliant, thanks. I've moved both methods to be in-line. Comments should be consistently formatted now.

src/main/java/com/esri/samples/na/closest_facility_static/ClosestFacilityStaticSample.java
```diff
@@ -0,0 +1,334 @@
...
+
+      // create a feature layer from the table, apply incident icon
+      FeatureLayer incidentsFeatureLayer = new FeatureLayer(incidentsFeatureTable);
+      incidentsFeatureLayer.setRenderer(new SimpleRenderer(incidentSymbol));
+
+      // add the layers to the map
+      map.getOperationalLayers().addAll(Arrays.asList(facilitiesFeatureLayer, incidentsFeatureLayer));
+
+      // wait for both layers to load
+      facilitiesFeatureLayer.loadAsync();
```
colinanderson: You don't these `loadAsync` calls as the layers will be loaded automatically by the map.

src/main/java/com/esri/samples/na/closest_facility_static/ClosestFacilityStaticSample.java
```diff
@@ -0,0 +1,334 @@
...
+      incidentsFeatureLayer.loadAsync();
+
+      facilitiesFeatureLayer.addDoneLoadingListener(() -> {
+        incidentsFeatureLayer.addDoneLoadingListener(() -> {
```
colinanderson: I think as previously discussed there are better ways to do this. I would suggest add a comment that explains why this works i.e. that adding a listener to something that is already done will cause that listener to immediately fire.

src/main/java/com/esri/samples/na/closest_facility_static/ClosestFacilityStaticSample.java
```diff
@@ -0,0 +1,334 @@
...
+                              for (int indexOfIncident = 0; indexOfIncident < incidentsList.size(); indexOfIncident++) {
+
+                                // get the index of the closest facility to incident (0 is the index of the closest facility)
+                                Integer closestFacilityIndex = closestFacilityResult.getRankedFacilityIndexes(indexOfIncident).get(0);
+
+                                // get the route to the closest facility
+                                ClosestFacilityRoute closestFacilityRoute = closestFacilityResult.getRoute(closestFacilityIndex, indexOfIncident);
+
+                                // display the route on the graphics overlay
+                                mapView.getGraphicsOverlays().get(0).getGraphics().add(new Graphic(closestFacilityRoute.getRouteGeometry(), simpleLineSymbol));
```
colinanderson: Maybe you can just access `graphicsOverlay` here instead?

src/main/java/com/esri/samples/na/closest_facility_static/ClosestFacilityStaticSample.java
```diff
@@ -0,0 +1,334 @@
...
+              } catch (Exception ex) {
+                ex.printStackTrace();
+              }
+            });
+
+            // handle reset button press
+            resetButton.setOnAction(e -> {
+              try {
+                // clear the route graphics
+                mapView.getGraphicsOverlays().get(0).getGraphics().clear();
```
colinanderson: Can you just use `graphicsOverlay` here?

src/main/java/com/esri/samples/na/closest_facility_static/ClosestFacilityStaticSample.java
```diff
@@ -0,0 +1,334 @@
...
+
+                // reset the buttons
+                solveRoutesButton.setDisable(false);
```
colinanderson: I think you could probably use a binding to keep the buttons in sync e.g.
`resetButton.bind(!solveRouteButton.isDisable());`
and the reset button should always have the opposite disable value of the solve route button. With this all you would ever need to set would be the enable/diable of the solve route button.

You would add this code in the same place as where you create the buttons.

src/main/java/com/esri/samples/na/closest_facility_static/ClosestFacilityStaticSample.java
```diff
@@ -0,0 +1,334 @@
...
+                  ex.printStackTrace();
+                }
+
+              } catch (Exception ex) {
+                ex.printStackTrace();
+              }
+            });
+
+            // handle reset button press
+            resetButton.setOnAction(e -> {
```
colinanderson: I would probably have used `a` for action rather than `e`... for event?

colinanderson (commented):
@JonLavi a few comments for you.

src/main/java/com/esri/samples/na/closest_facility_static/ClosestFacilityStaticSample.java
JonLavi: Agreed, there's an implication in that code that we know that one of the tasks will finish before the other, but as you said, it doesn't technically matter.
@tschie would you be let us know if this is acceptable and needs commenting on? Thanks!

Rachael-E (approved):

src/main/java/com/esri/samples/na/closest_facility_static/ClosestFacilityStaticSample.java
```diff
@@ -0,0 +1,329 @@
...
+      controlsVBox.setBackground(new Background(new BackgroundFill(Paint.valueOf("rgba(0,0,0,0.3)"), CornerRadii.EMPTY,
+              Insets.EMPTY)));
+      controlsVBox.setPadding(new Insets(10.0));
+      controlsVBox.setMaxSize(150, 50);
+
+      // create buttons
+      Button solveRoutesButton = new Button("Solve Routes");
+      solveRoutesButton.setMaxWidth(Double.MAX_VALUE);
+      solveRoutesButton.setDisable(true);
+      Button resetButton = new Button("Reset");
```
tschie: This button seems unnecessary. It solves the route the same way each time.

src/main/java/com/esri/samples/na/closest_facility_static/ClosestFacilityStaticSample.java
```diff
@@ -0,0 +1,329 @@
...
+      // set the map to be displayed in this view
+      mapView = new MapView();
+      mapView.setMap(map);
+
+      // create a graphics overlay and add it to the map
+      GraphicsOverlay graphicsOverlay = new GraphicsOverlay();
+      mapView.getGraphicsOverlays().add(graphicsOverlay);
+
+      // create Symbols for displaying facilities
+      PictureMarkerSymbol facilitySymbol = createSymbol("https://static.arcgis.com/images/Symbols/SafetyHealth/FireStation.png");
```
tschie: Use the other constructor:

```java
new PictureMarkerSymbol(new Image("https://static.arcgis.com/images/Symbols/SafetyHealth/FireStation.png", 30, 30, true, false));
```

src/main/java/com/esri/samples/na/closest_facility_static/ClosestFacilityStaticSample.java
```diff
@@ -0,0 +1,329 @@
...
+
+          // retrieve a list of all facilities
+          ListenableFuture<FeatureQueryResult> result = facilitiesFeatureTable.queryFeaturesAsync(queryParameters);
+          result.addDoneListener(() -> {
+            try {
+              FeatureQueryResult facilitiesResult = result.get();
+
+              // add the found facilities to the list
+              for (Feature facilityFeature : facilitiesResult) {
+                facilitiesList.add(new Facility(facilityFeature.getGeometry().getExtent().getCenter()));
```
tschie: You could just cast the geometry to a Point since we know we are working with point features (could indicate this in a comment for the reader).

tschie: If they were not point features, the closest point would probably be more useful than the center.

src/main/java/com/esri/samples/na/closest_facility_static/ClosestFacilityStaticSample.java
```diff
@@ -0,0 +1,329 @@
...
+                  // solve closest facilities
+                  try {
+                    // use the task to solve for the closest facility
+                    ListenableFuture<ClosestFacilityResult> closestFacilityTaskResult = closestFacilityTask.solveClosestFacilityAsync(closestFacilityParameters);
+                    closestFacilityTaskResult.addDoneListener(() -> {
+                      try {
+                        ClosestFacilityResult closestFacilityResult = closestFacilityTaskResult.get();
+
+                        // find the closest facility for each incident
+                        for (int indexOfIncident = 0; indexOfIncident < incidentsList.size(); indexOfIncident++) {
```
tschie: indexOfIncident -> incidentIndex (this is what the argument is called in `getRankedFacilityIndexes`).

src/main/java/com/esri/samples/na/closest_facility_static/ClosestFacilityStaticSample.java
```diff
@@ -0,0 +1,329 @@
...
+      FeatureTable incidentsFeatureTable = new ServiceFeatureTable("https://services2.arcgis.com/ZQgQTuoyBrtmoGdP/ArcGIS/rest/services/San_Diego_Incidents/FeatureServer/0");
+      // create a feature layer from the table, apply incident icon
+      FeatureLayer incidentsFeatureLayer = new FeatureLayer(incidentsFeatureTable);
+      incidentsFeatureLayer.setRenderer(new SimpleRenderer(incidentSymbol));
+
+      // add the layers to the map
+      map.getOperationalLayers().addAll(Arrays.asList(facilitiesFeatureLayer, incidentsFeatureLayer));
+
+      // create the list to store the facilities
+      ArrayList<Facility> facilitiesList = new ArrayList<>();
```
tschie: Can just be simply `facilities`.

tschie (commented):
Looks good. I didn't have time to finish my review. I'll finish it tomorrow.

src/main/java/com/esri/samples/na/closest_facility_static/ClosestFacilityStaticSample.java
JonLavi: Looks like I have two options:
1. Make a single button that toggles between 'solve' and 'reset'
2. Just leave the 'solve' button and have the `GraphicsOverlay` clear at the beginning, so we have a few seconds of the routes disappearing and reappearing once the task is solved.

In terms of code complexity option 1 would require creating a 'state' for the sample, so added complexity. But I think it would make a nicer sample since it would feel more interactive. What are your thoughts?

src/main/java/com/esri/samples/na/closest_facility_static/ClosestFacilityStaticSample.java
JonLavi: Great shout - using `.getCenter()` did feel very clumsy!

src/main/java/com/esri/samples/na/closest_facility_static/ClosestFacilityStaticSample.java
tschie: Sorry, I should have been more clear. Just have the sample solve once and then disable the single Solve button. The user doesn't need to be able to solve multiple times.

src/main/java/com/esri/samples/na/closest_facility_static/ClosestFacilityStaticSample.java
tschie: Sometimes we do use `getCenter` on the extent when we want to do something with the viewpoint, but that's really the only case where I've found it useful in samples thus far.

src/main/java/com/esri/samples/na/closest_facility_static/ClosestFacilityStaticSample.java
```diff
@@ -183,7 +185,7 @@ public void start(Stage stage) throws Exception {
 
               // add the found incidents to the list
               for (Feature incidentFeature : incidentsResult) {
-                incidentsList.add(new Incident(incidentFeature.getGeometry().getExtent().getCenter()));
+                incidents.add(new Incident((Point) incidentFeature.getGeometry()));
```
tschie, line 171: don't forget a comment here and above that indicates why we know we can cast to Point in this sample (because our feature layers only have point features).

src/main/java/com/esri/samples/na/closest_facility_static/ClosestFacilityStaticSample.java
```diff
@@ -0,0 +1,319 @@
...
+
+      // set title, size, and add scene to stage
+      stage.setTitle("Closest Facility (Static) Sample");
+      stage.setWidth(800);
+      stage.setHeight(700);
+      stage.setScene(scene);
+      stage.show();
+
+      // create control panel
+      VBox controlsVBox = new VBox(6);
```
tschie: When/if you remove the reset button, you could also get rid of this.

src/main/java/com/esri/samples/na/closest_facility_static/ClosestFacilityStaticSample.java
```diff
@@ -0,0 +1,319 @@
...
+      PictureMarkerSymbol facilitySymbol = new PictureMarkerSymbol(new Image("https://static.arcgis.com/images/Symbols/SafetyHealth/FireStation.png", 30, 30, true, false));
+      PictureMarkerSymbol incidentSymbol = new PictureMarkerSymbol(new Image("https://static.arcgis.com/images/Symbols/SafetyHealth/esriCrimeMarker_56_Gradient.png", 30, 30, true, false));
+
+      // create a line symbol to mark the route
+      SimpleLineSymbol simpleLineSymbol = new SimpleLineSymbol(SimpleLineSymbol.Style.SOLID, 0x4D0000FF, 5.0f);
+
+      // create a closest facility task from a network analysis service
+      ClosestFacilityTask closestFacilityTask = new ClosestFacilityTask("https://sampleserver6.arcgisonline.com/arcgis/rest/services/NetworkAnalysis/SanDiego/NAServer/ClosestFacility");
+
+      // create a table for facilities using the FeatureServer
```
tschie: FeatureServer -> feature service

src/main/java/com/esri/samples/na/closest_facility_static/ClosestFacilityStaticSample.java
```diff
@@ -0,0 +1,319 @@
...
+                          // display the route on the graphics overlay
+                          graphicsOverlay.getGraphics().add(new Graphic(closestFacilityRoute.getRouteGeometry(), simpleLineSymbol));
+
+                          // hide the progress indicator and enable the reset button
+                          progressIndicator.setVisible(false);
+                          resetButton.setDisable(false);
+                        }
+
+                      } catch (ExecutionException | InterruptedException ex) {
+                        displayMessage("Error getting the ClosestFacilityTask result", ex.getMessage());
```
tschie: Instead of `displayMessage` like we have in older samples, you could do

```java
new Alert(Alert.AlertType.ERROR, "Error getting the closest facility task result").show();
```

The user wouldn't see the stack trace, but in general you would never want to show a stack trace in an alert to an end user anyway.

src/main/java/com/esri/samples/na/closest_facility_static/README.md
```diff
@@ -0,0 +1,61 @@
...
+  <li>Add a list of all facilities to the task parameters: <code>closestFacilityParameters.setFacilities(facilitiesList)</code>.</li>
+  <li>Add a list of all incidents to the task parameters: <code>closestFacilityParameters.setIncidents(incidentsList)</code>.</li>
+  <li>Get <code>ClosestFacilityResult</code> from solving the task with the provided parameters: <code>closestFacilityTask.solveClosestFacilityAsync(closestFacilityParameters)</code>.</li>
+  <li>Find the closest facility for each incident by iterating over the previously created <code>incidentsList</code>:
+    <ul>
+      <li>Get index list of closet facilities to the incident, <code>closestFacilityResult.getRankedFacilityIndexes(indexOfIncident).get(0)</code>.</li>
+      <li>Find closest facility route, <code>closestFacilityResult.getRoute(closestFacilityIndex, indexOfIncident)</code>.</li>
+    </ul>
+  </li> 
+  <li>Display the route to <code>MapView</code>:
```
tschie: How about simply: "Display the route:"

tschie (approved):
Finished my review. Just a few more minor tweaks suggested, then it's good to go.

src/main/java/com/esri/samples/na/closest_facility_static/ClosestFacilityStaticSample.java
JonLavi: Thanks, I've noticed the two different ways samples show errors.
It sounds like the preference is now to do `new Alert(...)`?
What's the effective difference between the two ways, I can't quite tell what the `displayMessage` method is trying to achieve differently?

JonLavi:
Thanks @tschie, good pointers once again.
One last question: I noticed on closing the sample there's an error in the log:
>28 Graphics_hardware_resource or Device objects have not been destructed

It looks like there's a problem somwhere with the Apps' `stop()` method or `mapView.dispose()`?
I've done some digging around but can't seem to find what's causing this. Any ideas?

tschie:
Sorry @JonLavi, I didn't see that last message. You're right, `mapView.dispose()` should take care of that. I would mention it to Colin. I'm not familiar enough with how that method works.

JonLavi:
Merged after discussion with @Rachael-E. Will follow up the bug with the java team.

PR #348, opened 2019-05-14
GetElevationAtAPoint - update screenshot

JonLavi:
I've noticed that the screenshot wasn't updated when I changed the code to only get elevation to the closest metre. Simple fix.


Rachael-E (approved):
looks good to merge !

PR #351, opened 2019-05-30
Allows user to click a value on slider to change exaggeration

Rachael-E:
Made changes that now allows a user to click on the slider to adjust the terrain exaggeration, implemented using a new FXML method to handle onMouseReleased. @tschie could you have a look please? thanks.

### DIFF
--- a/src/main/java/com/esri/samples/scene/terrain_exaggeration/TerrainExaggerationController.java
+++ b/src/main/java/com/esri/samples/scene/terrain_exaggeration/TerrainExaggerationController.java
@@ -51,13 +51,6 @@ public class TerrainExaggerationController {
       surface.getElevationSources().add(new ArcGISTiledElevationSource(elevationImageService));
       scene.setBaseSurface(surface);
 
-      // set exaggeration of surface to the value the user selected
-      exaggerationSlider.valueChangingProperty().addListener(o -> {
-        if (!exaggerationSlider.isValueChanging()) {
-          changeElevationExaggeration();
-        }
-      });
-
       // add a camera and initial camera position
       Point initialLocation = new Point(-119.94891542688772, 46.75792111605992, 0, sceneView.getSpatialReference());
       Camera camera = new Camera(initialLocation, 15000.0, 40.0, 60.0, 0.0);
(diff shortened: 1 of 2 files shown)


src/main/java/com/esri/samples/scene/terrain_exaggeration/TerrainExaggerationController.java
```diff
@@ -67,6 +69,14 @@ public void initialize() {
     }
   }
 
+  @FXML
```
tschie: Annotation should go after javadoc

src/main/java/com/esri/samples/scene/terrain_exaggeration/TerrainExaggerationController.java
```diff
@@ -67,6 +69,14 @@ public void initialize() {
     }
   }
 
+  @FXML
+  /**
+   * Sets the elevation exaggeration to the float chosen via the JavaFX slider.
+   */
+  private void changeElevationExaggeration(){
```
tschie: space before `{`

src/main/java/com/esri/samples/scene/terrain_exaggeration/TerrainExaggerationController.java
```diff
@@ -44,18 +45,19 @@ public void initialize() {
...
-      Surface surface = new Surface();
+      surface = new Surface();
       final String elevationImageService =
-          "http://elevation3d.arcgis.com/arcgis/rest/services/WorldElevation3D/Terrain3D/ImageServer";
+              "http://elevation3d.arcgis.com/arcgis/rest/services/WorldElevation3D/Terrain3D/ImageServer";
       surface.getElevationSources().add(new ArcGISTiledElevationSource(elevationImageService));
       scene.setBaseSurface(surface);
 
       // set exaggeration of surface to the value the user selected
       exaggerationSlider.valueChangingProperty().addListener(o -> {
```
tschie: I don't think this listener is necessary anymore with the event listener you added.

tschie (approved):
Nice! Just a few little tweaks needed. I'm not sure why we used FXML for this one as it's pretty simple. It's okay though if you don't want to refactor it.

Rachael-E:
@tschie thanks! I decided to keep it as an FXML sample. I think it was separated out before because the main point of the sample (`setElevationExaggeration`) would have been drowned in the UX set up of the `Slider`.  At least having the two separated allows the user to identify the relevant API more easily.

PR #354, opened 2019-05-31
Routing around barriers

JonLavi:
Hi @tschie, after @nCastle1 revamped the sample design, I've finalised the java implementation.
Looking forward to getting some feedback once you find some time to have a look.


nCastle1 (approved):
Looks good!

src/main/java/com/esri/samples/na/routing_around_barriers/README.md
```diff
@@ -0,0 +1,91 @@
...
+ In some situations, it is further beneficial to find the most efficient route that reaches all stops, reordering them to reduce travel time. For example, a delivery service may target a number of drop-off addresses, specifically looking to avoid congested areas or closed roads, arranging the stops in the most time-effective order.</p>
+
+<h2>How to use the sample</h2>
+
+<p>Use the Edit Mode toggle buttons to select whether to add Stops or Barriers to the route. Click 'Determine Route' to find the route and display it. Select 'Allow stops to be re-ordered' to find the best sequence. Select 'Preserve first stop' if there is a known start point, and 'Preserve last stop' if there is a known final destination.</p>
+
+<h2>How it works</h2>
+
+<ol>
+  <li>Create the route task by calling <code>RouteTask.CreateAsync(_serviceUrl)</code> with the URL to a Network Analysis route service.</li>
```
Rachael-E: looks like there is a typo on this line, with the underscore before serviceURL used as a parameter. I think that might be used by .NET but we don't use it for Java. There are quite a few more instances of this underscore throughout the readme, they should all be removed.

src/main/java/com/esri/samples/na/routing_around_barriers/README.md
```diff
@@ -0,0 +1,91 @@
...
+      <li>Get the name of the stop by counting the existing stops - <code>_stepsOverlay.Graphics.Count + 1</code>.</li>
+      <li>Create a composite symbol for the stop. This sample uses a pushpin marker and a text symbol.</li>
+      <li>Create the graphic from the geometry and the symbol.</li>
+      <li>Add the graphic to the stops graphics overlay.</li>
+    </ol>
+  </li>
+
+  <li>When the user adds a barrier, create a polygon barrier and add it to the route parameters.
+    <ol>
+      <li>Normalize the geometry (see <strong>3i</strong> above).</li>
```
Rachael-E: I would exchange 3i to 3.1

src/main/java/com/esri/samples/na/routing_around_barriers/README.md
```diff
@@ -0,0 +1,91 @@
...
+  <li>Stop</li>
+</ul>
+
+<h2>About the data</h2>
+
+<p>This sample uses an Esri-hosted sample street network for San Diego.</p>
+
+<h2>Tags</h2>
+
+<p>network analysis, routing, barriers, directions, sequence, best sequence, stops, stop order, maneuver</p>
```
Rachael-E: alphabetical order (I think this readme template got approved before we agreed as a team to make this alphabetical, but may as well change this now)

src/main/resources/fxml/routing_around_barriers.fxml
```diff
@@ -0,0 +1,103 @@
...
+<?import javafx.geometry.Insets?>
+<?import javafx.scene.control.*?>
+<?import javafx.scene.layout.*?>
+<?import javafx.scene.paint.Color?>
+<StackPane fx:controller="com.esri.samples.na.routing_around_barriers.RoutingAroundBarriersController"
+           xmlns:fx="http://javafx.com/fxml"
+           stylesheets="/css/style.css">
+
+  <!--      SDK SceneView-->
+  <MapView fx:id="mapView"/>
```
Rachael-E: looks like the indents could be a little off here? In our other fxml files, the file spacing is 4 after the StackPane tag is set up. Here it is 2.

src/main/java/com/esri/samples/na/routing_around_barriers/RoutingAroundBarriersSample.java
```diff
@@ -0,0 +1,63 @@
...
+
+public class RoutingAroundBarriersSample extends Application {
+
+  private static RoutingAroundBarriersController controller;
+
+  @Override
+  public void start(Stage stage) throws Exception {
+
+      // set up the scene
+      FXMLLoader loader = new FXMLLoader(getClass().getResource("/fxml/routing_around_barriers.fxml"));
```
Rachael-E: looks like there are 4 indents here rather than 2

src/main/resources/fxml/routing_around_barriers.fxml
```diff
@@ -0,0 +1,103 @@
...
+<?import com.esri.arcgisruntime.mapping.view.MapView?>
+<?import javafx.geometry.Insets?>
+<?import javafx.scene.control.*?>
+<?import javafx.scene.layout.*?>
+<?import javafx.scene.paint.Color?>
+<StackPane fx:controller="com.esri.samples.na.routing_around_barriers.RoutingAroundBarriersController"
+           xmlns:fx="http://javafx.com/fxml"
+           stylesheets="/css/style.css">
+
+  <!--      SDK SceneView-->
```
Rachael-E: Remove the unnecessary additional white space before the text of each comment e.g.
`<!-- SDK SceneView -->`

src/main/java/com/esri/samples/na/routing_around_barriers/RoutingAroundBarriersController.java
```diff
@@ -0,0 +1,328 @@
+package com.esri.samples.na.routing_around_barriers;
```
Rachael-E: also need the Esri copyright statement in this file

src/main/java/com/esri/samples/na/routing_around_barriers/RoutingAroundBarriersController.java
```diff
@@ -0,0 +1,328 @@
...
+  @FXML private ToggleButton btnAddBarrier;
+  @FXML private Button btnDetermineRoute;
+  @FXML private Button btnReset;
+  @FXML private CheckBox findBestSequenceCheckBox;
+  @FXML private CheckBox preserveFirstStopCheckBox;
+  @FXML private CheckBox preserveLastStopCheckBox;
+  @FXML private ListView<String> directionsList;
+  @FXML private TitledPane routeInformationTitledPane;
+
+  // for displaying the route to the mapview
```
Rachael-E: This is personal preference, but I'd generally prefer to see these helpful comments within the body of the code rather than when they're being set up as member variables: it's a bit neater and the comment is then in context with the rest of the code using these, rather than having to scroll back up to the top to see what the comment is.

src/main/java/com/esri/samples/na/routing_around_barriers/RoutingAroundBarriersController.java
```diff
@@ -0,0 +1,328 @@
...
+
+    // add the graphics overlays to the map view
+    mapView.getGraphicsOverlays().addAll(Arrays.asList(stopsGraphicsOverlay, barriersGraphicsOverlay, routeGraphicsOverlay));
+
+    // create a list of stops and a list of barriers
+    stopsList = new ArrayList<>();
+    barriersList = new ArrayList<>();
+
+    // initialize the TitlePane of the Accordion box
+    routeInformationTitledPane.setText("No route to display");
```
Rachael-E: this can be set within the FXML file, in the TitledPane tag with `text = "No route to display"`

src/main/java/com/esri/samples/na/routing_around_barriers/RoutingAroundBarriersController.java
```diff
@@ -0,0 +1,328 @@
...
+
+            // enable the reset button
+            btnReset.setDisable(false);
+
+          } else {
+            new Alert(Alert.AlertType.ERROR, "No possible routes found").show();
+          }
+
+        } catch (InterruptedException | ExecutionException e) {
+          new Alert(Alert.AlertType.ERROR, "Solve RouteTask failed " + e.getMessage() + e.getMessage()).show();
```
Rachael-E: I got this message to appear by trying to break the sample. I placed 3 stops in a row, and placed a barrier over the middle stop. This causes this Alert, but then when I go back to the sample, the reset button is disabled. Would be good to add in a condition where if this alert occurs the user can still reset the sample to try again.

src/main/java/com/esri/samples/na/routing_around_barriers/RoutingAroundBarriersController.java
```diff
@@ -0,0 +1,328 @@
...
+    barriersList = new ArrayList<>();
+
+    // initialize the TitlePane of the Accordion box
+    routeInformationTitledPane.setText("No route to display");
+
+    // create symbols for displaying the barriers and the route line
+    SimpleFillSymbol barrierSymbol = new SimpleFillSymbol(SimpleFillSymbol.Style.DIAGONAL_CROSS, 0xFFFF0000, null);
+    routeLineSymbol = new SimpleLineSymbol(SimpleLineSymbol.Style.SOLID, 0x800000FF, 5.0f);
+
+    // disable the UI until ready
```
Rachael-E: these can also be moved to the FXML file, within each button's tag `disable = "true"`

src/main/java/com/esri/samples/na/routing_around_barriers/RoutingAroundBarriersController.java
```diff
@@ -0,0 +1,328 @@
...
+        new Alert(Alert.AlertType.ERROR, "Unable to load RouteTask " + routeTask.getLoadStatus().toString()).show();
+      }
+    });
+
+    // listen to mouse clicks to add/remove stops or barriers
+    mapView.setOnMouseClicked(e -> {
+      // convert clicked point to a map point
+      Point mapPoint = mapView.screenToLocation(new Point2D(e.getX(), e.getY()));
+
+      // if the primary mouse button was clicked, add a stop or barrier, respectively
```
Rachael-E: suggested rewording: "add a stop or barrier to the clicked map point"

Rachael-E (changes requested):
hey @JonLavi this is looking great! I've given this a quick first look over, I've commented on a few things to address.

src/main/java/com/esri/samples/na/routing_around_barriers/README.md
JonLavi: Isn't it more correct to say 3i, since the second-order lists use roman number?

src/main/java/com/esri/samples/na/routing_around_barriers/RoutingAroundBarriersController.java
JonLavi: Good spot. Most of the time, if there's a barrier on a stop, the route will just try and go as close to the stop as it can - I managed to reproduce this, and now the reset button is enabled regardless of whether the route is solved or not. One could always barriers/stops one by one with right-click, but that's really slow, so might as well enable the reset button as soon as the first attempt to solve a route is made.

JonLavi:
Thanks @Rachael-E, excellent comments once again. My FXML editing should be set up correctly now, so hopefully formatting won't be an issue again.

src/main/java/com/esri/samples/na/routing_around_barriers/README.md
Rachael-E: the HTML in your README will render to display 3.1, rather than 3.i (as Github shows it - I don't know why this occurs, different rendering of HTML in Markdown. I use MarkdownPad 2 to do README formatting checks). However, that reminds me there is a bug when rendering ordered lists within ordered lists as this README does, when pushed to our Dev samples page. Replace the nested ordered lists with unordered lists (this will lose the numbering but that's ok - can always say something like "normalize the geometry as in step 3").

JonLavi: Good point. It looks like the HTML Markdown gets rendered either 3.1 or 3.i depending on the system setup! Something to keep in mind.
Did not know about the ordered list bug. Changed all the second order lists to `<ul>` now.

src/main/java/com/esri/samples/na/routing_around_barriers/RoutingAroundBarriersController.java
```diff
@@ -0,0 +1,335 @@
...
+    if (stopsList.size() >= 2) {
+      // clear the previous route from the graphics overlay, if it exists
+      routeGraphicsOverlay.getGraphics().clear();
+
+      // add the existing stops and barriers to the route parameters
+      routeParameters.setStops(stopsList);
+      routeParameters.setPolygonBarriers(barriersList);
+
+      // apply the requested route finding parameters
+      routeParameters.setFindBestSequence(findBestSequenceCheckBox.isSelected());
```
tschie: In the UI, the preserve options are disabled until you check find best sequence. Do those parameters not work independently? If they are independent, all checkboxes should be enabled. If they are not independent, you could just show the two preserve checkboxes enabled and do 
```java
routeParameters.setFindBestSequence(preserveFirstStopCheckBox.isSelected() || preserveLastStopCheckBox.isSelected())
```

src/main/java/com/esri/samples/na/routing_around_barriers/RoutingAroundBarriersController.java
```diff
@@ -0,0 +1,335 @@
...
+    // clear barriers from route parameters and barriers list
+    routeParameters.clearPointBarriers();
+    barriersList.clear();
+
+    // clear the directions list
+    directionsList.getItems().clear();
+
+    // clear all graphics overlays
+    mapView.getGraphicsOverlays().forEach(graphicsOverlay -> graphicsOverlay.getGraphics().clear());
+
```
tschie: The `routeInformationTitledPane` doesn't seem to get cleared on reset.

src/main/java/com/esri/samples/na/routing_around_barriers/RoutingAroundBarriersController.java
```diff
@@ -0,0 +1,335 @@
...
+  private CompositeSymbol createCompositeStopSymbol(Integer stopNumber) {
+    // create a marker with a pin image and position it
+    Image pinImage = new Image(getClass().getResourceAsStream("/symbols/orange_symbol.png"), 0, 40, true, true);
+    PictureMarkerSymbol pinSymbol = new PictureMarkerSymbol(pinImage);
+    pinSymbol.setOffsetY(20);
+    pinSymbol.loadAsync();
+
+    // determine the stop number and create a new label
+    String stopNumberText = ((stopNumber).toString());
+    TextSymbol stopTextSymbol = new TextSymbol(16, stopNumberText, 0xFFFFFFFF, TextSymbol.HorizontalAlignment.CENTER, TextSymbol.VerticalAlignment.BOTTOM);
```
tschie: No need for the variable above, just substitute `stopNumberText` with `stopNumber.toString()` here

src/main/java/com/esri/samples/na/routing_around_barriers/RoutingAroundBarriersController.java
```diff
@@ -0,0 +1,335 @@
...
+
+  /**
+   * Builds a composite symbol out of a pin symbol and a text symbol marking the stop number.
+   *
+   * @param stopNumber The number of the current stop being added
+   * @return A composite symbol to mark the current stop
+   */
+  private CompositeSymbol createCompositeStopSymbol(Integer stopNumber) {
+    // create a marker with a pin image and position it
+    Image pinImage = new Image(getClass().getResourceAsStream("/symbols/orange_symbol.png"), 0, 40, true, true);
```
tschie: Anyway we can reuse this instead of creating it multiple times?

src/main/java/com/esri/samples/na/routing_around_barriers/RoutingAroundBarriersController.java
```diff
@@ -0,0 +1,335 @@
...
+
+            // retrieve route travel time, rounded to the nearest minute
+            String routeTravelTime = ((Long) Math.round(firstRoute.getTravelTime())).toString();
+
+            // retrieve route length (m), convert to km, and round to two decimal points
+            double routeLengthKm = firstRoute.getTotalLength() / 1000;
+            double routeLengthKmRounded = Math.round(routeLengthKm * 100d) / 100d;
+
+            // set the title of the TitledPane to display the information
+            String output = String.format("Route directions: %s min (%s mi)", routeTravelTime, routeLengthKmRounded);
```
tschie:
```java
String.format("Route directions: %d min (%.2f km)", Math.round(travelTime), travelLength / 1000)
```

src/main/java/com/esri/samples/na/routing_around_barriers/RoutingAroundBarriersController.java
```diff
@@ -0,0 +1,335 @@
...
+      } else if (e.getButton() == MouseButton.SECONDARY && e.isStillSincePress()) {
+        // clear the displayed route, if it exists, since it might not be up to date any more
+        routeGraphicsOverlay.getGraphics().clear();
+
+        if (btnAddStop.isSelected()) {
+          // check if there are stops to remove
+          if (!stopsList.isEmpty()) {
+
+            // remove the last stop from the stop list and the graphics overlay
+            stopsList.remove(stopsList.size() - 1);
```
tschie: You could use a `LinkedList` instead and do `stopsList.removeLast()`.

src/main/java/com/esri/samples/na/routing_around_barriers/RoutingAroundBarriersController.java
```diff
@@ -0,0 +1,335 @@
...
+          }
+        });
+
+      } else {
+        new Alert(Alert.AlertType.ERROR, "Unable to load RouteTask " + routeTask.getLoadStatus().toString()).show();
+      }
+    });
+
+    // listen to mouse clicks to add/remove stops or barriers
+    mapView.setOnMouseClicked(e -> {
```
tschie: Could separate this out into another method and reference it in the fxml with
```xml
<MapView fx:id="mapView" onMouseClicked="#handleMapViewClicked"/>
```

src/main/java/com/esri/samples/na/routing_around_barriers/RoutingAroundBarriersController.java
```diff
@@ -0,0 +1,335 @@
...
+        }
+
+        // if the secondary mouse button was clicked, delete the last stop or barrier, respectively
+      } else if (e.getButton() == MouseButton.SECONDARY && e.isStillSincePress()) {
+        // clear the displayed route, if it exists, since it might not be up to date any more
+        routeGraphicsOverlay.getGraphics().clear();
+
+        if (btnAddStop.isSelected()) {
+          // check if there are stops to remove
+          if (!stopsList.isEmpty()) {
```
tschie: Add this condition into the `if` above. Same for the `else-if` branch below.

tschie:
I'll finish my review later when I have time

src/main/java/com/esri/samples/na/routing_around_barriers/RoutingAroundBarriersController.java
JonLavi: Made `Image pinImage` and `PictureMarkerSymbol pinSymbol` into class members

src/main/java/com/esri/samples/na/routing_around_barriers/RoutingAroundBarriersController.java
JonLavi: I'm guessing what you mean is to query the route travel time and length inline, and to show it in km?
In that case, I did:
`String.format("Route directions: %d min (%.2f km)", Math.round(firstRoute.getTravelTime()), firstRoute.getTotalLength() / 1000);`

src/main/java/com/esri/samples/na/routing_around_barriers/RoutingAroundBarriersController.java
JonLavi: I don't think they do? If we ask the API to rearrange the stops (`find best sequence`), it will arrange all of them to make the shortest trip possible. Once we set that flag, the first or last stop aren't guaranteed to be in the same place any more. So after that we should get the option to fore the first or last stop to stay in that position.

I'm not sure if I understand what you are suggesting in the second part? 
Should remove the `findBestSequenceCheckBox`, and replace with just the `preserveFirstStopCheckBox` and `preserveLastStopCheckBox`, always enabled, and the `setFindBestSequence` only set to `true` once any of these boxes are ticked? If that's the case, then I don't think it'll be very clear to the user what is going on when they are ticking either of the boxes. Also, there's always the possibility the user want's the stops rearranged but doesn't insist on the first/last stops being preserved (very rare use case, I think, but it's a part of the API).

src/main/java/com/esri/samples/na/routing_around_barriers/RoutingAroundBarriersController.java
JonLavi: Good call, done!

src/main/java/com/esri/samples/na/routing_around_barriers/RoutingAroundBarriersController.java
tschie: Ok, I just looked at the API Ref. You're right. UI is correct.

src/main/java/com/esri/samples/na/routing_around_barriers/RoutingAroundBarriersSample.java
```diff
@@ -29,35 +29,35 @@
...
+    stage.setHeight(700);
+    stage.setScene(scene);
+    stage.show();
   }
 
+  /**
+   * Stops and releases all resources used in application.
+   */
+  @Override
+  public void stop () { controller.terminate(); }
```
Rachael-E: we should follow normal convention here with a new line and correct indentation for methods, like this:

```
public void stop () {
  controller.terminate();
}
```

Rachael-E (approved):
This is looking really good! Just a small thing to address I noticed with the format for the stop method.

src/main/java/com/esri/samples/na/routing_around_barriers/RoutingAroundBarriersSample.java
JonLavi: Thanks! I thought it was intellij auto-collapsing it when I was looking over it.

src/main/java/com/esri/samples/na/routing_around_barriers/README.md
```diff
@@ -0,0 +1,91 @@
+<h1>Routing around barriers</h1>
+
+<p>Find a route that reaches all stops without crossing any barriers.</p>
```
tschie: Find a route with multiple stops and barriers.

src/main/java/com/esri/samples/na/routing_around_barriers/README.md
```diff
@@ -0,0 +1,91 @@
...
+
+<p><img src="RoutingAroundBarriers.png"/></p>
+
+<h2>Use case</h2>
+
+<p>You can define barriers to avoid unsafe areas, for example flooded roads, when planning the most efficient route to evacuate a hurricane zone. When solving a route, barriers allow you to define portions of the road network that cannot be traversed. You could also use this functionality to plan routes when you know an area will be inaccessible due to a community activity like an organized race or a market night.
+   
+ In some situations, it is further beneficial to find the most efficient route that reaches all stops, reordering them to reduce travel time. For example, a delivery service may target a number of drop-off addresses, specifically looking to avoid congested areas or closed roads, arranging the stops in the most time-effective order.</p>
```
tschie: I don't think we support calculating travel times based on traffic congestion or temporarily closed roads.

src/main/java/com/esri/samples/na/routing_around_barriers/README.md
```diff
@@ -0,0 +1,91 @@
+<h1>Routing around barriers</h1>
+
+<p>Find a route that reaches all stops without crossing any barriers.</p>
+
+<p><img src="RoutingAroundBarriers.png"/></p>
+
+<h2>Use case</h2>
+
+<p>You can define barriers to avoid unsafe areas, for example flooded roads, when planning the most efficient route to evacuate a hurricane zone. When solving a route, barriers allow you to define portions of the road network that cannot be traversed. You could also use this functionality to plan routes when you know an area will be inaccessible due to a community activity like an organized race or a market night.
```
tschie: Let's simplify this to one example:

Delivery trucks need to find efficient routes to multiple stops, taking into account barriers such as traffic accidents or blocked roads. The route may be optimized to reduce overall travel time or travel distance. If stops are added to an existing route, it may be more efficient to reorder the stops to achieve the most efficient route.

tschie: As a general guide, try not to describe how the feature works in this section. Stick to the real-life use case.

src/main/java/com/esri/samples/na/routing_around_barriers/README.md
```diff
@@ -0,0 +1,91 @@
...
+   
+ In some situations, it is further beneficial to find the most efficient route that reaches all stops, reordering them to reduce travel time. For example, a delivery service may target a number of drop-off addresses, specifically looking to avoid congested areas or closed roads, arranging the stops in the most time-effective order.</p>
+
+<h2>How to use the sample</h2>
+
+<p>Use the Edit Mode toggle buttons to select whether to add Stops or Barriers to the route. Click 'Determine Route' to find the route and display it. Select 'Allow stops to be re-ordered' to find the best sequence. Select 'Preserve first stop' if there is a known start point, and 'Preserve last stop' if there is a known final destination.</p>
```
tschie: Select 'Preserve first stop' to preserve the first stop. Select 'Preserve last stop' to preserve the last stop.

tschie: 'Allow stops to be re-ordered' is not an option.

src/main/java/com/esri/samples/na/routing_around_barriers/README.md
```diff
@@ -0,0 +1,91 @@
...
+
+<h2>How it works</h2>
+
+<ol>
+  <li>Create the route task by calling <code>RouteTask.CreateAsync(serviceUrl)</code> with the URL to a Network Analysis route service.</li>
+  <li>Get the default route parameters for the service by calling <code>routeTask.CreateDefaultParametersAsync</code>.</li>
+
+  <li>When the user adds a stop, add it to the route parameters.
+    <ul>
+      <li>Normalize the geometry; otherwise the route job would fail if the user included any stops over the 180th degree meridian.</li>
```
tschie: Is this beyond the scope of this sample?

src/main/java/com/esri/samples/na/routing_around_barriers/README.md
```diff
@@ -0,0 +1,91 @@
...
+
+<p>You can define barriers to avoid unsafe areas, for example flooded roads, when planning the most efficient route to evacuate a hurricane zone. When solving a route, barriers allow you to define portions of the road network that cannot be traversed. You could also use this functionality to plan routes when you know an area will be inaccessible due to a community activity like an organized race or a market night.
+   
+ In some situations, it is further beneficial to find the most efficient route that reaches all stops, reordering them to reduce travel time. For example, a delivery service may target a number of drop-off addresses, specifically looking to avoid congested areas or closed roads, arranging the stops in the most time-effective order.</p>
+
+<h2>How to use the sample</h2>
+
+<p>Use the Edit Mode toggle buttons to select whether to add Stops or Barriers to the route. Click 'Determine Route' to find the route and display it. Select 'Allow stops to be re-ordered' to find the best sequence. Select 'Preserve first stop' if there is a known start point, and 'Preserve last stop' if there is a known final destination.</p>
+
+<h2>How it works</h2>
```
tschie: Don't use sublists. This can be higher level. It should describe how to do routing with options in general. You don't even need to describe how to display the route, just how you get the result geometry.

src/main/java/com/esri/samples/na/routing_around_barriers/README.md
```diff
@@ -0,0 +1,91 @@
...
+
+<ul>
+  <li>DirectionManeuver</li>
+  <li>PolygonBarrier</li>
+  <li>Route</li>
+  <li>Route.DirectionManeuver</li>
+  <li>Route.RouteGeometry</li>
+  <li>RouteParameters.ClearPolygonBarriers</li>
+  <li>RouteParameters.FindBestSequence</li>
+  <li>RouteParameters.PreserveFirstStop</li>
```
tschie: These are methods. We don't list those here.

src/main/java/com/esri/samples/na/routing_around_barriers/RoutingAroundBarriersController.java
```diff
@@ -0,0 +1,331 @@
...
+  private LinkedList<PolygonBarrier> barriersList = new LinkedList<>();
+  private PictureMarkerSymbol pinSymbol;
+  private RouteTask routeTask;
+  private RouteParameters routeParameters;
+  private SimpleFillSymbol barrierSymbol;
+  private SimpleLineSymbol routeLineSymbol;
+
+  @FXML
+  public void initialize() {
+    // create an ArcGISMap with a streets basemap
```
tschie: You can keep the comments light on parts not dealing with the subject of the sample.
```java
// add a map with the streets basemap to the map view, centered on San Diego
ArcGISMap map = new ArcGISMap(Basemap.Type.STREETS, 32.727, -117.1750, 40000);
mapView.setMap(map);
```

src/main/java/com/esri/samples/na/routing_around_barriers/RoutingAroundBarriersController.java
tschie: This can be removed.

src/main/java/com/esri/samples/na/routing_around_barriers/RoutingAroundBarriersController.java
```diff
@@ -0,0 +1,331 @@
...
+          RouteResult routeResult = routeResultFuture.get();
+          if (!routeResult.getRoutes().isEmpty()) {
+            // get the first route result
+            Route firstRoute = routeResult.getRoutes().get(0);
+
+            // create a geometry for this route
+            Geometry routeGeometry = firstRoute.getRouteGeometry();
+
+            // create a graphic for the route and add it to the graphics overlay
+            Graphic routeGraphic = new Graphic(routeGeometry, routeLineSymbol);
```
tschie: Could probably use a renderer since all routes have the same symbol.

src/main/java/com/esri/samples/na/routing_around_barriers/RoutingAroundBarriersController.java
```diff
@@ -0,0 +1,331 @@
...
+  /**
+   * Builds a composite symbol out of a pin symbol and a text symbol marking the stop number.
+   *
+   * @param stopNumber The number of the current stop being added
+   * @return A composite symbol to mark the current stop
+   */
+  private CompositeSymbol createCompositeStopSymbol(Integer stopNumber) {
+    // determine the stop number and create a new label
+    TextSymbol stopTextSymbol = new TextSymbol(16, (stopNumber).toString(), 0xFFFFFFFF, TextSymbol.HorizontalAlignment.CENTER, TextSymbol.VerticalAlignment.BOTTOM);
+    stopTextSymbol.setOffsetY((float) pinImage.getHeight() / 2);
```
tschie: The pin symbol may need an offset too because it is not placed where you click.

tschie:
We should also consider if a 'Determine Route' button is needed here. Could the route just update with every change? Would be easier to see how each additional stop/barrier or option affects the route.

src/main/java/com/esri/samples/na/routing_around_barriers/README.md
JonLavi: Thanks, that would be a good point to make in the readme formatting documentation. My approach so far was to describe the feature in a general, technical way, then move on to explain what use case it may apply to.

src/main/java/com/esri/samples/na/routing_around_barriers/README.md
JonLavi: Changed to `Select 'Find best sequence' to allow stops to be re-ordered in order to find an optimum route.`

src/main/java/com/esri/samples/na/routing_around_barriers/README.md
JonLavi: I've missed out on asking. @nCastle1 made this point during the sample design state. Apparently there is a rare bug that may occur if we do not normalize the geometry. I've noticed no other Java sample does this, so I am wondering whether it doesn't generally affect java?
Happy to remove this if you think it's unnecessary. Maybe we should discuss whether/where this step is necessary in a separate issue.

src/main/java/com/esri/samples/na/routing_around_barriers/README.md
JonLavi: Shortened substantially. Since this readme is fairly recently created, perhaps let's discuss suggesting a change to the design readme.

src/main/java/com/esri/samples/na/routing_around_barriers/RoutingAroundBarriersController.java
JonLavi: Can you share a screenshot/gif perhaps? On my machine it seems to be directly where clicked

98725532-C20D-4510-8EF6-B3C93EB45E48

JonLavi:
Thanks for the suggestions @tschie. I like the idea of calculating the route as soon as there's a change to stops/barriers. I'll test how long the delay is when calclating the route and whether it feels smooth that way, and if so, will change and remove button.

JonLavi:
@tschie, I've tried out your suggestion on having the route calculate and update every time a feature is added/removed. It works really well and the delay is very small, so I think this is a great enhancement to the sample. However I think it would require a little bit of optimizing to the code to pull off cleanly. I don't think we should deviate from the original sample design here too much (the are already  .NET/mac/iOS implementations like this, after all), so perhaps it's worth logging and issue for enhancing this design?
Would be your call then on whether to publish this sample as is currently, and update it once the design enhancement has been accepted, or hold this sample back and only release in the enhanced state.

src/main/java/com/esri/samples/na/routing_around_barriers/RoutingAroundBarriersController.java
JonLavi: That's a great suggestion - should we implement this for Java straight away, or incorporate it into a suggested design enhancement (as discussed below)?

tschie:
@JonLavi Always go with the best implementation of the sample you can for this repo. Then please share the sample at our next review meeting. If others like the enhancement, we can create an issue.

JonLavi:
Changed to update the route every time there's a user interaction. Looking forward to hearing what you think. I feel there might be ways to improve `createRouteAndDisplay()`?
Additionally, there's a few places we call
```
routeInformationTitledPane.setText("No route to display");
directionsList.getItems().clear();
```
My habit would be to put that in a separate method, but not sure whether that's warranted here.

network_analysis/routing-around-barriers/README.md
```diff
@@ -0,0 +1,45 @@
...
+
+Delivery trucks need to find efficient routes to multiple stops, taking into account barriers such as traffic accidents or blocked roads. The route may be optimized to reduce overall travel time or travel distance. If stops are added to an existing route, it may be more efficient to reorder the stops to achieve the most efficient route.
+
+## How to use the sample
+
+Use the Edit Mode toggle buttons to select whether to add Stops or Barriers to the route. Click 'Determine Route' to find the route and display it. Select 'Find best sequence' to allow stops to be re-ordered in order to find an optimum route. Select 'Preserve first stop' to preserve the first stop. Select 'Preserve last stop' to preserve the last stop.
+
+## How it works
+
+1. Create the route task by calling `RouteTask.CreateAsync(serviceUrl)` with the URL to a Network Analysis route service.
```
tschie: That doesn't appear to be a method in the Java API

network_analysis/routing-around-barriers/README.md
```diff
@@ -0,0 +1,45 @@
...
+2. Get the default `RouteParameters` for the service by calling `routeTask.CreateDefaultParametersAsync`.
+3. Create the desired `Stop`s and `PolygonBarrier`s, and add a graphics for these to the graphics overlay.
```
tschie: Combine steps 3 & 4 and remove bit about the graphics and graphics overlay

network_analysis/routing-around-barriers/README.md
```diff
@@ -0,0 +1,45 @@
...
+4. Add the stops and barriers to the route's parameters, `routeParameters.SetStops(routeStops)` and `routeParameters.SetPolygonBarriers(routeBarriers)`.
+5. Set the `ReturnStops` and `ReturnDirections` to `true`.
+6. If the user will accept routes with the stops in any order, set `FindBestSequence` to `true` to find the most optimal route.
+7. If the user has a definite start point, set `PreserveFirstStop` to `true`.
+8. If the user has a definite final destination, set `PreserveLastStop` to `true`.
+9. Call `routeTask.SolveRouteAsync(routeParameters)` to get a `RouteResult`.
+10. Get the first returned route by calling `calculatedRoute.Routes.First()`.
```
tschie: There is no `first` method.

network_analysis/routing-around-barriers/README.md
```diff
@@ -0,0 +1,45 @@
...
+11. Get the geometry from the route as a polyline by accessing the `firstResult.RouteGeometry` property, and use the geometry to create a graphic display the route to the map.
```
tschie: remove the bit about creating a graphic, just tell them how to get the geometry

tschie (approved):
Just a few changes needed with the readme. Good to merge after that.

I added the gradle wrapper jar. Make sure you don't have a gitignore file somewhere that's ignoring the wrapper jar

JonLavi:
Thanks @tschie , I somehow ended up with the .NET version of the readme, after submitting a PR to change the design so that the route is solved on every user interaction.. All fixed and shortened now.

PR #355, opened 2019-06-03
Set local server appDataPath to short temp dir

tschie:
Updates local server services sample to set app data path to short temp dir (required for some geoprocessing services).

### DIFF
--- a/src/main/java/com/esri/samples/localserver/local_server_services/LocalServerServicesController.java
+++ b/src/main/java/com/esri/samples/localserver/local_server_services/LocalServerServicesController.java
@@ -17,6 +17,9 @@
 package com.esri.samples.localserver.local_server_services;
 
 import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
 
 import javafx.application.HostServices;
 import javafx.application.Platform;
@@ -31,6 +34,7 @@ import javafx.scene.control.TextArea;
 import javafx.scene.control.TextField;
 import javafx.stage.FileChooser;
 import javafx.stage.FileChooser.ExtensionFilter;
+import org.apache.commons.io.FileUtils;
 
 import com.esri.arcgisruntime.localserver.LocalFeatureService;
 import com.esri.arcgisruntime.localserver.LocalGeoprocessingService;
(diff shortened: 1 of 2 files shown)


tschie:
@Rachael-E Can you please review? I'll message you on Slack with a way to test it.

src/main/java/com/esri/samples/localserver/local_server_services/LocalServerServicesController.java
```diff
@@ -50,12 +54,25 @@
...
   private FileChooser packageChooser;
+  private Path appDataPath;
 
   @FXML
   private void initialize() {
 
     if (LocalServer.INSTANCE.checkInstallValid()) {
       LocalServer server = LocalServer.INSTANCE;
+
+      // configure app data path (path length must be short for some services)
```
JonLavi, line 65: @Rachael-E and I think it might be worth specifying what 'short' means in this context? Character limit, or depth?

JonLavi (approved):
Hi @tschie, tested and working as expected (with the .gpkx file sitting on my desktop).
Only suggestion might be to clarify in the comment what a 'short' path actually means.

src/main/java/com/esri/samples/localserver/local_server_services/LocalServerServicesController.java
Rachael-E: @JonLavi it looks like this will be covered in the updated doc, so I'd be happy to leave comment as it is.

Rachael-E (changes requested):
@tschie this looks good and the temporary path is a good solution. However, the temporary directory cannot delete unless the service is stopped before the app is closed. If it's not stopped, when the app closes it results in an IOException with `Unable to delete file` (specifically, the lock file of the .gdb file/s used). 

I tried stopping the service/s in the terminate method with `localService.stopAsync()` but that leads to a FileNotFoundException (the same .gdb file).

tschie:
@Rachael-E Interesting, I'm not able to repro it. I can see it running in the browser. I close down the app, no exception, and the directory gets deleted. @JonLavi do you see the same thing?

tschie:
@JonLavi @Rachael-E I think my latest change fixes the issue. Please try it again.

JonLavi:
No more errors for me. Tested with all three services and the provided files in `sample-data\`.

PR #357, opened 2019-06-06
Read symbols from mobile style file

JonLavi:
Hi @tschie and @Rachael-E, I've finished the symbols from mobile style file sample.
There's slight differences to [Android](https://github.com/Esri/arcgis-runtime-samples-android/tree/develop/java/read-symbols-mobile-style-file) and [.NET](https://github.com/Esri/arcgis-runtime-samples-dotnet/pull/633):

- I've stored the `SymbolStyleSearchResult` search result items directly in the `ListView`, meaning there's no need for a helper class or hash map to go between user selection and the list of symbol "keys" used to create the symbol.

- At least .NET have kept the preview `Image` and the creation of the map `Graphic` for the `GrahpicsOverlay` in separate workflows. In this implementation I'm creating the `Graphic` for the `MapView` directly from the preview `Image`. 

Another note: there's quite a bit of setting up of the `CellFactory`s for the `ListView`s in the beginning. Would I be able to move those class definition to the bottom of the file, or would it be better to keep them in line?

Looking forward to feedback, let me know what you think of these two different approaches.


src/main/java/com/esri/samples/symbology/read_symbols_from_mobile_style_file/README.md
```diff
@@ -0,0 +1,45 @@
+# Read symbols from a mobile style
+
+Open a mobile style (.stylx) and read its contents. Combine several symbols from the style into a single multilayer point symbol, then use it to display graphics in the map view.
```
Rachael-E: As per the common README templates, this should be a single short sentence summarising the purpose of the sample. Something like "Read symbols from a mobile style file and combine them to make customized symbology for geoelements". 

(We should update this in the common-samples READMEs too, though I couldn't find this design for now).

src/main/java/com/esri/samples/symbology/read_symbols_from_mobile_style_file/README.md
```diff
@@ -0,0 +1,45 @@
...
+![](ReadSymbolsFromMobileStyleFile.png)
+
+## Use case
+
+In many applications, geoelements are represented by a symbol composed of a number of components. For example, individual elements of a water infrastructure network, such as valves, nodes, or endpoints, will be represented on the map as symbols of the same base shape, with additional components or colors added to represent their unique characteristics, such as technical specifications. 
+
+While each of these symbols can be created from scratch, a more convenient workflow is to author them using ArcGIS Pro and store them in a mobile style file (.stylx). ArcGIS Runtime can read symbols from a mobile style, and you can modify and combine them as needed in your app.
+
+## How to use the sample
+
```
Rachael-E: Our readme template also prefers this as a paragraph, this should be refactored, and "tap" replaced for "click":
 
"Select a symbol and a color from each of the category lists to create an emoji. A preview of the symbol is updated as selections are made. The size of the symbol can be set using the slider. Click the map to create a point graphic using the customized emoji symbol, and click "Reset" to clear all graphics from the display.

src/main/java/com/esri/samples/symbology/read_symbols_from_mobile_style_file/README.md
```diff
@@ -0,0 +1,45 @@
...
+## Relevant API
+
+* MultilayerPointSymbol
+* MultilayerSymbol.CreateSwatchAsync
+* SymbolLayer
+* SymbolStyle
+* SymbolStyle.GetSymbolAsync
+* SymbolStyleSearchParameters
+
+## Offline Data
```
Rachael-E: For Java, we don't access offline data as Android do (it's downloaded and saved locally instead), so this section title isn't appropriate for us in this section.  We should rename this header "About the data".

src/main/java/com/esri/samples/symbology/read_symbols_from_mobile_style_file/README.md
```diff
@@ -0,0 +1,45 @@
...
+
+A mobile style file (created using ArcGIS Pro) provides the symbols used by the sample.
```
Rachael-E: Suggested expansion/rewrite: 

"The mobile style file used in this sample was created using ArcGIS Pro, and is hosted on [ArcGIS Online](https://arcgisruntime.maps.arcgis.com/home/item.html?id=1bd036f221f54a99abc9e46ff3511cbf). It contains symbol layers that can be combined to create emojis."

src/main/java/com/esri/samples/symbology/read_symbols_from_mobile_style_file/README.md
```diff
@@ -0,0 +1,45 @@
...
+
+[Emoji mobile style](https://www.arcgis.com/home/item.html?id=1bd036f221f54a99abc9e46ff3511cbf)
+
+## Tags
+
+advanced symbology, multilayer, mobile style, stylx
```
Rachael-E: alphabetical order

src/main/resources/fxml/read_symbols_from_mobile_style_file.fxml
```diff
@@ -0,0 +1,81 @@
...
+                <ListView fx:id="mouthSelectionListView" maxHeight="120" maxWidth="60"/>
+            </VBox>
+            <VBox>
+                <Label>Color:</Label>
+                <ListView fx:id="colorSelectionListView" maxHeight="120" maxWidth="60"/>
+            </VBox>
+        </HBox>
+        <HBox>
+            <Label>Size:</Label>
+            <Slider fx:id="sizeSlider" min="10" value="50" snapToTicks="true" majorTickUnit="1" minWidth="200" onMouseReleased="#updateSymbolSize"/>
```
Rachael-E: I think using `onMouseDragged` here would make for a better UX: that way, the size of the symbol is updated as the slider is dragged, rather than the mouse button having to be dragged and then released to update the preview.

src/main/java/com/esri/samples/symbology/read_symbols_from_mobile_style_file/ReadSymbolsFromMobileStyleFileController.java
```diff
@@ -0,0 +1,354 @@
...
+import com.esri.arcgisruntime.symbology.ColorUtil;
+import com.esri.arcgisruntime.symbology.MultilayerPointSymbol;
+import com.esri.arcgisruntime.symbology.Symbol;
+import com.esri.arcgisruntime.symbology.SymbolStyle;
+import com.esri.arcgisruntime.symbology.SymbolStyleSearchParameters;
+import com.esri.arcgisruntime.symbology.SymbolStyleSearchResult;
+
+public class ReadSymbolsFromMobileStyleFileController {
+
+  @FXML private MapView mapView = new MapView();
```
Rachael-E: Member variables should only be declared here, not instantiated: we normally instantiate them where they are first used in the code. In this example, the member variables are instantiated in the FXML. So all the `= new` etc can be removed from this section.

src/main/java/com/esri/samples/symbology/read_symbols_from_mobile_style_file/ReadSymbolsFromMobileStyleFileController.java
```diff
@@ -0,0 +1,354 @@
...
+          ListenableFuture<List<SymbolStyleSearchResult>> symbolStyleSearchResultFuture = emojiStyle.searchSymbolsAsync(defaultSearchParameters);
+          symbolStyleSearchResultFuture.addDoneListener(() -> {
+            try {
+
+              // loop through the results and add each item to a list view according to category
+              List<SymbolStyleSearchResult> symbolStyleSearchResults = symbolStyleSearchResultFuture.get();
+              symbolStyleSearchResults.forEach(symbolStyleSearchResult -> {
+
+                // add the SymbolStyleSearchResult object to the correct list for its category
+                switch (symbolStyleSearchResult.getCategory().toLowerCase()) {
```
Rachael-E: why do these have to be lower case? Could make the code more readable here by just returning the categories with `.getCategory()` and updating the cases to `Hat`, `Eyes` etc.

src/main/java/com/esri/samples/symbology/read_symbols_from_mobile_style_file/ReadSymbolsFromMobileStyleFileController.java
```diff
@@ -0,0 +1,354 @@
...
+                  case "hat":
+                    // add the SymbolStyleSearchResult to the list view
```
Rachael-E: These individual comments in each case aren't required, as the comment above the switch statement already says the intent for this code block well.

src/main/java/com/esri/samples/symbology/read_symbols_from_mobile_style_file/ReadSymbolsFromMobileStyleFileController.java
```diff
@@ -0,0 +1,354 @@
...
+
+    // create a graphics overlay and add it to the map
+    graphicsOverlay = new GraphicsOverlay();
+    mapView.getGraphicsOverlays().add(graphicsOverlay);
+
+    // load the available symbols from the style file
+    loadSymbolsFromStyleFile();
+
+    // create a listener that builds the composite symbol when an item from the list view is selected
+    ChangeListener<Object> changeListener = (ObservableValue<?> observable, Object oldValue, Object newValue) -> buildCompositeSymbol();
```
Rachael-E: this can be shortened to: 
`    ChangeListener<Object> changeListener = (obs, oldValue, newValue) -> buildCompositeSymbol();
`

src/main/java/com/esri/samples/symbology/read_symbols_from_mobile_style_file/ReadSymbolsFromMobileStyleFileController.java
```diff
@@ -0,0 +1,354 @@
...
+
+    // create an list of the ListView objects and iterate over it. This is for adding cell factories, listeners, and an empty entry
```
Rachael-E: grammar: create **a** list

Rachael-E: another suggestion for comments. Since you have comments later for adding in cell factories etc, can remove mentions of them up here to avoid repetition.

src/main/java/com/esri/samples/symbology/read_symbols_from_mobile_style_file/ReadSymbolsFromMobileStyleFileController.java
```diff
@@ -0,0 +1,354 @@
...
+      // add the cell factory to show the symbol within the list view
+      listView.setCellFactory(c -> new SymbolLayerInfoListCell());
+      // add the change listener to rebuild the preview when a selection is made
+      listView.getSelectionModel().selectedItemProperty().addListener(changeListener);
+      // add an empty entry to the list view to allow selecting 'nothing', and make it selected by default
+      listView.getItems().add(null);
+      listView.getSelectionModel().select(0);
+    }
+
+    // add colors to the color selection list view. We require the 0xAARRGGBB format to color the symbols
```
Rachael-E: it's obvious from the code what the format is for color, so can remove comment about that here.

src/main/java/com/esri/samples/symbology/read_symbols_from_mobile_style_file/ReadSymbolsFromMobileStyleFileController.java
```diff
@@ -0,0 +1,354 @@
...
+          });
+        } catch (InterruptedException | ExecutionException e) {
+          new Alert(Alert.AlertType.ERROR, "Error retrieving default search parameters for symbol search" + e.getMessage()).show();
+        }
+      });
+    });
+  }
+
+  /**
+   * Gets the keys of the selected symbol components, and constructs a multilayer point symbol
```
Rachael-E: Javadoc comments should have full stops at the end of each sentence

src/main/java/com/esri/samples/symbology/read_symbols_from_mobile_style_file/ReadSymbolsFromMobileStyleFileController.java
```diff
@@ -0,0 +1,354 @@
...
+   */
+  @FXML
+  private void buildCompositeSymbol() {
+
+    // retrieve the requested key for the requested hat symbol
```
Rachael-E: there is a lot of repetition for these 3 blocks of code relating to SymbolStyleSearchResult. Could they be handled in a more DRY way by putting them in a method?

src/main/java/com/esri/samples/symbology/read_symbols_from_mobile_style_file/ReadSymbolsFromMobileStyleFileController.java
```diff
@@ -0,0 +1,354 @@
...
+        updateSymbolPreview(faceSymbol);
+
+      } catch (ExecutionException | InterruptedException e) {
+        new Alert(Alert.AlertType.ERROR, "Error creating symbol with the provided symbol keys" + e.getMessage()).show();
+      }
+    });
+  }
+
+  /**
+   * Creates an ImageView object from a provided symbol and displays it in the preview area
```
Rachael-E: javadoc comment, full stop at end of sentence

src/main/java/com/esri/samples/symbology/read_symbols_from_mobile_style_file/ReadSymbolsFromMobileStyleFileController.java
```diff
@@ -0,0 +1,354 @@
...
+      // display the image view in the preview area
+      symbolPreviewHBox.getChildren().add(symbolImageView);
+
+    } catch (InterruptedException | ExecutionException e) {
+      new Alert(Alert.AlertType.ERROR, "Error creating preview ImageView from provided MultilayerPointSymbol" + e.getMessage()).show();
+    }
+  }
+
+  /**
+   * Updates the size of the symbol preview and the label when the size bar is interacted with.
```
Rachael-E: size bar -> slider

src/main/java/com/esri/samples/symbology/read_symbols_from_mobile_style_file/ReadSymbolsFromMobileStyleFileController.java
```diff
@@ -0,0 +1,354 @@
...
+      graphicsOverlay.getGraphics().add(graphic);
+    }
+  }
+
+  /**
+   * Clears all the graphics from the graphics overlay.
+   */
+  @FXML
+  private void resetView() {
+    mapView.getGraphicsOverlays().get(0).getGraphics().clear();
```
Rachael-E: this can be replaced with `    graphicsOverlay.getGraphics().clear();
`

src/main/resources/fxml/read_symbols_from_mobile_style_file.fxml
```diff
@@ -0,0 +1,81 @@
...
+<?import javafx.scene.layout.BackgroundFill?>
+<?import javafx.scene.layout.HBox?>
+<?import javafx.scene.layout.StackPane?>
+<?import javafx.scene.layout.VBox?>
+<?import javafx.scene.paint.Color?>
+<?import com.esri.arcgisruntime.mapping.view.MapView?>
+<?import javafx.scene.control.Label?>
+<?import javafx.scene.control.Button?>
+<StackPane
+        fx:controller="com.esri.samples.symbology.read_symbols_from_mobile_style_file.ReadSymbolsFromMobileStyleFileController"
```
Rachael-E: this line beginning fx:controller should be on the same line as `<StackPane`

src/main/resources/fxml/read_symbols_from_mobile_style_file.fxml
```diff
@@ -0,0 +1,81 @@
...
+        xmlns:fx="http://javafx.com/fxml" stylesheets="/css/style.css">
+    <!--SDK SceneView-->
```
Rachael-E: this comment should refer to MapView, not SceneView

src/main/resources/fxml/read_symbols_from_mobile_style_file.fxml
Rachael-E: I don't think we need snapToTicks set here.

Rachael-E (changes requested):
good job @JonLavi, a few more changes required. When writing comments, bear in mind what comments follow or what has already written before to avoid repetition: I've made a few suggestions here to help you avoid that.  

As a general comment for the application, I'd prefer to see less strong colours for the emojis. The blue and red especially are quite intense and clash with the symbol layers (e.g. red face and red eyes, blue is very dark and difficult to pick out features). Would suggest making the red and blue slightly more pastelly for a better visual UI experience.

src/main/java/com/esri/samples/symbology/read_symbols_from_mobile_style_file/README.md
JonLavi: I see! I thought this section was mandatory whenever any offline data is included in the sample, hence why I put a link in, but no explanation as to how it's integrated.

src/main/java/com/esri/samples/symbology/read_symbols_from_mobile_style_file/README.md
JonLavi: Good spot - I should practice my abc's 😆

src/main/resources/fxml/read_symbols_from_mobile_style_file.fxml
JonLavi: While 'onMouseDragged' gives you a live preview when dragging, it means the symbol doesn't resize when you simply click anywhere on the slider, which means it's kind of broken. I'll investigate if there is any option to get both of these functionalities together.

Re 'snapToTicks': absolutely right, I had it in there before I rounded the number displayed in the label, which helped avoid getting awkward pixel sizes with many digits after the decimal.

src/main/java/com/esri/samples/symbology/read_symbols_from_mobile_style_file/ReadSymbolsFromMobileStyleFileController.java
JonLavi: I took over this from the original design. I suppose in theory they don't have to be made lower case, if we know exactly how the categories are formatted in the stylx file. 
I'd say it's a way to sanitize the data? In this case, it would catch `HAT`, `Hat`, `HaT`, etc.
Not entirely necessary here as we know the file is set up neatly, but I think it's considered good practice?

Long story short, I removed it, since it's technically unnecessary here.

JonLavi: I took over this from the original design. I suppose in theory they don't have to be made lower case, if we know exactly how the categories are formatted in the stylx file. 
I'd say it's a way to sanitize the data? In this case, it would catch `HAT`, `Hat`, `HaT`, etc.
Not entirely necessary here as we know the file is set up neatly, but I think it's considered good practice?

Long story short, I removed it, since it's technically unnecessary here.

src/main/java/com/esri/samples/symbology/read_symbols_from_mobile_style_file/ReadSymbolsFromMobileStyleFileController.java
JonLavi: Great spot! This has bugged me for a while but I was blind as to how to shorten it.

src/main/java/com/esri/samples/symbology/read_symbols_from_mobile_style_file/ReadSymbolsFromMobileStyleFileController.java
JonLavi: Absolutely. Done as you suggested. Not too happy with the Javadoc comment as of yet, so might still change the wording.

JonLavi:
Thanks @Rachael-E, all improvements added. Only the slider interaction left to improve somehow.

src/main/resources/fxml/read_symbols_from_mobile_style_file.fxml
Rachael-E: ah, good point re mouse events. I've been looking into it, the only way I can see past that is to add both `onMouseDragged="#updateSymbolSize" onMouseReleased="#updateSymbolSize"` to the fxml file, and that should handle all eventualities.

src/main/java/com/esri/samples/symbology/read_symbols_from_mobile_style_file/ReadSymbolsFromMobileStyleFileSample.java
```diff
@@ -0,0 +1,63 @@
...
+    stage.setWidth(800);
+    stage.setHeight(700);
+    stage.setScene(scene);
+    stage.show();
+  }
+
+  /**
+   * Stops and releases all resources used in application.
+   */
+  @Override
```
Rachael-E: the `stop` method should always be the last method in the launcher class, for consistency with our other samples.

Rachael-E (approved):
great changes, thanks for that @JonLavi ! I added another comment for handling the slider interactions in the conversation above.

src/main/java/com/esri/samples/symbology/read_symbols_from_mobile_style_file/ReadSymbolsFromMobileStyleFileController.java
```diff
@@ -0,0 +1,353 @@
...
+  @FXML private ListView<SymbolStyleSearchResult> eyesSelectionListView;
+  @FXML private ListView<SymbolStyleSearchResult> mouthSelectionListView;
+  @FXML private ListView<Integer> colorSelectionListView;
+  @FXML private Slider sizeSlider;
+  @FXML private Label sizeLabel;
+  @FXML private HBox symbolPreviewHBox;
+
+  private GraphicsOverlay graphicsOverlay;
+  private MultilayerPointSymbol faceSymbol;
+  private SymbolStyle emojiStyle;
```
alan-edi: You'll know much better than me what's normal in the samples, but in the API code we use a prefix of `m` for class member fields (to help distinguish them from local variables).

I looked at one sample, picked at random, to see what's done there and it follows that convention too: https://developers.arcgis.com/android/latest/java/sample-code/portal-user-info/

However, feel free to stick with what you've got if that's normal in the samples.

src/main/java/com/esri/samples/symbology/read_symbols_from_mobile_style_file/ReadSymbolsFromMobileStyleFileController.java
```diff
@@ -0,0 +1,353 @@
...
+   */
+  @FXML
+  private void buildCompositeSymbol() {
+
+    // retrieve the requested keys for the selected hat, eyes and mouth symbols
+    String hatKey = getRequestedSymbolKey(hatSelectionListView);
+    String eyesKey = getRequestedSymbolKey(eyesSelectionListView);
+    String mouthKey = getRequestedSymbolKey(mouthSelectionListView);
+
+    // create a list of keys to use for getting the requested symbol.
```
alan-edi: Other single-line comments don't have a full stop - **very** good consistency apart from this one I think :)

src/main/java/com/esri/samples/symbology/read_symbols_from_mobile_style_file/ReadSymbolsFromMobileStyleFileController.java
```diff
@@ -0,0 +1,353 @@
...
+      try {
+        faceSymbol = (MultilayerPointSymbol) symbolFuture.get();
+        if (faceSymbol == null) {
+          return;
+        }
+
+        // loop through all the symbol layers and lock the color
+        faceSymbol.getSymbolLayers().forEach(symbolLayer -> symbolLayer.setColorLocked(true));
+
+        // unlock the color of the base layer. Changing the symbol layer will now only change this layer's color
```
alan-edi: Do you mean the following?

> Changing the symbol layer **color** will...

src/main/java/com/esri/samples/symbology/read_symbols_from_mobile_style_file/ReadSymbolsFromMobileStyleFileController.java
```diff
@@ -0,0 +1,353 @@
...
+          });
+        } catch (InterruptedException | ExecutionException e) {
+          new Alert(Alert.AlertType.ERROR, "Error retrieving default search parameters for symbol search" + e.getMessage()).show();
+        }
+      });
+    });
+  }
+
+  /**
+   * Gets the keys of the selected symbol components, and constructs a multilayer point symbol.
```
alan-edi: ...and updates the symbol preview

src/main/java/com/esri/samples/symbology/read_symbols_from_mobile_style_file/ReadSymbolsFromMobileStyleFileController.java
```diff
@@ -0,0 +1,353 @@
...
+
+      } catch (ExecutionException | InterruptedException e) {
+        new Alert(Alert.AlertType.ERROR, "Error creating symbol with the provided symbol keys" + e.getMessage()).show();
+      }
+    });
+  }
+
+  /**
+   * Returns the symbol key of the symbol style search result that is currently selected in the list view.
+   * @param listView the list view of which to query the selected item
```
alan-edi: I think it's normal (at least it is in our API code) to have a blank line before the tags, like you have in `updateSymbolPreview` below

src/main/java/com/esri/samples/symbology/read_symbols_from_mobile_style_file/ReadSymbolsFromMobileStyleFileController.java
```diff
@@ -0,0 +1,353 @@
...
+   * @param multilayerPointSymbol the symbol used to create the image view
+   */
+  private void updateSymbolPreview(MultilayerPointSymbol multilayerPointSymbol) {
+    // remove the previously displayed image view
+    symbolPreviewHBox.getChildren().clear();
+
+    // create an image and image view from the symbol
+    ListenableFuture<Image> symbolImageFuture = multilayerPointSymbol.createSwatchAsync(0x00000000, 1);
+    try {
+      Image symbolImage = symbolImageFuture.get();
```
alan-edi: Since you're not using a DoneListener, the `createSwatchAsync` will run on the main thread. Is that deliberate? Probably best to promote the pattern of using DoneListeners for all async operations?

src/main/java/com/esri/samples/symbology/read_symbols_from_mobile_style_file/ReadSymbolsFromMobileStyleFileController.java
```diff
@@ -0,0 +1,353 @@
...
+
+    if (mapView != null) {
+      mapView.dispose();
+    }
+  }
+
+  /**
+   * Shows the available symbol of the SymbolStyleSearchResult in the symbol selection list view.
+   */
+  class SymbolLayerInfoListCell extends ListCell<SymbolStyleSearchResult> {
```
alan-edi: Declare the class as `private`?

src/main/java/com/esri/samples/symbology/read_symbols_from_mobile_style_file/ReadSymbolsFromMobileStyleFileController.java
```diff
@@ -0,0 +1,353 @@
...
+
+    private SymbolLayerInfoListCell() {
+      // set the cell to display only a graphic
+      setContentDisplay(ContentDisplay.GRAPHIC_ONLY);
+    }
+
+    protected void updateItem(SymbolStyleSearchResult item, boolean empty) {
```
alan-edi: ` @Override` ?

src/main/java/com/esri/samples/symbology/read_symbols_from_mobile_style_file/ReadSymbolsFromMobileStyleFileController.java
```diff
@@ -0,0 +1,353 @@
...
+    private final Rectangle rectangle;
+
+    private ColorListCell() {
+      // set the cell to display only a graphic
+      setContentDisplay(ContentDisplay.GRAPHIC_ONLY);
+      // create a rectangle to display in the cell
+      rectangle = new Rectangle(10, 10);
+    }
+
+    protected void updateItem(Integer item, boolean empty) {
```
alan-edi: ` @Override` ?

src/main/java/com/esri/samples/symbology/read_symbols_from_mobile_style_file/ReadSymbolsFromMobileStyleFileController.java
```diff
@@ -0,0 +1,353 @@
...
+    protected void updateItem(SymbolStyleSearchResult item, boolean empty) {
+      super.updateItem(item, empty);
+
+      if (item == null || empty) {
+        // if the item in the list view is an empty item, show nothing
+        setGraphic(null);
+      } else {
+        // get the symbol from the list view entry, and create an image from it
+        ListenableFuture<Image> symbolImageFuture = item.getSymbol().createSwatchAsync(0x00000000, 1);
+        try {
```
alan-edi: Use a DoneListener?

alan-edi (commented):
@JonLavi Looks very good! Just a few comments

src/main/java/com/esri/samples/symbology/read_symbols_from_mobile_style_file/README.md
```diff
@@ -0,0 +1,40 @@
+# Read symbols from a mobile style
+
+Read symbols from a mobile style file and combine them to make customized symbology for geoelements.
```
tschie: Combine multiple symbols from a mobile style file into a single symbol.

src/main/java/com/esri/samples/symbology/read_symbols_from_mobile_style_file/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+![](ReadSymbolsFromMobileStyleFile.png)
+
+## Use case
+
+In many applications, geoelements are represented by a symbol composed of a number of components. For example, individual elements of a water infrastructure network, such as valves, nodes, or endpoints, will be represented on the map as symbols of the same base shape, with additional components or colors added to represent their unique characteristics, such as technical specifications. 
+
+While each of these symbols can be created from scratch, a more convenient workflow is to author them using ArcGIS Pro and store them in a mobile style file (.stylx). ArcGIS Runtime can read symbols from a mobile style, and you can modify and combine them as needed in your app.
```
tschie: Sounds like this belongs in the Additional Information section.

src/main/java/com/esri/samples/symbology/read_symbols_from_mobile_style_file/README.md
```diff
@@ -0,0 +1,40 @@
+# Read symbols from a mobile style
+
+Read symbols from a mobile style file and combine them to make customized symbology for geoelements.
+
+![](ReadSymbolsFromMobileStyleFile.png)
+
+## Use case
+
+In many applications, geoelements are represented by a symbol composed of a number of components. For example, individual elements of a water infrastructure network, such as valves, nodes, or endpoints, will be represented on the map as symbols of the same base shape, with additional components or colors added to represent their unique characteristics, such as technical specifications.
```
tschie: Revise without the first sentence.

src/main/java/com/esri/samples/symbology/read_symbols_from_mobile_style_file/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+## How to use the sample
+
+Select a symbol and a color from each of the category lists to create an emoji. A preview of the symbol is updated as selections are made. The size of the symbol can be set using the slider. Click the map to create a point graphic using the customized emoji symbol, and click "Reset" to clear all graphics from the display.
+
+## How it works
+
+1. Create a new `SymbolStyle` from a stylx file, and load it: `symbolStyle.loadAsync()`.
+2. Get a set of default search parameters, `symbolStyle.getDefaultSearchParametersAsync()`, and use these to retrieve a list of all symbols within the style file: `symbolStyle.searchSymbolsAsync(defaultSearchParameters)`.
+3. Iterate the list of `SymbolStyleSearchResult` and add symbols to list boxes according to their category. Display a preview of each symbol with `multilayerPointSymbol.createSwatchAsync()`.
```
tschie: The details about list boxes, displaying the preview, and emojis are sample-specific. This section should describe the workflow or creating multilayer symbols from a mobile style _generically_.

src/main/java/com/esri/samples/symbology/read_symbols_from_mobile_style_file/README.md
```diff
@@ -0,0 +1,40 @@
...
+4. When symbol selections change, create a new multilayer symbol by passing the keys for the selected symbols into `symbolStyle.emojiStyle.getSymbolAsync(symbolKeys)`. To only allow coloring in the emoji face, color lock all symbol layers except the base layer and update the current symbol preview image.
+5. Create graphics symbolized with the current symbol when the user clicks on the map view.
```
tschie: Say "when the map view is clicked" rather than when the "user clicks on the map." (just for reference - this step isn't needed due to the point I made above).

src/main/java/com/esri/samples/symbology/read_symbols_from_mobile_style_file/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+## Relevant API
+
+* MultilayerPointSymbol
+* MultilayerSymbol.CreateSwatchAsync
```
tschie: No methods here

src/main/java/com/esri/samples/symbology/read_symbols_from_mobile_style_file/ReadSymbolsFromMobileStyleFileController.java
JonLavi: Good spot - the Android samples follow this convention, but the Java SE samples don't put the `m` prefix.

src/main/java/com/esri/samples/symbology/read_symbols_from_mobile_style_file/ReadSymbolsFromMobileStyleFileController.java
JonLavi: Great spot, thanks. There's always one or two that sneak in!

src/main/java/com/esri/samples/symbology/read_symbols_from_mobile_style_file/ReadSymbolsFromMobileStyleFileController.java
JonLavi: Possibly just an oversight on my part, or in the sample design. Added the `doneListener`, as you suggested.

src/main/java/com/esri/samples/symbology/read_symbols_from_mobile_style_file/ReadSymbolsFromMobileStyleFileController.java
JonLavi: Other uses of this pattern in our samples don't do an `@Override` here. What are the implication of leaving it out?

src/main/java/com/esri/samples/symbology/read_symbols_from_mobile_style_file/ReadSymbolsFromMobileStyleFileController.java
JonLavi: On testing this, it seems that running the `createSwatchAsync` on the main thread might have been deliberate. Putting the block in a `doneListener` now makes the symbol preview flicker and show graphical bugs. I'll do some invesitgating on why.

src/main/java/com/esri/samples/symbology/read_symbols_from_mobile_style_file/ReadSymbolsFromMobileStyleFileController.java
JonLavi: Other uses of this pattern in our samples don't do an @Override here. What are the implication of leaving it out?

src/main/java/com/esri/samples/symbology/read_symbols_from_mobile_style_file/ReadSymbolsFromMobileStyleFileController.java
JonLavi: As with the above, putting this in a `doneListener` seems to cause some graphical bugs in the `ListView` when selecting different symbols. Not sure whether this is a JavaFX issue or a runtime issue though.

src/main/java/com/esri/samples/symbology/read_symbols_from_mobile_style_file/ReadSymbolsFromMobileStyleFileController.java
alan-edi: I'm used to seeing compiler warnings if it's missing, but that will depend on how the IDE is set up re inspections/warnings.

More importantly, it allows the compiler to check if you're really overriding as you intended. There's a description of how that helps in the first answer in https://stackoverflow.com/questions/4736212/the-meaning-of-override-in-android-studio

colinanderson: It is in Effective Java - Item 36 "Consistently use Override annotation". I don't think it meant be consistent about *not* using it!

JonLavi: Thanks both, no errors popped up on my side, but I see how it's good practice to `@Override`. 
Will probably need to go through our samples to see if it's missing anywhere else.

JonLavi:
@tschie `How it works` section updated. Let me know what you think. I'll raise a point for the team that this particular Readme needs a review.

symbology/read-symbols-from-mobile-style-file/src/main/java/com/esri/samples/read_symbols_from_mobile_style_file/ReadSymbolsFromMobileStyleFileController.java
```diff
@@ -0,0 +1,356 @@
...
+
+    if (mapView != null) {
+      mapView.dispose();
+    }
+  }
+
+  /**
+   * Shows the available symbol of the SymbolStyleSearchResult in the symbol selection list view.
+   */
+  private class SymbolLayerInfoListCell extends ListCell<SymbolStyleSearchResult> {
```
tschie: These private classes can go in their own files

symbology/read-symbols-from-mobile-style-file/src/main/resources/read_symbols_from_mobile_style_file.fxml
```diff
@@ -0,0 +1,80 @@
...
+        <HBox>
+            <Label>Size:</Label>
+            <Slider fx:id="sizeSlider" min="10" value="50" majorTickUnit="1" minWidth="200" onMouseDragged="#updateSymbolSize" onMouseReleased="#updateSymbolSize"/>
+            <Label fx:id="sizeLabel" text="50 px"/>
+        </HBox>
+        <HBox alignment="CENTER_LEFT">
+            <Label>Preview:</Label>
+            <HBox alignment="CENTER" fx:id="symbolPreviewHBox" minHeight="140" minWidth="140"/>
+        </HBox>
+        <Button text="Reset" onAction="#resetView" minWidth="80"/>
```
tschie: 'Reset' makes me thing the symbol preview will be reset. I think 'Clear' would be more appropriate.

symbology/read-symbols-from-mobile-style-file/src/main/java/com/esri/samples/read_symbols_from_mobile_style_file/ReadSymbolsFromMobileStyleFileController.java
```diff
@@ -0,0 +1,356 @@
...
+    @Override
+    protected void updateItem(SymbolStyleSearchResult item, boolean empty) {
+      super.updateItem(item, empty);
+
+      if (item == null || empty) {
+        // if the item in the list view is an empty item, show nothing
+        setGraphic(null);
+      } else {
+        // get the symbol from the list view entry, and create an image from it
+        ListenableFuture<Image> symbolImageFuture = item.getSymbol().createSwatchAsync(0x00000000, 1);
```
tschie: I think you'd want a `doneListener` here, otherwise there's no point in doing this asynchronously.

symbology/read-symbols-from-mobile-style-file/README.md
```diff
@@ -0,0 +1,41 @@
...
+1. Create a new `SymbolStyle` from a stylx file, and load it: `symbolStyle.loadAsync()`.
+2. Get a set of default search parameters, `symbolStyle.getDefaultSearchParametersAsync()`, and use these to retrieve a list of all symbols within the style file: `symbolStyle.searchSymbolsAsync(defaultSearchParameters)`.
+3. Get the `SymbolStyleSearchResult`, which contains the symbols, as well as their names, keys, and categories.
+4. Use a `List` of keys of the desired symbols to build a composite symbol, `symbolStyle.emojiStyle.getSymbolAsync(symbolKeys)`.
+5. Get the resulting `MultilayerPointSymbol`, use it to create a `Graphic` and display it to the map.
+
+## Relevant API
+
+* MultilayerPointSymbol
+* MultilayerSymbol.CreateSwatchAsync
```
tschie: No methods

symbology/read-symbols-from-mobile-style-file/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+Individual elements of a water infrastructure network, such as valves, nodes, or endpoints, can be represented on the map as symbols of the same base shape, with additional components or colors added to represent their unique characteristics, such as technical specifications.
+
+## How to use the sample
+
+Select a symbol and a color from each of the category lists to create an emoji. A preview of the symbol is updated as selections are made. The size of the symbol can be set using the slider. Click the map to create a point graphic using the customized emoji symbol, and click "Reset" to clear all graphics from the display.
+
+## How it works
+
+1. Create a new `SymbolStyle` from a stylx file, and load it: `symbolStyle.loadAsync()`.
```
tschie: Can remove the code snippet for loading.

symbology/read-symbols-from-mobile-style-file/README.md
```diff
@@ -0,0 +1,41 @@
...
+2. Get a set of default search parameters, `symbolStyle.getDefaultSearchParametersAsync()`, and use these to retrieve a list of all symbols within the style file: `symbolStyle.searchSymbolsAsync(defaultSearchParameters)`.
+3. Get the `SymbolStyleSearchResult`, which contains the symbols, as well as their names, keys, and categories.
+4. Use a `List` of keys of the desired symbols to build a composite symbol, `symbolStyle.emojiStyle.getSymbolAsync(symbolKeys)`.
```
tschie: The code snippet doesn't appear to be correct

symbology/read-symbols-from-mobile-style-file/README.md
```diff
@@ -0,0 +1,41 @@
...
+5. Get the resulting `MultilayerPointSymbol`, use it to create a `Graphic` and display it to the map.
```
tschie: Create a `Graphic` using the `MultilayerPointSymbol`.

symbology/read-symbols-from-mobile-style-file/src/main/java/com/esri/samples/read_symbols_from_mobile_style_file/ReadSymbolsFromMobileStyleFileController.java
```diff
@@ -0,0 +1,356 @@
...
+    // remove the previously displayed image view
+    symbolPreviewHBox.getChildren().clear();
+
+    // create an image and image view from the symbol
+    ListenableFuture<Image> symbolImageFuture = multilayerPointSymbol.createSwatchAsync(0x00000000, 1);
+    try {
+      Image symbolImage = symbolImageFuture.get();
+      ImageView symbolImageView = new ImageView(symbolImage);
+      // display the image view in the preview area
+      symbolPreviewHBox.getChildren().add(symbolImageView);
```
tschie: Could you reference an `ImageView` in the FXML and just change its image?

symbology/read-symbols-from-mobile-style-file/src/main/java/com/esri/samples/read_symbols_from_mobile_style_file/ReadSymbolsFromMobileStyleFileController.java
```diff
@@ -0,0 +1,356 @@
...
+    }
+  }
+
+  /**
+   * Updates the size of the symbol preview and the label when the slider is interacted with.
+   */
+  @FXML
+  private void updateSymbolSize() {
+    // get the slider value and convert to a string
+    String sizeValue = Integer.toString((int) sizeSlider.getValue());
```
tschie: Can you just add ticks to the slider?

symbology/read-symbols-from-mobile-style-file/src/main/java/com/esri/samples/read_symbols_from_mobile_style_file/ReadSymbolsFromMobileStyleFileController.java
```diff
@@ -0,0 +1,356 @@
...
+                  case "Hat":
+                    hatSelectionListView.getItems().add(symbolStyleSearchResult);
+                    break;
+                  case "Eyes":
+                    eyesSelectionListView.getItems().add(symbolStyleSearchResult);
+                    break;
+                  case "Mouth":
+                    mouthSelectionListView.getItems().add(symbolStyleSearchResult);
+                    break;
+                  default:
```
tschie: This isn't needed

tschie (changes requested):

symbology/read-symbols-from-mobile-style-file/src/main/java/com/esri/samples/read_symbols_from_mobile_style_file/ReadSymbolsFromMobileStyleFileController.java
JonLavi: Yes! Works really nicely.

symbology/read-symbols-from-mobile-style-file/src/main/java/com/esri/samples/read_symbols_from_mobile_style_file/ReadSymbolsFromMobileStyleFileController.java
JonLavi: Can do. Removed that method in the process.

symbology/read-symbols-from-mobile-style-file/src/main/java/com/esri/samples/read_symbols_from_mobile_style_file/ReadSymbolsFromMobileStyleFileController.java
JonLavi: I have added the `doneListener` but I think it adds a slightly weird UI bug (see gif).
A symbol appears under `eyes`, and when selecting a different one it disappears and leaves the 'empty' box, what it should have been before.
3C2AAE7E-84BF-43BD-A389-C07B93A29F76

tschie: You don't want to set the cell's graphic asynchronously. Try setting it to an empty `ImageView`, then update the view's image when the swatch is ready.

```java
ImageView symbolImageView = new ImageView();
setGraphic(symbolImageView);
ListenableFuture<Image> symbolImageFuture = item.getSymbol().createSwatchAsync(0x00000000, 1);
symbolImageFuture.addDoneListener(() -> {
    try {
          // get the symbol swatch
          Image symbolImage = symbolImageFuture.get();
          // update the image view with the symbol swatch
          symbolImageView.setImage(symbolImage);
    } catch (InterruptedException | ExecutionException e) {
          new Alert(Alert.AlertType.ERROR, "Error creating preview image for symbol in mobile style file" + e.getMessage()).show();
    }
});
```

symbology/read-symbols-from-mobile-style-file/src/main/java/com/esri/samples/read_symbols_from_mobile_style_file/ReadSymbolsFromMobileStyleFileController.java
```diff
@@ -0,0 +1,272 @@
...
+      listView.setCellFactory(c -> new SymbolLayerInfoListCell());
+      // add the change listener to rebuild the preview when a selection is made
+      listView.getSelectionModel().selectedItemProperty().addListener(changeListener);
+      // add an empty entry to the list view to allow selecting 'nothing', and make it selected by default
+      listView.getItems().add(null);
+      listView.getSelectionModel().select(0);
+    }
+
+    // add colors to the color selection list view
+    colorSelectionListView.getItems().addAll(0xFFFFFF00, 0xFF00FF00, 0xFFFFB6C1);
```
tschie: Can we use a `ColorPicker` instead? Perhaps place it between the lists and the slider?

symbology/read-symbols-from-mobile-style-file/src/main/java/com/esri/samples/read_symbols_from_mobile_style_file/ReadSymbolsFromMobileStyleFileController.java
JonLavi: Thanks for that! I see how it fixes it, but obviously didn't think about this causing the issue to begin with.

symbology/read-symbols-from-mobile-style-file/src/main/java/com/esri/samples/read_symbols_from_mobile_style_file/ReadSymbolsFromMobileStyleFileController.java
JonLavi: That's a great observation, I wasn't aware ColorPicker exists. Added, and in the process there's a slight refactor

JonLavi: I wasn't aware ColorPicker existed! Added, and done some refactoring in the process.

tschie (approved):

PR #360, opened 2019-06-06
update gradle wrapper to 5.0

tschie:
- uses Gradle `wrapper` configuration
- updates wrapper version to 5.0
- updates wrapper files
- adds Gson dependency (Gradle 5 enforces api/implementation on transitive dependencies)

Gradle 4.x had some issues with Java 11 where it would have to recompile (run `compileJava`) each time you ran a sample. With Gradle 5, it stays up to date.


JonLavi (approved):
Everything seems to work just fine for me.

Rachael-E (approved):
looks good!

PR #361, opened 2019-06-07
refactor to use fxml

tschie:
Reactorings for https://github.com/Esri/arcgis-runtime-samples-java/pull/352. The diff looks like I rewrote everything, but mostly all I did was refactor to use fxml. I think it helped separate out some of the view logic and cleaned up the initialize method.


JonLavi:
Thanks @tschie. I see quite a few changes in there actually. You took out the enums and replaced with `private Feature selectedFeature`.
Looks like the enums and decision making when clicking were the original cause of the mess, so I'm happy with your suggestions.

PR #366, opened 2019-06-17
patch oAuth sample custom challengehandler

JonLavi:
Hi @Rachael-E & @tschie,
as discussed, there is a slight bug with some URLs and the custom challenge handler in the oAuth sample.
The fix involves adding `"&display=classic"` to the authorization URL. I've also added an `Additional information` section to the readme to explain why this is necessary.
Let me know what you think.
Thanks!


tschie:
@JonLavi We also use this custom OAuth handler in the Create and Save Map sample? Can you please make the change there too? In the next release, we'll just change both of them to use the default authentication handler.

tschie (commented):
@JonLavi We also use this custom OAuth handler in the Create and Save Map sample? Can you please make the change there too? In the next release, we'll just change both of them to use the default authentication handler.

JonLavi:
Thanks @tschie, I didn't expect other samples to use this pattern too.
Added a fix for Create and Save Map as well, and added a reference to the `OAuthChallengeHandler` in the new readme paragraph too so it's more specific as to where the problem is.

tschie (approved):
Looks good. Thanks

PR #373, opened 2019-06-27
autosplit

JonLavi:
Hey @tschie, as discussed a PR on your split branch. an overview on what happened:
* 👍 in each category, we should have a folder structure like so: `category_name\sample-name\src\main\java\com\esri\samples\sample_name\SampleNameSample.java`

* 👍 put all .java files in `sample-name\src\main\java\com\esri\samples\sample_name\`

* 👍 put all other files in the sample root foler `sample-name\`

* 👍  placed a copy of `build.gradle`/`gradlew`/`gradlew.bat`/`settings.gradle` in each samples' root folder

* 👎 I had a look, and it seems we have to modify `build.gradle` manually to match the sample

* 👍 edited `settings.gradle` to match the pattern `rootProject.name = 'sample-name'`

* 👍 the other two files (`gradlew` and `gradlew.bat`) should be identical for each sample project, I think?

* 👍 put a copy of `SampleNameLauncher.java` in the folder `sample-name\src\main\java\com\esri\samples\sample_name\`. It should have the correct contents, with one caviat:

* 👎 didn't remove the category name from the package name within each `.java` file. It should be gone now since every sample is it's own project. This means the launcher file can't find the class name, so we need to change those. I'm not good enough with regex to delete everything in the package name at a certain spot between the `.`.

* 👎 can't automate moving the sample resources / fxml / css files too easily. I think this might have to be a manual job

If this step seems reasonable, feel free to go ahead and merge.


tschie (approved):
Excellent work! Good to merge!

PR #376, opened 2019-07-02
reformat all readmes from html to md

JonLavi:
Hi @tschie,  I had a look through each file, they should be 99% ok, but I might have missed a thing or two. Especially unusual character combinations may have tripped up the search and replace. Let me know what you think. Cheers!


tschie (approved):
Yep, definitely a few minor character issues still in there but this is a great place to start from. Good to merge.

PR #387, opened 2019-07-05
Tschie/split

tschie:
This PR restructures the samples into individual Gradle projects that can be run standalone.


tschie:
@mbcoder @Rachael-E @JonLavi This should be ready to go now. I've ran through every sample to at least make sure they start up and have their initial data. I've also made sure that they all work in the updated sample viewer. Please review.

README.md
```diff
@@ -1,80 +1,67 @@
...
+2. Run `./gradlew run` on Linux/Mac or `gradlew.bat run` on Windows to run the app.
+
+### Java 11
 Java 11 users may find exceptions when running the project if their library path is still set for Oracle JDK 1.8 (see the [OpenJavaFX docs](https://openjfx.io/openjfx-docs/) for more information). A workaround for this is to add the following argument in the `run` task of the Gradle buildscript:
 ```
 systemProperty "java.library.path", "C:\tmp"
 ```
 
-## Offline sample data
+### Offline sample data
```
JonLavi, line 39: Should this be a second-order heading `##`? Or did you intend to put it under `## Instructions`?

JonLavi (approved):
This is great! Just a small comment asking about a header in the main readme. 
I've done some spot testing in each category and all seems to work smoothly. 🥇

README.md
tschie: Yep, these sections should all be under Instructions. Esri repos have some guidelines for the sections the readme should contain. This information fits best under their Instructions section.

README.md
```diff
@@ -1,80 +1,67 @@
 # ArcGIS Runtime Java SDK Samples
-### 100.5.0
-This project includes a set of samples demonstrating how to accomplish various mapping and GIS tasks with the ArcGIS Runtime SDK for Java.
 
-Be sure to also check out our [toolkit on GitHub](https://github.com/Esri/arcgis-runtime-toolkit-java).
+Here is a collection of sample projects demonstrating how to accomplish various mapping and GIS tasks with the ArcGIS Runtime SDK for Java.
```
Rachael-E, line 3: Minor suggestion here - this intro seems a little casual to me. I prefer the more structured/professional approach of our initial intro, and would suggest the following intro: 

"This repository contains a collection of sample projects...."

README.md
```diff
@@ -1,80 +1,67 @@
...
 
-On Linux/Mac
-```
-$ ./gradlew run -PmainClass="com.esri.samples.scene.display_scene.DisplaySceneSample"
-```
+1. Open IntelliJ IDEA and select _File > Open..._.
+2. Choose one of the sample project directories (not the category folder) and click _OK_.
+3. Select _File > Project Structure..._ and ensure that the Project SDK and language level are set to use Java 11.
+4. Open the Gradle view with _View > Tool Windows > Gradle_.
+5. In the Gradle view, double-click the `run` task under _Tasks > application_ to run the app.
```
Rachael-E, line 17: We need some additional doc here to describe the option of running the samples using the launcher. Using the launcher doesn't require JavaFX to be installed, whereas running the class directly does (and the module path set to the JavaFX install directory): do we want to alert our users to this alternative workflow?

Rachael-E (changes requested):
@tschie my suggested changes are around the README.md content to clarify the option of using the Launcher class to run a sample. However, there are some deeper issues with this PR that we need to get to the bottom of before merging.

1. There seems to be discrepancies using the Launcher class and the Gradle Run task to run the sample. I've found that samples which use a `ModelSceneSymbol` will run fine with Gradle but with the Launcher, the model fails to render and the sample is broken (with no errors). Samples include: Animate 3D Graphic, Choose Camera Controller and Line of Sight Location.

2. When running using the Launcher class, multiple samples can be opened. However, when running the sample with Gradle, another sample often can't run due to the execution failing for task `copyNatives`: the jniLibs runtimecore.dll is being used by another process. This isn't idea since users might want to run multiple samples concurrently.

3. Using the Gradle run task to launch a sample takes a lot longer than using the Launcher class (I don't know if there is a solution to this).  

4. General observation, nothing to be done probably: this set up is heavier on IDE memory than previous set up: I guess this is to be expected.

tschie:
@Rachael-E 
1) The Launcher class is not a supported way of running the samples, hence why I omitted it from the readme. Gradle projects should always be run through Gradle. The Launcher class's main use is to run the app as a jar without having to add arguments to specify the location of the JavaFX modules. The fact that you can right-click the class to run the app is a convenience provided by the IDE. But it is only useful when the app is simple. It cannot know about other dependencies that are needed by the app (like local sample data and native libraries) which are defined in the buildscript. This is probably the cause of your first issue.

2) Each Gradle project needs to fetch its dependencies, and then cache the hash of that dependency so it knows if everything is up-to-date. This means that the first time you run a specific sample, it will redownload the native libraries which can cause an issue if there is another sample running and using those files. This only happens once though. After samples are up-to-date, you can run any number of samples at the same time. Finally, being able to open the root project in your IDE and view/run any of the samples from there is a convenience for us. Users should still open each sample individually to run them.

3) That's probably because using the Launcher doesn't download extra dependencies like data and the native libraries (it may still work because you already had them in .arcgis). Once Gradle is up-to-date, it should run just as fast. You would need to run these tasks in Gradle before you used the Launcher right-click method.

4) That's expected with lots of Gradle projects. Again, users should open each sample they want to work with individually.

README.md
tschie: I noticed that we were not following the Esri guidelines for our repo structure, which is why I changed it. But your suggestion sounds better.

Rachael-E:
@tschie Ah ok, sure, that all makes perfect sense and I understand now. Thanks for taking the time to explain, appreciated! :)

Rachael-E (approved):

PR #388, opened 2019-07-09
Patch/create and save map - default challenge handler

JonLavi:
Hi all,
since the old 'Create and Save Map' sample also used the custom challenge handler, I've changed it to use the Default Authentication Challenge Handler. It's a very small change and shouldn't affect the way the sample behaves. This one still required the user to provide a portal URL, client Id and redirect URI for the portal where they would like to save their map to.
Tested this using a test app I've created on AGOL, and works as expected.


src/main/java/com/esri/samples/map/create_and_save_map/CreateAndSaveMapController.java
```diff
@@ -147,12 +150,11 @@ void authenticate() {
...
 
             // show alert message on error
-            showMessage("Authentication failed", portal.getLoadError().getMessage(), Alert.AlertType.ERROR);
+            new Alert(Alert.AlertType.ERROR, "Authentication failed: " + portal.getLoadError().getMessage()).show();
           }
         });
 
-        // loading the portal info of a secured resource
-        // this will invoke the authentication challenge
+        // load the portal info of a secured resource. This will invoke the authentication challenge
```
Rachael-E: just spotted a typo from the original code, but GitHub won't allow me to add comments to existing code :D can you correct the spelling of "unsuccessful" on line 205 please?

Rachael-E (approved):
@JonLavi yup this looks good to go! I just spotted a pre-existing typo which could be addressed as part of this PR. 

I also think for this sample we should think about disabling the "Save" button once the map has been created and saved. I noticed that the same Title can be saved over and over, allowing creation of multiple  new web maps with the same name (which, interestingly, when making a webmap directly within AGOL itself isn't possible). This is probably beyond the scope of this PR though.

src/main/java/com/esri/samples/map/create_and_save_map/README.md
```diff
@@ -20,19 +20,18 @@ chosen folder.</p>
   <li>Create an <code>ArcGISMap</code> with a <code>Basemap</code> and operational layers</li>
   <li>Create an <code>OAuthConfiguration</code> with your portal and app credential.</li>
   <li>Add the configuration to the <code>AuthenticationMangager</code></li>
-  <li>Create a <code>Portal</code> and load it. Use a custom <code>AuthenticationChallengeHandler</code> to 
-  authenticate with your username and password</li>
+  <li>Create a <code>Portal</code> and load it. Use a <code>DefaultAuthenticationChallengeHandler</code> to authenticate with your username and password</li>
```
Rachael-E: Whoops just spotted another original typo in the line above this: 
`AuthenticationMangager` -> `AuthenticationManager`

src/main/java/com/esri/samples/map/create_and_save_map/README.md
```diff
@@ -20,19 +20,18 @@ chosen folder.</p>
...
 </ol>
 
 <h2>Relevant API</h2>
 
 <ul>
-  <li>AuthenticationChallengeHandler</li>
   <li>ArcGISMap</li>
   <li>ArcGISMapImageLayer</li>
   <li>Basemap</li>
+  <li>DefaultAuthenticationChallengeHandler</li>
```
Rachael-E: I don't think we need the included Additional Information section anymore, now that we are using the `DefaultAuthenticationChallengeHandler`? The information about using &display=classic is now redundant.

JonLavi:
Thanks @Rachael-E. That's an interesting observation! I'll look into disabling the button - it would have to be reinabled if there is some sort of input / change to the name, and if we're going there, perhaps we need a check whether a map of the same name already exists on the portal?

Rachael-E (commented):
spotted an additional change needed for readme

src/main/java/com/esri/samples/map/create_and_save_map/README.md
JonLavi: You are 100% right, I could have sworn I deleted this.
Also added `AuthenticationManager` to the Readme as it's quite an important bit of the API here

tschie (approved):

map/create-and-save-map/src/main/java/com/esri/samples/create_and_save_map/CreateAndSaveMapController.java
```diff
@@ -147,12 +150,11 @@ void authenticate() {
           } else if (portal.getLoadStatus() == LoadStatus.FAILED_TO_LOAD) {
 
             // show alert message on error
-            showMessage("Authentication failed", portal.getLoadError().getMessage(), Alert.AlertType.ERROR);
+            new Alert(Alert.AlertType.ERROR, "Authentication failed: " + portal.getLoadError().getMessage()).show();
```
tschie, line 162: You can get rid of `showMessage` if it's not needed.

map/create-and-save-map/src/main/java/com/esri/samples/create_and_save_map/CreateAndSaveMapController.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017 Esri.
```
tschie, line 2: This is fine, but in the future, we shouldn't change these.

map/create-and-save-map/README.md
```diff
@@ -16,25 +16,20 @@ To create and save a map to your portal for use in an app:
...
 
 ## Relevant API
 
-* AuthenticationChallengeHandler
 * ArcGISMap
 * ArcGISMapImageLayer
+* AuthenticationManager
 * Basemap
+* DefaultAuthenticationChallengeHandler
 * MapView
```
tschie: Remove `ArcGISMap`, `ArcGISMapImageLayer`, `Basemap`, and `MapView`  as they are not directly related to the purpose of the sample.

map/create-and-save-map/src/main/java/com/esri/samples/create_and_save_map/CreateAndSaveMapController.java
```diff
@@ -213,9 +215,10 @@ private void saveMap() {
 
   /**
    * Display an alert to the user with the specified information.
-   * @param title alert title
+   *
+   * @param title       alert title
```
tschie: I don't think this much space is needed between the parameter name and description

tschie (approved):
Just a couple minor things. Good to go after that.

map/create-and-save-map/src/main/java/com/esri/samples/create_and_save_map/CreateAndSaveMapController.java
JonLavi: 👍  I've removed the rest of the method calls and replaced with the same style 'new Alert().show'

map/create-and-save-map/src/main/java/com/esri/samples/create_and_save_map/CreateAndSaveMapController.java
JonLavi: Good to know. I thought the change to the default challenge handler would warrant seeing this as a revamped sample. I'll leave the year be next time.

map/create-and-save-map/README.md
JonLavi: Good spot on this, I've also updated the readme to have the additional sections that are considered mandatory (use case and tags)

map/create-and-save-map/src/main/java/com/esri/samples/create_and_save_map/CreateAndSaveMapController.java
JonLavi: Thanks, I think that was Intellij automatically formatting things. That method is removed now, so no longer applicable.

JonLavi:
Thanks @tschie, all suggestions implemented!

PR #389, opened 2019-07-12
corrects WKID to 4326

Rachael-E:
@JonLavi, a patch for a small change: I've updated the WKID in this sample from 4236 to 4326 (spotted during the iOS implementation of this sample), please can you review.

### DIFF
--- a/src/main/java/com/esri/samples/geometry/project/ProjectSample.java
+++ b/src/main/java/com/esri/samples/geometry/project/ProjectSample.java
@@ -89,7 +89,7 @@ public class ProjectSample extends Application {
           Point originalPoint = mapView.screenToLocation(point2D);
           inputPointGraphic.setGeometry(originalPoint);
           // project the web mercator point to WGS84 (WKID 4326)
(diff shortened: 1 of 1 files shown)


src/main/java/com/esri/samples/geometry/project/ProjectSample.java
```diff
@@ -89,7 +89,7 @@ public void start(Stage stage) {
           Point originalPoint = mapView.screenToLocation(point2D);
           inputPointGraphic.setGeometry(originalPoint);
           // project the web mercator point to WGS84 (WKID 4326)
-          Point projectedPoint = (Point) GeometryEngine.project(originalPoint, SpatialReference.create(4236));
+          Point projectedPoint = (Point) GeometryEngine.project(originalPoint, SpatialReference.create(4326));
```
colinanderson: You can use `SpatialReferences.getWgs84()`

JonLavi (approved):
Good spot, works fine for me.
Tested @colinanderson's suggestion with `SpatialReferences.getWgs84()`, which also works fine.
Not sure whether there's any preference for using the WKID approach in this sample, but I think promoting the `.getWgs84()` method is a nice way to do it?

PR #392, opened 2019-07-24
Apply scheduled updates to preplanned map area

JonLavi:
Hi both,
This is the new sample for applying updates to a preplanned map area.
Hopefully the gradle files are all correct this time.
Looking forward to seeing your improvement suggestions.

Let me know whether to keep the GIF or not. I added it since some of the samples in the sample viewers have one. However it's not the most interactive sample, so happy to remove it again.

Thanks!


map/apply-scheduled-updates-to-preplanned-map-area/build.gradle
```diff
@@ -0,0 +1,75 @@
+plugins {
+    id 'application'
+    id 'org.openjfx.javafxplugin' version '0.0.5'
+}
+
+group = 'com.esri.samples'
+
+ext {
```
Rachael-E, line 8: this is a process I'm still not clear on: when we merge new samples to dev, should we ensure the arcgisVersion is set to "100.6.0" or will we do a bulk change before dev is merged into master?

map/apply-scheduled-updates-to-preplanned-map-area/src/main/java/com/esri/samples/apply_scheduled_updates_to_preplanned_map_area/ApplyScheduledUpdatesToPreplannedMapAreaSample.java
```diff
@@ -0,0 +1,234 @@
+package com.esri.samples.apply_scheduled_updates_to_preplanned_map_area;
+
```
Rachael-E, line 18: copyright statement to be added

map/apply-scheduled-updates-to-preplanned-map-area/src/main/java/com/esri/samples/apply_scheduled_updates_to_preplanned_map_area/ApplyScheduledUpdatesToPreplannedMapAreaSample.java
```diff
@@ -0,0 +1,234 @@
...
+  @Override
+  public void start(Stage stage) {
+    try {
+      // create stack pane and application scene
+      StackPane stackPane = new StackPane();
+      Scene scene = new Scene(stackPane);
+
+      // set title, size, and add scene to stage
+      stage.setTitle("Update Offline Map Sample");
```
Rachael-E: this title needs to be updated

map/apply-scheduled-updates-to-preplanned-map-area/src/main/java/com/esri/samples/apply_scheduled_updates_to_preplanned_map_area/ApplyScheduledUpdatesToPreplannedMapAreaSample.java
```diff
@@ -0,0 +1,234 @@
...
+
+                  // create default parameters for the sync task
+                  ListenableFuture<OfflineMapSyncParameters> offlineMapSyncParametersFuture = offlineMapSyncTask.createDefaultOfflineMapSyncParametersAsync();
+                  offlineMapSyncParametersFuture.addDoneListener(() -> {
+                    try {
+                      OfflineMapSyncParameters offlineMapSyncParameters = offlineMapSyncParametersFuture.get();
+
+                      // set the sync direction to none, since we only want to update
+                      offlineMapSyncParameters.setSyncDirection(SyncGeodatabaseParameters.SyncDirection.NONE);
+                      // set the parameters to download all updated for the mobile map packages
```
Rachael-E: updated -> updates

map/apply-scheduled-updates-to-preplanned-map-area/src/main/java/com/esri/samples/apply_scheduled_updates_to_preplanned_map_area/ApplyScheduledUpdatesToPreplannedMapAreaSample.java
```diff
@@ -0,0 +1,234 @@
...
+                      // create a sync job using the parameters
+                      OfflineMapSyncJob offlineMapSyncJob = offlineMapSyncTask.syncOfflineMap(offlineMapSyncParameters);
+
+                      // start the job and get the results
+                      offlineMapSyncJob.start();
+                      offlineMapSyncJob.addJobDoneListener(() -> {
+                        if (offlineMapSyncJob.getStatus() == Job.Status.SUCCEEDED) {
+                          OfflineMapSyncResult offlineMapSyncResult = offlineMapSyncJob.getResult();
+
+                          // reopen the mobile map package if required
```
Rachael-E: "if mobile map package reopen is required, close the existing mobile map package and load it again"

map/apply-scheduled-updates-to-preplanned-map-area/src/main/java/com/esri/samples/apply_scheduled_updates_to_preplanned_map_area/ApplyScheduledUpdatesToPreplannedMapAreaSample.java
```diff
@@ -0,0 +1,234 @@
...
+      tempMobileMapPackageDirectory.toFile().deleteOnExit();
+      Path sourceDirectory = Paths.get("./samples-data/canyonlands/");
+      FileUtils.copyDirectory(sourceDirectory.toFile(), tempMobileMapPackageDirectory.toFile());
+
+      // load the offline map as a mobile map package
+      MobileMapPackage mobileMapPackage = new MobileMapPackage(tempMobileMapPackageDirectory.toString());
+      mobileMapPackage.loadAsync();
+      mobileMapPackage.addDoneLoadingListener(() -> {
+        if (mobileMapPackage.getLoadStatus() == LoadStatus.LOADED && !mobileMapPackage.getMaps().isEmpty()) {
+
```
Rachael-E, line 82: you could add:
`ArcGISMap offlineMap = mobileMapPackage.getMaps().get(0));`

and use that on line 97 and 103 to avoid repetition

map/apply-scheduled-updates-to-preplanned-map-area/src/main/java/com/esri/samples/apply_scheduled_updates_to_preplanned_map_area/ApplyScheduledUpdatesToPreplannedMapAreaSample.java
```diff
@@ -0,0 +1,234 @@
...
+          //add the map from the mobile map package to the map view
```
Rachael-E: extra space between "//" and "add"

map/apply-scheduled-updates-to-preplanned-map-area/build.gradle
```diff
@@ -0,0 +1,75 @@
+plugins {
```
Rachael-E, line 1: like we discussed today, this build.gradle is missing the necessary components to download the data, thanks for working to bring those in!

Rachael-E (changes requested):
hey @JonLavi this is looking good! Once you're added in the Gradle changes I'll take another look, otherwise just a few minor things to tidy up in the meanwhile.  I don't think we need a GIF here, we tend to include a GIF if there's something interesting to show e.g. the change atmosphere effect cycling through different modes.

map/apply-scheduled-updates-to-preplanned-map-area/src/main/java/com/esri/samples/apply_scheduled_updates_to_preplanned_map_area/ApplyScheduledUpdatesToPreplannedMapAreaSample.java
```diff
@@ -0,0 +1,252 @@
...
+
+              } else {
+                updateAvailableLabel.setText("Updates: NOT AVAILABLE");
+              }
+            } catch (Exception e) {
+              new Alert(Alert.AlertType.ERROR, "Error checking for Scheduled Updates Availability: " + e.getMessage()).show();
+            }
+          });
+        } else {
+          new Alert(Alert.AlertType.ERROR, "Failed to load the mobile map package.").show();
```
TADraeseke, line 94: Could get the message from the mobile map package here like you do for your other errors.
`mobileMapPackage.getLoadError().getMessage()`

Usefully, it will report a file not found in case users haven't provisioned correctly or the data is corrupted etc.

JonLavi:
Hi all, I've added an additional check for updates to occur after the available updates are applied. This should return 'no updates available' and set the lables based on that.
**At the moment there's a core bug in this API feature, but the fix will be in the Friday build (2505?).**
It's also not in the original design as of yet, so will update.
The code is getting a little bit nested, but since we're always waiting on an async task to finish, I can't see an elegant way out of it just now. If anyone has refactoring suggestions, please do share!
Thanks

map/apply-scheduled-updates-to-preplanned-map-area/build.gradle
```diff
@@ -0,0 +1,93 @@
...
+}
+
+dependencies {
+    compile "com.esri.arcgisruntime:arcgis-java:$arcgisVersion"
+    natives "com.esri.arcgisruntime:arcgis-java-jnilibs:$arcgisVersion"
+    natives "com.esri.arcgisruntime:arcgis-java-resources:$arcgisVersion"
+    compile 'commons-io:commons-io:2.4'
+}
+
+task downloadData(type: Download) {
```
Rachael-E: check with @tschie but you might want to move these into settings.gradle now to be consistent with the latest changes there?

map/apply-scheduled-updates-to-preplanned-map-area/src/main/java/com/esri/samples/apply_scheduled_updates_to_preplanned_map_area/ApplyScheduledUpdatesToPreplannedMapAreaSample.java
```diff
@@ -0,0 +1,274 @@
...
+
+  @Override
+  public void start(Stage stage) {
+    try {
+      // create stack pane and application scene
+      StackPane stackPane = new StackPane();
+      Scene scene = new Scene(stackPane);
+
+      // set title, size, and add scene to stage
+      stage.setTitle("Apply Schedueld Updates to Preplanned Map Area");
```
Rachael-E: typo in title "scheduled"

map/apply-scheduled-updates-to-preplanned-map-area/src/main/java/com/esri/samples/apply_scheduled_updates_to_preplanned_map_area/ApplyScheduledUpdatesToPreplannedMapAreaSample.java
```diff
@@ -0,0 +1,274 @@
...
+      applyUpdatesButton.setDisable(true);
+
+      // create labels to show update availability and size
+      Label updateAvailableLabel = new Label("Updates: ");
+      updateAvailableLabel.setTextFill(Color.WHITE);
+      Label updateSizeLabel = new Label("Update size: ");
+      updateSizeLabel.setTextFill(Color.WHITE);
+
+      // create a control panel for the UI elements
+      VBox controlsVBox = new VBox(6);
```
Rachael-E: on a Mac, the control panel is too narrow and the labels and button are cut off (this has happened to me before with other samples, must be a different resolution involved). Shouldn't have to be too much wider, enough to fit in 4 extra characters.

map/apply-scheduled-updates-to-preplanned-map-area/src/main/java/com/esri/samples/apply_scheduled_updates_to_preplanned_map_area/ApplyScheduledUpdatesToPreplannedMapAreaSample.java
```diff
@@ -0,0 +1,274 @@
...
+                                // add the map from the mobile map package to the map view
+                                mapView.setMap(mobileMapPackage.getMaps().get(0));
+
+                              } else {
+                                new Alert(Alert.AlertType.ERROR, "Failed to load the mobile map package.").show();
+                              }
+                            });
+                          }
+
+                          // perform an additional check for updates
```
Rachael-E: why are we performing an additional check for updates here? could do with an expanded comment.

map/apply-scheduled-updates-to-preplanned-map-area/src/main/java/com/esri/samples/apply_scheduled_updates_to_preplanned_map_area/ApplyScheduledUpdatesToPreplannedMapAreaSample.java
```diff
@@ -0,0 +1,274 @@
...
+                          ListenableFuture<OfflineMapUpdatesInfo> offlineMapUpdatesInfoFuture2 = offlineMapSyncTask.checkForUpdatesAsync();
```
Rachael-E: much of this code is a repeat of the above: can abstract most of this code out into a separate method. This avoids the use of a variable name with a number at the end, we should try to avoid that for readability.

map/apply-scheduled-updates-to-preplanned-map-area/src/main/java/com/esri/samples/apply_scheduled_updates_to_preplanned_map_area/ApplyScheduledUpdatesToPreplannedMapAreaSample.java
```diff
@@ -0,0 +1,274 @@
...
+                            try {
+                              // get and check the results
+                              OfflineMapUpdatesInfo offlineMapUpdatesInfo2 = offlineMapUpdatesInfoFuture2.get();
+
+                              // update UI for available updates
+                              if (offlineMapUpdatesInfo2.getDownloadAvailability() == OfflineUpdateAvailability.AVAILABLE) {
+                                updateAvailableLabel.setText("Updates: AVAILABLE");
+
+                                // check and show update size
+                                long updateSize2 = offlineMapUpdatesInfo2.getScheduledUpdatesDownloadSize();
```
Rachael-E: should avoid using duplicate variable names with numbers after: though in this instance it's probably a handy indicator this code is being reused elsewhere and can be abstracted out into a method!

map/apply-scheduled-updates-to-preplanned-map-area/src/main/java/com/esri/samples/apply_scheduled_updates_to_preplanned_map_area/ApplyScheduledUpdatesToPreplannedMapAreaSample.java
```diff
@@ -0,0 +1,274 @@
...
+import javafx.scene.layout.VBox;
+import javafx.scene.paint.Color;
+import javafx.scene.paint.Paint;
+import javafx.stage.Stage;
+
+import com.esri.arcgisruntime.concurrent.Job;
+import com.esri.arcgisruntime.concurrent.ListenableFuture;
+import com.esri.arcgisruntime.loadable.LoadStatus;
+import com.esri.arcgisruntime.mapping.ArcGISMap;
+import com.esri.arcgisruntime.mapping.Expiration;
```
Rachael-E: unused import

Rachael-E (changes requested):
thanks @JonLavi, there's some refactoring that can be done to make this a bit neater by pulling some of the repeated code out into a method. Some other things to look at too.

map/apply-scheduled-updates-to-preplanned-map-area/src/main/java/com/esri/samples/apply_scheduled_updates_to_preplanned_map_area/ApplyScheduledUpdatesToPreplannedMapAreaSample.java
JonLavi: Expanded the comment to say 
`// perform another check for updates, to make sure that the newest update was applied`

map/apply-scheduled-updates-to-preplanned-map-area/src/main/java/com/esri/samples/apply_scheduled_updates_to_preplanned_map_area/ApplyScheduledUpdatesToPreplannedMapAreaSample.java
JonLavi: Refactored as discussed!

map/apply-scheduled-updates-to-preplanned-map-area/build.gradle
```diff
@@ -1,15 +1,12 @@
...
-    id "de.undercouch.download" version "3.4.3"
 }
 
-import de.undercouch.gradle.tasks.download.Download
-
 group = 'com.esri.samples'
 
 ext {
-    arcgisVersion = '100.6.0-2484'
+    arcgisVersion = '100.6.0-2505'
```
Rachael-E: can make this 100.6.0 now that is available

Rachael-E (approved):
looks good to me, only change to suggest is we can make the version number 100.6.0 now

map/apply-scheduled-updates-to-preplanned-map-area/build.gradle
JonLavi: Gladly!

tschie (changes requested):
Can we change the UI so that it shows an `Alert` when an update is available? The user should be able to ignore it or apply it. The alert text can show the size of the update.

JonLavi:
Done!

map/apply-scheduled-updates-to-preplanned-map-area/src/main/java/com/esri/samples/apply_scheduled_updates_to_preplanned_map_area/ApplyScheduledUpdatesToPreplannedMapAreaSample.java
JonLavi: We generally don't do this in java since the content of the alert is not considered useful for the end-user. I suppose devs can print these to the stack trace to debug their apps if they need?

map/apply-scheduled-updates-to-preplanned-map-area/src/main/java/com/esri/samples/apply_scheduled_updates_to_preplanned_map_area/ApplyScheduledUpdatesToPreplannedMapAreaSample.java
```diff
@@ -0,0 +1,245 @@
...
+      stage.setWidth(800);
+      stage.setHeight(700);
+      stage.setScene(scene);
+      stage.show();
+
+      // create a map view
+      mapView = new MapView();
+
+      // create a progress indicator
+      progressIndicator = new ProgressIndicator();
```
tschie: I think this should only be shown when the update is being applied, if at all.

JonLavi: Happy to remove it - I put it there to indicate the MMPK was still loading when starting the sample, but it does happen very fast anyway.

map/apply-scheduled-updates-to-preplanned-map-area/src/main/java/com/esri/samples/apply_scheduled_updates_to_preplanned_map_area/ApplyScheduledUpdatesToPreplannedMapAreaSample.java
```diff
@@ -0,0 +1,245 @@
...
+
+          // add the map from the mobile map package to the map view
+          ArcGISMap offlineMap = mobileMapPackage.getMaps().get(0);
+          mapView.setMap(offlineMap);
+
+          // create an offline map sync task with the preplanned area
+          offlineMapSyncTask = new OfflineMapSyncTask(offlineMap);
+
+          // check for available updates to the mobile map package
+          checkForScheduledUpdates();
```
tschie, line 91: I'm not sure this is necessary (don't know enough about scheduled updates). If this is really a concern, maybe the user should be able to always manually check for updates? I think in a simple sample like this, we should just trust this is the most recent update.

map/apply-scheduled-updates-to-preplanned-map-area/src/main/java/com/esri/samples/apply_scheduled_updates_to_preplanned_map_area/ApplyScheduledUpdatesToPreplannedMapAreaSample.java
```diff
@@ -0,0 +1,245 @@
...
+        // hide the progress indicator
+        progressIndicator.setVisible(false);
+
+      } catch (Exception ex) {
+        new Alert(Alert.AlertType.ERROR, "Error checking for Scheduled Updates Availability.").show();
+      }
+    });
+  }
+
+  private void applyScheduledUpdates() {
```
tschie, line 152: needs javadoc

tschie (changes requested):
Alert looks good. Some more ideas in comments.

map/apply-scheduled-updates-to-preplanned-map-area/src/main/java/com/esri/samples/apply_scheduled_updates_to_preplanned_map_area/ApplyScheduledUpdatesToPreplannedMapAreaSample.java
tschie: Remove it 👍

map/apply-scheduled-updates-to-preplanned-map-area/src/main/java/com/esri/samples/apply_scheduled_updates_to_preplanned_map_area/ApplyScheduledUpdatesToPreplannedMapAreaSample.java
JonLavi: The idea was to show the ability to check for updates in the first place `checkForUpdatesAsync`.
Then I've written the method `checkForScheduledUpdates()` to do this when the sample starts up, and then again when the update is applied, to show how a different result comes back (`OfflineUpdateAvailability.NONE`).
I can remove the second check for updates though, and have the sample apply the update as soon as it is launched, if that is what you mean?

tschie: Yes

JonLavi:
@tschie as discussed internally: reverted this to do the check for updates on startup and show update info, and to do a second check for updates to show it applied successfuly. 
I've no preference whether to use Alerts or a UI, so just let me know which one you think we should go for.

tschie (approved):
Good to merge

PR #393, opened 2019-07-25
Patch/dictionary symbol style update constructor

JonLavi:
Hi both,
Updated the DictionarySymbolStyle constructor to the upcoming one (using only a path to a stylex file).

Four samples are affected:

- [ ] Feature Layer Dictionary Renderer
- [ ] Graphics Overlay Dicrionary Renderer 3D
- [ ] Symbols Dictionary
- [ ] Dictionary Renderer with Graphics Overlay

**note:** XML data was updated for the first two

For the last one, while I was at it, I've updated the README to match the revised version of the common design README.
Let me know if there's any other improvements you spot.
Thanks!


display_information/dictionary-renderer-graphics-overlay/README.md
```diff
@@ -1,41 +1,34 @@
-# Dictionary Renderer with Graphics Overlay
+# Dictionary renderer with graphics overlay
 
-Display mil2525d symbols.
+Create graphics using a local mil2525d style file and an XML file with key/value pairs for each graphic.
```
Rachael-E: this seems quite wordy and too specific for the intro. This probably came from the shared wiki. I'd prefer to see it more succinct, i.e. as it was before "Display mil2525d symbols", the simplest phrasing possible.

display_information/dictionary-renderer-graphics-overlay/README.md
```diff
@@ -1,41 +1,34 @@
...
+![Dictionary Renderer Graphics Overlay Sample](dictionary-renderer-graphics-overlay.png)
 
-![](DictionaryRendererGraphicsOverlay.png)
+## Use case
+
+Use a dictionary renderer on a graphics overlay to display more transient data, such as military messages coming through a local tactical network.
+
+## How to use the sample
+
+When the sample is started, a dictionary renderer is applied and contents of a symbol style file are shown.
```
Rachael-E: This should be more of a how to use the sample rather than what it's doing: 

"Pan and zoom to see military symbols within the map"

display_information/dictionary-renderer-graphics-overlay/README.md
```diff
@@ -1,41 +1,34 @@
...
-    * \_wkid key, holds the geometry's spatial reference
-    * *control* points, creates the shape of the geometry
-    * other attributes explain to dictionary symbol how to display graphic
-    * add graphic to `GraphicsOverlay.getGraphics().add(Graphic)`
+1. Create a new `DictionarySymbolStyle` from a stylx file.
+2. Create a new `DictionaryRenderer` from the dictionary symbol style.
+3. Create a new `GraphicsOverlay` and set the  dictionary renderer to the graphics overlay.
+4. Parse through the local XML file creating a map of key/value pairs for each block of attributes.
+5. Create a `Graphic` for each attribute.
+6. Use the _`wkid` key to get the geometry's spatial reference.
```
Rachael-E: points 6 and 7 should be subsections under point 5.
Also _`wkid` -> `_wkid`

display_information/dictionary-renderer-graphics-overlay/README.md
```diff
@@ -1,41 +1,34 @@
...
-* MapView
-* Point
-* PointCollection
-* Polygon
-* Polyline
-* SymbolDictionary
+
+## Tags
+
+Visualization
```
Rachael-E: We should have more than one tag preferably for more involved samples. You can always fall back on the suggest tag keywords in the common samples repo if stuck for ideas. The following are there for this sample: 

> dictionary renderer, dictionary symbol style, military, symbol

Rachael-E (changes requested):
@JonLavi thanks for this. It's important that when doing patch work like this we don't forget to update the README.metadata.json file: it needs to be updated to match the README so that things like keywords, description, relevant apis and titles are consistent. Other than there are a few changes to make to the README.

JonLavi:
Thanks! It was only yesterday that I learnt that the metadata files don't get generated automatically. They will be updated now.

display_information/dictionary-renderer-graphics-overlay/README.md
JonLavi: I agree that more than one tag should be added. Lets raise the question at the upcoming readme discussion whether tags should repreat the API though, because that feels a little clunky to me.

JonLavi:
Hi @tschie, @Rachael-E and I had a discussion and agreed it could be useful to not package the xml file in `resources/` but get it from AGOL instead. Some other samples use this pattern too, where the xml is treated like data and is downloaded into `samples-data/`. I've gone ahead and made that change. Had to change the parsing of the XML around to make it work. I suspect there might be a more concise way of doing it. If you have any suggestions, please let us know. Cheers!

display_information/dictionary-renderer-graphics-overlay/build.gradle
```diff
@@ -37,6 +40,27 @@ dependencies {
     compile 'org.jooq:joox:1.4.0'
 }
 
+task downloadData() {
+    download {
+        src "https://arcgisruntime.maps.arcgis.com/sharing/rest/content/items/c78b149a1d52414682c86a5feeb13d30/data"
+        dest './samples-data/stylx/mil2525d.stylx'
+        overwrite false
+    }
+    download {
```
tschie: why is there a version of the messages xml in samples-data and in resources?

tschie (changes requested):
Sounds fine to move it into samples-data. You'll want to delete the ones in /resources though. This PR still has those.

JonLavi:
I simply forgot to remove them after testing that the new approach works. All removed now.

tschie (approved):

PR #396, opened 2019-08-02
Set annotation sublayer visibility

JonLavi:
Hi both,
Could you please review this sample when you get a chance?
It's quite simple, so shouldn't be a big task.
Thanks.


display_information/control-annotation-sublayer-visibility/README.md
```diff
@@ -0,0 +1,41 @@
+# Control annotation sublayer visibility
+
+Use annotation sublayers to gain finer control of annotation layer subtypes.
+
+![Control annotation sublayer visibility sample](ControlAnnotationSublayerVisibility.png)
+
+## Use case
+
+Annotation, which differs from labels by having a fixed place and size, is typically only relevant at particular scales. Annotation sublayers allow for finer control of annotation by allowing properties (like visibility in the map and legend) to be set and others to be read (like name) on subtypes of an annotation layer.
```
Rachael-E: I would prefer if we could use "e.g." instead of "like" here, but that may come down to personal preference!

display_information/control-annotation-sublayer-visibility/README.md
```diff
@@ -0,0 +1,41 @@
...
+* The "Open" annotation sublayer has its maximum scale set to 1:500 and its minimum scale set to 1:2000.
+* The "Closed" annotation sublayer has no minimum or maximum scales set, so will be drawn at all scales.
+
+Use the checkboxes to manually set "Open" and "Closed" annotation sublayers visibility to on or off.
+
+## How it works
+
+1. Load the `MobileMapPackage`.
+2. Populate checkbox text with the `AnnotationSublayer` names (including maximum and minimum scales, where relevant).
+3. Wire up the checkboxes to toggle the annotation sublayer's visibility.
```
Rachael-E: I find this phrasing a bit loose and too tightly connected to the UI. Let's go for something like this instead: 

"Set the annotation sublayer's visibility as required."

display_information/control-annotation-sublayer-visibility/README.md
```diff
@@ -0,0 +1,41 @@
...
+4. Add a listener for changes in map view navigation and add code to:
```
Rachael-E: this isn't API specific: this is relating to UI code. It's my understanding that we shouldn't have this in the How it works section, in the same way we don't provide instructions to set up a map view in all our samples. It's not directly relevant to the API functionality we're showing here.

display_information/control-annotation-sublayer-visibility/build.gradle
```diff
@@ -0,0 +1,83 @@
+plugins {
+    id 'application'
+    id 'org.openjfx.javafxplugin' version '0.0.5'
+    id "de.undercouch.download" version "3.4.3"
+}
+
+import de.undercouch.gradle.tasks.download.Download
```
Rachael-E: move these download instructions to settings.gradle

display_information/control-annotation-sublayer-visibility/build.gradle
```diff
@@ -0,0 +1,83 @@
...
+    natives "com.esri.arcgisruntime:arcgis-java-jnilibs:$arcgisVersion"
+    natives "com.esri.arcgisruntime:arcgis-java-resources:$arcgisVersion"
+}
+
+task downloadData(type: Download) {
+    src "https://arcgisruntime.maps.arcgis.com/sharing/rest/content/items/b87307dcfb26411eb2e92e1627cb615b/data"
+    dest 'samples-data/mmpk/GasDeviceAnno.mmpk'
+    overwrite false
+}
+
```
Rachael-E: we are missing the unpackData task: when I run the sample the data isn't downloaded. See the OpenMobileMapPackage sample for how to plug that all together.

display_information/control-annotation-sublayer-visibility/src/main/java/com/esri/samples/control_annotation_sublayer_visibility/ControlAnnotationSublayerVisibilitySample.java
```diff
@@ -0,0 +1,184 @@
...
+      openSublayerCheckbox.setSelected(true);
+
+      // create a control panel and label for the checkboxes
+      VBox controlsVBox = new VBox(6);
+      controlsVBox.setBackground(new Background(new BackgroundFill(Paint.valueOf("rgba(0,0,0,0.3)"), CornerRadii.EMPTY,
+              Insets.EMPTY)));
+      controlsVBox.setPadding(new Insets(10.0));
+      controlsVBox.setMaxSize(180, 50);
+      controlsVBox.getStyleClass().add("panel-region");
+      Label controlsVBoxLabel = new Label("Annotation Sublayer Visibility");
```
Rachael-E: we don't need this label, it should be clear from the instructions what this UI is responsible for (the .png will need updating with this change).

display_information/control-annotation-sublayer-visibility/src/main/java/com/esri/samples/control_annotation_sublayer_visibility/ControlAnnotationSublayerVisibilitySample.java
```diff
@@ -0,0 +1,184 @@
...
+      // add checkboxes and label to the control panel
+      controlsVBox.getChildren().addAll(controlsVBoxLabel, closedSublayerCheckbox, openSublayerCheckbox);
+
+      // create a map view
+      mapView = new MapView();
+
+      // show current map scale in a label at the bottom of the screen
```
Rachael-E: on a mobile this works well, however for desktop the scale is lost, and it's a long way for the eye to scan between the current scale and the min/max scale range of the open layer. It could be made more visible by moving it into the same vBox as the checkboxes, preferably below the checkboxes.

display_information/control-annotation-sublayer-visibility/src/main/java/com/esri/samples/control_annotation_sublayer_visibility/ControlAnnotationSublayerVisibilitySample.java
```diff
@@ -0,0 +1,184 @@
...
+      Label currentMapScaleLabel = new Label();
+      currentMapScaleLabel.setTextFill(Color.WHITE);
```
Rachael-E, line 82: this isn't changing the colour of the label to white, it's showing as black

display_information/control-annotation-sublayer-visibility/src/main/java/com/esri/samples/control_annotation_sublayer_visibility/ControlAnnotationSublayerVisibilitySample.java
```diff
@@ -0,0 +1,184 @@
...
+      mapView.addMapScaleChangedListener(mapScaleChangedEvent -> currentMapScaleLabel.setText("Scale: 1:" + Math.round(mapView.getMapScale())));
+
+      // load the mobile map package
+      mobileMapPackage = new MobileMapPackage("./samples-data/mmpk/GasDeviceAnno.mmpk");
+      mobileMapPackage.loadAsync();
+      mobileMapPackage.addDoneLoadingListener(() -> {
+        if (mobileMapPackage.getLoadStatus() == LoadStatus.LOADED) {
```
Rachael-E: also do a check to make sure there are maps in the mmpk

display_information/control-annotation-sublayer-visibility/src/main/java/com/esri/samples/control_annotation_sublayer_visibility/ControlAnnotationSublayerVisibilitySample.java
```diff
@@ -0,0 +1,184 @@
...
+          // set the mobile map package's map to the map view
+          mapView.setMap(mobileMapPackage.getMaps().get(0));
+          // find the annotation layer withing the map
```
Rachael-E: typo "within"

display_information/control-annotation-sublayer-visibility/src/main/resources/style.css
```diff
@@ -0,0 +1,27 @@
+.panel-region .label {
```
Rachael-E: only include relevant css for this sample

display_information/control-annotation-sublayer-visibility/src/main/java/com/esri/samples/control_annotation_sublayer_visibility/ControlAnnotationSublayerVisibilitySample.java
```diff
@@ -0,0 +1,184 @@
...
+              // load the annotation layer
+              layer.loadAsync();
+              layer.addDoneLoadingListener(() -> {
+                if (layer.getLoadStatus() == LoadStatus.LOADED) {
+                  // get annotation sublayer name from sublayer content
+                  AnnotationSublayer closedSublayer = (AnnotationSublayer) layer.getSubLayerContents().get(0);
+                  AnnotationSublayer openSublayer = (AnnotationSublayer) layer.getSubLayerContents().get(1);
+
+                  // set the layer name for the checkboxes
+                  closedSublayerCheckbox.setText(buildLayerName(closedSublayer));
```
Rachael-E: Was there a reason this is abstracted out as a method? It hides away the .getName functionality and scale of the annotation layers, which we want to showcase. How about something simple like this instead?
```
 closedSublayerCheckbox.setText(closedSublayer.getName());
 openSublayerCheckbox.setText(openSublayer.getName() + 
        " (" + "1:" + Math.round(openSublayer.getMaxScale()) + 
        " - 1:" + Math.round(openSublayer.getMinScale()) + ")");
```

display_information/control-annotation-sublayer-visibility/src/main/java/com/esri/samples/control_annotation_sublayer_visibility/ControlAnnotationSublayerVisibilitySample.java
```diff
@@ -0,0 +1,184 @@
...
+      StackPane.setMargin(controlsVBox, new Insets(20, 0, 0, 20));
+
+    } catch (Exception e) {
+      // on any error, print the stack trace
+      e.printStackTrace();
+    }
+  }
+
+  /**
+   * Get name, and where relevant, append min and max scales of each annotation sublayer.
```
Rachael-E: I think we can get rid of this method, see above

Rachael-E (changes requested):
@JonLavi looks like an accidental name change has occurred within the ShowCallout sample: their package name has now been renamed com.esri.samples.control_annotation_sublayer_visibility. 

Otherwise regarding this sample, we're missing the unpackData task for downloading the data and I've suggested a few UI, README and code changes.

display_information/control-annotation-sublayer-visibility/build.gradle
JonLavi: That's odd! It shouldn't need to unpack anything as it's downlloading an MMPK file and not a zip. I'll move it into the settings.gradle and test again.

display_information/control-annotation-sublayer-visibility/src/main/resources/style.css
JonLavi: Apparently the CSS doesn't apply to the text set directly to checkboxes.
i.e. we have to do `closedSublayerCheckbox.setTextFill(Color.WHITE);` ?
I've added those lines and removed the CSS file.

display_information/control-annotation-sublayer-visibility/src/main/java/com/esri/samples/control_annotation_sublayer_visibility/ControlAnnotationSublayerVisibilitySample.java
JonLavi: LGTM. I was following the Android implementation, and couldn't resist a good method 😄 . Changed now

JonLavi:
Thanks @Rachael-E, I've implemented your suggestions. Please let me know if the download through settings.gradle works ok for you, since it is being a bit difficult.
The UI changes were all good. However apparently we would need to apply the CSS rules to each Checkbox individually, as well as to the label that's within the control box. Because that's just as many lines of code as applying the colour directly in Java for each object (`currentMapScaleLabel.setTextFill(Color.WHITE);`), i've opted to do that and get rid of the CSS file instead.
Let me know what you think!

display_information/control-annotation-sublayer-visibility/README.md
```diff
@@ -0,0 +1,36 @@
+# Control annotation sublayer visibility
+
+Use annotation sublayers to gain finer control of annotation layer subtypes.
```
tschie: Why not just "Change the visibility of annotation sublayers."?

display_information/control-annotation-sublayer-visibility/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+![Control annotation sublayer visibility sample](ControlAnnotationSublayerVisibility.png)
+
+## Use case
+
+Annotation, which differs from labels by having a fixed place and size, is typically only relevant at particular scales. Annotation sublayers allow for finer control of annotation by allowing properties (e.g. visibility in the map and legend) to be set and others to be read (e.g. name) on subtypes of an annotation layer.
```
tschie: This first detail should go in Additional Information

display_information/control-annotation-sublayer-visibility/README.md
```diff
@@ -0,0 +1,36 @@
...
+An annotation dataset which marks valves as "Opened" or "Closed", might be set to display the "Closed" valves over a broader range of scales than the "Opened" valves, if the "Closed" data is considered more relevant by the map's author. Regardless, the user can be given a manual option to set visibility of annotation sublayers on and off, if required.
+
+## How to use the sample
+
+Start the sample and take note of the visibility of the annotation. Zoom in and out to see the annotation turn on and off based on scale ranges set on the data. The scale ranges were set by the map's author using ArcGIS Pro:
+
+* The "Open" annotation sublayer has its maximum scale set to 1:500 and its minimum scale set to 1:2000.
+* The "Closed" annotation sublayer has no minimum or maximum scales set, so will be drawn at all scales.
+
+Use the checkboxes to manually set "Open" and "Closed" annotation sublayers visibility to on or off.
```
tschie: Use the checkboxes to toggle a sublayer's visibility.

display_information/control-annotation-sublayer-visibility/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+## Use case
+
+Annotation, which differs from labels by having a fixed place and size, is typically only relevant at particular scales. Annotation sublayers allow for finer control of annotation by allowing properties (e.g. visibility in the map and legend) to be set and others to be read (e.g. name) on subtypes of an annotation layer.
+
+An annotation dataset which marks valves as "Opened" or "Closed", might be set to display the "Closed" valves over a broader range of scales than the "Opened" valves, if the "Closed" data is considered more relevant by the map's author. Regardless, the user can be given a manual option to set visibility of annotation sublayers on and off, if required.
+
+## How to use the sample
+
+Start the sample and take note of the visibility of the annotation. Zoom in and out to see the annotation turn on and off based on scale ranges set on the data. The scale ranges were set by the map's author using ArcGIS Pro:
```
tschie: Zoom in to change the map's scale. The annotation sublayers were authored so that the "Open" sublayer will be visible for scales 1:500-1:2000 and the "Closed" sublayer is always visible.

display_information/control-annotation-sublayer-visibility/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+* The "Open" annotation sublayer has its maximum scale set to 1:500 and its minimum scale set to 1:2000.
+* The "Closed" annotation sublayer has no minimum or maximum scales set, so will be drawn at all scales.
+
+Use the checkboxes to manually set "Open" and "Closed" annotation sublayers visibility to on or off.
+
+## How it works
+
+1. Load the `MobileMapPackage`.
```
tschie: Load a `MobileMapPackage` containing annotation layers.

display_information/control-annotation-sublayer-visibility/README.md
```diff
@@ -0,0 +1,36 @@
...
+2. Populate checkbox text with the `AnnotationSublayer` names (including maximum and minimum scales, where relevant).
```
tschie: This is sample-specific. These should be based on workflow. You can skip this step. 

Need a step about getting annotation sublayers from an annotation layer.

tschie (commented):
Looks good. Mostly readme feedback

display_information/control-annotation-sublayer-visibility/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+Use annotation sublayers to gain finer control of annotation layer subtypes.
+
+![Control annotation sublayer visibility sample](ControlAnnotationSublayerVisibility.png)
+
+## Use case
+
+Annotation, which differs from labels by having a fixed place and size, is typically only relevant at particular scales. Annotation sublayers allow for finer control of annotation by allowing properties (e.g. visibility in the map and legend) to be set and others to be read (e.g. name) on subtypes of an annotation layer.
+
+An annotation dataset which marks valves as "Opened" or "Closed", might be set to display the "Closed" valves over a broader range of scales than the "Opened" valves, if the "Closed" data is considered more relevant by the map's author. Regardless, the user can be given a manual option to set visibility of annotation sublayers on and off, if required.
```
tschie: This use case makes it sound like this is more about min/max scale than sublayer visibility. While I understand the sample demonstrates both, we have named it "... sublayer visibility". Can we simplify to:

"An annotation layer for water utilities may contain sublayers for water mains, pipes, and valves. One might turn off the visibility of water main and pipe sublayers to make the valves easier to identify."

I also have no idea what kinds of annotation sublayers would actually exist in this scenario. So might be worth checking what I wrote makes sense.

JonLavi: That works perfectly, great suggestion!

PR #399, opened 2019-08-06
Patch/geometry engine simplify - add isSimple check

JonLavi:
Hi both,
As discussed when reviewing the readme, this is a quick change to the 'Geometry Engine Simplify' sample. I've added a `.isSimple(geometry)` check to happen before applying the 'simplify' operation.
@Rachael-E and I have discussed perhaps changing the Data to be slightly overlapping circles instead of the three boxes (we found them a little confusing). But I'm thinking to maybe log an issue for that and keep it for later.
Let me know what you think! Thanks

### DIFF
--- a/geometry/geometry-engine-simplify/README.md
+++ b/geometry/geometry-engine-simplify/README.md
@@ -1,29 +1,31 @@
-# Geometry Engine Simplify
+# Geometry engine simplify
 
-Simplify a polygon that has multiple parts.
+Simplify a polygon with a self-intersecting geometry.
 
-![](GeometryEngineSimplify.png)
+![Geometry Engine Simplify Sample](GeometryEngineSimplify.png)
+
+## Use case
+
+A user may draw a polygon which is self-intersecting or contains incorrect ring orientations. The geometry must be simplified before it can be saved to a geodatabase.
 
 ## How to use the sample
 
-Click on the simplify button to apply the simplify geometry operation between the intersecting polygons. Click reset to restart the sample.
+Click the 'Simplify' button to simplify the geometry. Click 'Reset' to reset to the original geometry.
 
 ## How it works
 
-To perform the simplify geometry operation on a `Polygon`:
-
-1. Create a `GraphicsOverlay` and add it to the `MapView`.
-2. Define the `PointCollection` of the `Geometry`.
-3. Add the polygons to the GraphicsOverlay.
-4. Determine the simplified geometry by using the `GeometryEngine.simplify(polygon.getGeometry())`.
+1. Check if the polygon geometry needs to be simplified using `GeometryEngine.isSimple(Geometry)`.
+2. Simplify the polygon's geometry using `GeometryEngine.simplify(Geometry)`.
 
 ## Relevant API
 
 * Geometry
-* Graphic
-* GraphicsOverlay
-* MapView
-* Point
-* PointCollection
-* SimpleLineSymbol
-* SimpleFillSymbol
+* GeometryEngine
+
+## Additional information
+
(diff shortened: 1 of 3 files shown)


geometry/geometry-engine-simplify/src/main/java/com/esri/samples/geometry_engine_simplify/GeometryEngineSimplifySample.java
```diff
@@ -90,16 +90,22 @@ public void start(Stage stage) {
       resetButton.setMaxWidth(Double.MAX_VALUE);
       resetButton.setDisable(true);
 
-      // perform the simplify geometry operation
+      // resolve a button press
```
Rachael-E: this comment is a bit vague, I liked it the way it was before (ultimately, the button is still responsible for performing the simplify geometry operation).

> // perform the simplify geometry operation

geometry/geometry-engine-simplify/src/main/java/com/esri/samples/geometry_engine_simplify/GeometryEngineSimplifySample.java
```diff
@@ -90,16 +90,22 @@ public void start(Stage stage) {
...
-        resultGeomOverlay.getGraphics().add(new Graphic(resultPolygon, redSymbol));
+        // check if the geometry needs to be simplified
+        if (!GeometryEngine.isSimple(polygon.getGeometry())) {
 
-        resetButton.setDisable(false);
-        simplifyButton.setDisable(true);
+          // perform the simplify geometry operation
+          Geometry resultPolygon = GeometryEngine.simplify(polygon.getGeometry());
+
+          // update result as a red (0xFFE91F1F) geometry
```
Rachael-E: we don't need the hex code in this comment, it's already in the code in the line below. Also, I notice this comment was from the original design, but I think it would be more accurate to say "update result as a red graphic".

geometry/geometry-engine-simplify/src/main/java/com/esri/samples/geometry_engine_simplify/GeometryEngineSimplifySample.java
```diff
@@ -90,16 +90,22 @@ public void start(Stage stage) {
...
+          SimpleFillSymbol redSymbol = new SimpleFillSymbol(SimpleFillSymbol.Style.SOLID, 0xFFE91F1F, line);
+          resultGeomOverlay.getGraphics().add(new Graphic(resultPolygon, redSymbol));
+
+          resetButton.setDisable(false);
+          simplifyButton.setDisable(true);
+        }
       });
 
       // clear result layer
```
Rachael-E, line 111: can we add in 
`resetButton.setDisable(true)` here, so that it's disabled when it can't do anything.

geometry/geometry-engine-simplify/README.md
```diff
@@ -1,29 +1,31 @@
-# Geometry Engine Simplify
+# Geometry engine simplify
```
Rachael-E, line 1: With the update of this README, we should also make sure the metadata files are up to date too (I don't see a change for those through in this PR?)

Rachael-E (commented):
Thanks @JonLavi, the main feedback from me is to update the metadata file, other than that small changes suggested!

geometry/geometry-engine-simplify/README.md
JonLavi: Thanks for pointing that out, I've still not gotten used to the metadata files!

tschie (approved):

PR #401, opened 2019-08-06
move local data download logic to settings.gradle

tschie:
This PR moves all the data download logic to settings.gradle. This means that data will be downloaded during Gradle's initialization phase (before any task is run). The user should see the data download as soon as the project is imported. You can redownload the data by deleting the samples-data directory and refreshing the project.

By doing it this way, the sample viewer can include all of the sample projects and copy their samples-data folder into its own.


JonLavi:
I think merging the recent changes to how the XML gets downloaded in `feature-layer-dictionary-renderer` is causing some conflicts now. This was added in the meanwhile to `build.gradle`:
```
    download {
        src "https://arcgisruntime.maps.arcgis.com/sharing/rest/content/items/5491f01d92554751902ecad85cc5d6ca/data"
        dest 'build/data/dictionary.zip'
        overwrite false
    }
```
Should be a quick fix to convert into settings.gradle?

feature_layers/feature-layer-dictionary-renderer/settings.gradle
```diff
@@ -1 +1,16 @@
...
+
+if (!file("./samples-data").exists()) {
+    ant.get(
+            src: "https://arcgisruntime.maps.arcgis.com/sharing/rest/content/items/5491f01d92554751902ecad85cc5d6ca/data",
+            dest: file("./data.zip")
+    )
+    ant.unzip(
+            src: file("./data.zip"),
+            dest: file("./samples-data/dictionary")
+    )
```
JonLavi, line 14: Missing the stylx download here:
```
    ant.get(
            src: "https://arcgisruntime.maps.arcgis.com/sharing/rest/content/items/c78b149a1d52414682c86a5feeb13d30/data",
            dest: file("./samples-data/stylx/mil2525d.stylx")
    )        
```

JonLavi (changes requested):

tschie:
@JonLavi Good catch. Should be all fixed now.

JonLavi (approved):

symbology/read-symbols-from-mobile-style-file/settings.gradle
```diff
@@ -1 +1,11 @@
 rootProject.name = 'read-symbols-from-mobile-style-file'
+
+// download sample data
+def ant = new groovy.util.AntBuilder()
+
+if (!file("./samples-data").exists()) {
+    ant.get(
```
Rachael-E, line 7: When I imported the project, I got an error for a missing directory for this emoji-mobile.stylx file. This was fixed by adding this line in: 
   
 `file("./samples-data/stylx").mkdirs()`

Rachael-E (commented):
@tschie thanks, this is a good approach and gets past the problem of slow initial loading time when first opening a sample! It's looking good, just a few things to highlight:

- When I first imported the project, I got an error (see comment below). 

I also spotted that three samples (which require sample-data) are not getting the data from settings.gradle but still from build.gradle. These are: 
  + DictionaryRendererGraphicsOverlay
  + GraphicsOverlayDictionaryRenderer
  + SymbolDictionary

map/generate-offline-map-with-local-basemap/settings.gradle
```diff
@@ -1 +1,12 @@
 rootProject.name = 'generate-offline-map-with-local-basemap'
+
+// download sample data
+def ant = new groovy.util.AntBuilder()
+
+if (!file("./samples-data").exists()) {
+    file("./samples-data/naperville").mkdirs()
+    ant.get(
+            src: "https://arcgisruntime.maps.arcgis.com/sharing/rest/content/items/628e8e3521cf45e9a28a12fe10c02c4d/data",
+            dest: file("./samples-data/naperville")
```
JonLavi: I think it needs a specification of the true file name here, otherwise it downloads a `data` file without extension.
`dest: file("./samples-data/naperville/naperville_imagery.tpk")`
I will try and find the rest of the samples that work this way and point them out. I think it's only an issue when the files are not archived.

kml/play-a-kml-tour/settings.gradle
```diff
@@ -1 +1,12 @@
 rootProject.name = 'play-a-kml-tour'
+
+// download sample data
+def ant = new groovy.util.AntBuilder()
+
+if (!file("./samples-data").exists()) {
+    file("./samples-data/kml").mkdirs()
+    ant.get(
+            src: "https://arcgisruntime.maps.arcgis.com/sharing/rest/content/items/f10b1d37fdd645c9bc9b189fb546307c/data",
+            dest: file("./samples-data/kml")
```
JonLavi: dest: file("./samples-data/kml/Esri_tour.kmz")

JonLavi (changes requested):
Just one more behaviour I noticed - if we're downloading a file that doesn't need extracting, we have to specify the full file name with extension, otherwise it places a file named `download` in the folder, without an extension. Only found two ocurreces like this. 
Looks like it would be better practice to zip any file we host on AGOL

PR #390, opened 2019-07-22
Download preplanned map

JonLavi:
Hi both,
Now that all the new Offline features are wrapped, I'm submitting this PR for `Download Preplanned Map`.

Looking forward to seeing what you think.

PR #405, opened 2019-08-13
Tschie/preplanned map

tschie:
I thought that the implementation here https://github.com/Esri/arcgis-runtime-samples-java/pull/390 was too complex. Most of the logic was managing app state or showing mobile map package practices (closing, opening, and managing) rather than preplanned map practices.

I created this PR with what I think is a simpler alternative. It would have been too many changes to address via comments in a PR review. Please try it out and let me know what you think.

Note that this implementation allows you to download a preplanned area more than once. I think this is fine because allowing these areas to only be downloaded once was an arbitrary restriction in the sample design.

The readme and image are not up to date.


JonLavi (approved):
Looks nice, nothing to fault here 👍

JonLavi:
@tschie, @TADraeseke will implement it in a same way in Android, and we should put through a PR to update the common design.

map/download-preplanned-map/README.metadata.json
```diff
@@ -0,0 +1,29 @@
...
+    "redirect_from": "/java/latest/sample-code/download-preplanned-map.htm",
+    "relevant_apis": [
+        "DownloadPreplannedOfflineMapJob",
+        "DownloadPreplannedOfflineMapParameters",
+        "DownloadPreplannedOfflineMapResult",
+        "OfflineMapTask",
+        "PreplannedMapArea"
+    ],
+    "snippets": [
+        "src/main/java/com/esri/samples/download_preplanned_map/DownloadPreplannedMapController.java",
```
Rachael-E, line 23: I think that the `DownloadPreplannedOfflineMapJobListCell` should also be in this snippet list?

Rachael-E (approved):
@tschie this looks really good: still captures the essence of the original design but in a more streamlined and focussed way!  One suggestion, when a preplanned map area is clicked on to download, the red outlines of the areas disappear and only return again once another area is selected. I think it would be good if we could still see the red outlines on the map after an area has been downloaded, currently if you zoom out or pan around after download there are no red outlines to view.

PR #406, opened 2019-08-14
Create and save kml file

JonLavi:
Hi both,
I've implemented this sample similarly to .NET.
There's a fair bit of UI work again, and the FXML is fairly complex. I didn't see a way of showing and hiding the 'apply style' UI while keeping it in the same file, and still being able to click through it when it's invisible. I've opted for putting the 'apply style' UI directly above the other UI using AnchorPane, simply because we have to click on one of the options anyway. If you have any suggestions for refactoring/restructuring the whole sample let me know.
Cheers!


JonLavi:
Also, not entirely sure why some very old commits are showing up here, the files themselves don't appear in the changed list.

tschie:
In a mobile app, it makes sense to have the UI change like this because there is a lot going on. But on desktop, with a lot more screen space, I don't think the UI would ever be step-by-step like this with the UI changing all the time.

How about something like this:
create-save-kml

Click the sketch button to start sketching. The sketch button should be disabled after this. Complete the sketch and keep it with the Enter key or discard it with the Escape key.

The style combobox should be switched out with the one relevant to the geometry type selected. The sketch editor should use the symbol as well as the result feature when you complete the sketch. You should be able to see the sketch update on-the-fly when you change the style. You should be able to get back to the default style easily. Can the default symbol be shown in the combobox?

All you should need is a VBox on a StackPane. You can hide the style UI elements which are not relevant to the selected geometry type. Bind the managed property to the visible property so it doesn't take up space when hidden (should be able to do this in FXML).

JonLavi:
@tschie implemented your suggestions, and should be pretty close to your proposed solution. Only problem I seem to be having is that the `EventHandler<KeyEvent>` doesn't seem to catch while the `SketchEditor` is 'in focus'. As soon as you interact with either of the combo-boxes, the `EventHandler` reacts to the keyboard and you can finish/abort the sketch. Have you got any suggestions on how to fix this? Once this is sorted, we can disable the comboboxes as well while sketching.

tschie (approved):

PR #407, opened 2019-08-23
simplify UI

tschie:
I've proposed a few things to simplify the sample:
- the sketch editor is always active. It restarts whenever the sketch creation mode changes or a sketch is discarded or committed.
- sketch is committed or discarded with the ENTER and ESCAPE keys respectively.
- you can't 'reset' the KML document. You can restart the sample instead.

### DIFF
--- a/kml/create-and-save-kml-file/src/main/java/com/esri/samples/create_and_save_kml_file/CreateAndSaveKMLFileController.java
+++ b/kml/create-and-save-kml-file/src/main/java/com/esri/samples/create_and_save_kml_file/CreateAndSaveKMLFileController.java
@@ -51,6 +51,7 @@ public class CreateAndSaveKMLFileController {
 
   private KmlDocument kmlDocument;
   private SketchEditor sketchEditor;
+  private FileChooser fileChooser;
 
   @FXML
   public void initialize() {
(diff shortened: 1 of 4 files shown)


kml/create-and-save-kml-file/src/main/java/com/esri/samples/create_and_save_kml_file/CreateAndSaveKMLFileController.java
```diff
@@ -42,282 +25,176 @@
...
-      // write the KMZ file to the path chosen
-      if (kmzFile != null) {
-        kmlDocument.saveAsAsync(kmzFile.getPath());
-      } else {
-        new Alert(Alert.AlertType.WARNING, "KMZ file not saved.").show();
-      }
-    } catch (Exception e) {
-      new Alert(Alert.AlertType.ERROR, "Error saving KMZ file.").show();
+    // create a file chooser to get a path for saving the KMZ file
+    FileChooser fileChooser = new FileChooser();
```
JonLavi: Why not create the file chooser in initialize() and then reuse it?

JonLavi (approved):
Looking good - `stackPane.requestFocus();` was the bit I was missing previously.
Just curious why you chose to create a new FileChooser every time a save happens?

kml/create-and-save-kml-file/src/main/java/com/esri/samples/create_and_save_kml_file/CreateAndSaveKMLFileController.java
tschie: Good catch

PR #409, opened 2019-09-16
New Sample - Edit kml ground overlay

JonLavi:
Hi both,
I've implemented this fairly simple sample for editing a KML ground overlay.
Let me know if you spot any errors/enhancements. Not too happy with having to do both `slider.setOnMouseDragged` and `setOnMouseReleased`, but I'm not sure there's any better way to make the slider work more smoothly?
Thanks!


kml/edit-kml-ground-overlay/src/main/java/com/esri/samples/edit_kml_ground_overlay/EditKMLGroundOverlaySample.java
```diff
@@ -0,0 +1,151 @@
+/*
+ * Copyright 2018 Esri.
```
Rachael-E: should be 2019

kml/edit-kml-ground-overlay/src/main/java/com/esri/samples/edit_kml_ground_overlay/EditKMLGroundOverlaySample.java
```diff
@@ -0,0 +1,151 @@
...
+
+      // create a map and add it to the map view
+      ArcGISScene scene = new ArcGISScene(Basemap.createImagery());
+      sceneView = new SceneView();
+      sceneView.setArcGISScene(scene);
+
+      // create a slider for adjusting the rotation
+      Slider slider = new Slider(0, 1, 1);
+      slider.setMaxWidth(300);
+      slider.setShowTickLabels(true);
```
Rachael-E: we don't need to show tick labels or tick marks for this simple sample: showing 0 and 1 aren't too helpful to a user when it comes to opacity. I think that it's intuitive enough to have the slider without a legend, as shown in the original design.  If we had more complex data ranges, then showing tick labels on a slider would be more appropriate.

kml/edit-kml-ground-overlay/src/main/java/com/esri/samples/edit_kml_ground_overlay/EditKMLGroundOverlaySample.java
```diff
@@ -0,0 +1,151 @@
...
+      slider.setShowTickMarks(true);
```
Rachael-E: this can be removed too

kml/edit-kml-ground-overlay/src/main/java/com/esri/samples/edit_kml_ground_overlay/EditKMLGroundOverlaySample.java
```diff
@@ -0,0 +1,151 @@
...
+      // create a label for the slider
+      Label label = new Label("Overlay Opacity: ");
```
Rachael-E: The text and the slider are a little offset when aligned together on the same row. With this in mind, I'd prefer to see the layout as it is in the design: label above and then the slider below. I think this would look a little neater (and would follow for example the Terrain Exaggeration layout). I'd therefore suggest using a VBox rather than an HBox (might have to do some tweaking to get the start of the text to align with the start of the slider).

kml/edit-kml-ground-overlay/src/main/java/com/esri/samples/edit_kml_ground_overlay/EditKMLGroundOverlaySample.java
```diff
@@ -0,0 +1,151 @@
...
+      stage.setHeight(700);
+      stage.setScene(fxScene);
+      stage.show();
+
+      // create a map and add it to the map view
+      ArcGISScene scene = new ArcGISScene(Basemap.createImagery());
+      sceneView = new SceneView();
+      sceneView.setArcGISScene(scene);
+
+      // create a slider for adjusting the rotation
```
Rachael-E: I would move the instantiation of the slider to just above where it is first used, it's easier for a user to read that way.

kml/edit-kml-ground-overlay/src/main/java/com/esri/samples/edit_kml_ground_overlay/EditKMLGroundOverlaySample.java
Rachael-E: I would move the label and other UI set up here (the HBox/VBox set up) to the end of the code (before adding the map view and controls to the stack pane), so that the focus on the main body of the code is our API rather than JavaFX set up.

kml/edit-kml-ground-overlay/src/main/java/com/esri/samples/edit_kml_ground_overlay/EditKMLGroundOverlaySample.java
```diff
@@ -0,0 +1,151 @@
...
+      // create the KML ground overlay
+      KmlGroundOverlay kmlGroundOverlay = new KmlGroundOverlay(overlayGeometry, overlayImage);
+
+      // set the rotation of the ground overlay
+      kmlGroundOverlay.setRotation(-3.046024799346924);
+
+      // create a KML dataset with the ground overlay as the root node
+      KmlDataset kmlDataset = new KmlDataset(kmlGroundOverlay);
+
+      // create a KML layer for the scene view
```
Rachael-E: > create a KML layer using the KML dataset

kml/edit-kml-ground-overlay/src/main/java/com/esri/samples/edit_kml_ground_overlay/EditKMLGroundOverlaySample.java
```diff
@@ -0,0 +1,151 @@
...
+      KmlLayer kmlLayer = new KmlLayer(kmlDataset);
+
+      // add the KML layer to the scene
```
Rachael-E: > add the KML layer to the scene view

kml/edit-kml-ground-overlay/src/main/java/com/esri/samples/edit_kml_ground_overlay/EditKMLGroundOverlaySample.java
```diff
@@ -0,0 +1,151 @@
...
+      sceneView.getArcGISScene().getOperationalLayers().add(kmlLayer);
+
+      // move the viewpoint to the ground overlay
```
Rachael-E: > set the viewpoint to the ground overlay

kml/edit-kml-ground-overlay/src/main/java/com/esri/samples/edit_kml_ground_overlay/EditKMLGroundOverlaySample.java
Rachael-E: I think this comment is meant to be for adjusting the opacity, not the rotation?

kml/edit-kml-ground-overlay/src/main/java/com/esri/samples/edit_kml_ground_overlay/EditKMLGroundOverlaySample.java
```diff
@@ -0,0 +1,151 @@
...
+      sceneView.setViewpoint(new Viewpoint(kmlGroundOverlay.getExtent(), new Camera(kmlGroundOverlay.getGeometry().getExtent().getCenter(), 1250, 45, 60, 0)));
+
+      // add an event handler to the slider to apply opacity to the KML ground overlay based on the slider value
+      slider.setOnMouseReleased(event -> kmlGroundOverlay.setColor(ColorUtil.colorToArgb(new Color(0, 0, 0, slider.getValue()))));
```
Rachael-E: to avoid repetition, you could set up a new Color and apply that to both, e.g.

`Color opacity = new Color(0, 0, 0, slider.getValue());`

`slider.setOnMouseReleased(event -> kmlGroundOverlay.setColor(ColorUtil.colorToArgb(opacity)))`

Or even better, set up a method changeOpacity to pass to both events, or just a method which handles the slider mouse interactions and call that method here.

kml/edit-kml-ground-overlay/src/main/java/com/esri/samples/edit_kml_ground_overlay/EditKMLGroundOverlaySample.java
```diff
@@ -0,0 +1,151 @@
...
+      slider.setOnMouseDragged(event -> kmlGroundOverlay.setColor(ColorUtil.colorToArgb(new Color(0, 0, 0, slider.getValue()))));
+
+      // add the map view and controls to the stack pane
```
Rachael-E: map view -> scene view here

kml/edit-kml-ground-overlay/src/main/java/com/esri/samples/edit_kml_ground_overlay/EditKMLGroundOverlaySample.java
```diff
@@ -0,0 +1,151 @@
...
+      fxScene.getStylesheets().add(getClass().getResource("/style.css").toExternalForm());
+
+      // set title, size, and add scene to stage
+      stage.setTitle("Edit KML Ground Overlay Sample");
+      stage.setWidth(800);
+      stage.setHeight(700);
+      stage.setScene(fxScene);
+      stage.show();
+
+      // create a map and add it to the map view
```
Rachael-E: this sample is for a scene, so 

> create a scene and add it to the scene view

Rachael-E (changes requested):
@JonLavi thanks for pulling this together! I think we can improve ever so slightly on the UI, and there are also some code and comment refactorings that could tighten this sample up. 

And I think it's fine to add the two event listeners for the slider: we do the same in the Terrain Exaggeration sample, it's just the code for that is hidden in the FXML file.

kml/edit-kml-ground-overlay/src/main/java/com/esri/samples/edit_kml_ground_overlay/EditKMLGroundOverlaySample.java
JonLavi: Having to fiddle around with the veritcal alignment is exactly the reason I've gone for an HBox.
Changed to VBox, you'll see that the slider is vertically aligned with the label, *if* the slider button itself is brought all the way to the right. If it's not in the rightmost position, it will look slightly offset.

JonLavi:
Thanks @Rachael-E, all comments addressed.
I've added a method for updating the Opacity of a layer, as you suggested. However looking at it now, the previous way was slightly simpler (although somewhat repetitive).
Let me know what you think.

kml/edit-kml-ground-overlay/src/main/java/com/esri/samples/edit_kml_ground_overlay/EditKMLGroundOverlaySample.java
```diff
@@ -0,0 +1,162 @@
...
+      VBox controlsVBox = new VBox();
+      controlsVBox.setBackground(new Background(new BackgroundFill(Paint.valueOf("rgba(0,0,0,0.5)"), CornerRadii.EMPTY,
+              Insets.EMPTY)));
+      controlsVBox.setPadding(new Insets(10.0));
+      controlsVBox.setSpacing(10);
+      controlsVBox.setMaxSize(250, 20);
+      controlsVBox.getStyleClass().add("panel-region");
+
+      // create a label for the slider
+      Label label = new Label("Overlay Opacity: ");
```
Rachael-E: `Label label = new Label(" Overlay Opacity");`

Adding an extra space before "Overlay" will line up the text with the slider so that it's aligned, and we also don't need the colon.

kml/edit-kml-ground-overlay/src/main/java/com/esri/samples/edit_kml_ground_overlay/EditKMLGroundOverlaySample.java
```diff
@@ -0,0 +1,162 @@
...
+      StackPane.setMargin(controlsVBox, new Insets(10, 0, 0, 10));
+
+    } catch (Exception e) {
+      // on any error, display the stack trace.
+      e.printStackTrace();
+    }
+  }
+
+  /**
+   * Updates the color of a KmlGroundOverlay with a provided alpha value.
```
Rachael-E: to be more accurate to what this method is doing, we should change this to:

"Changes the opacity of a KmlGroundOverlay using a provided alpha value."

kml/edit-kml-ground-overlay/src/main/java/com/esri/samples/edit_kml_ground_overlay/EditKMLGroundOverlaySample.java
```diff
@@ -104,30 +87,58 @@ public void start(Stage stage) {
...
     }
   }
 
+  /**
+   * Updates the color of a KmlGroundOverlay with a provided alpha value.
+   *
+   * @param kmlGroundOverlay a provided KmlGroundOverlay.
+   * @param alphaValue       a double value representing the opacity to be applied.
+   */
+  private void updateGroundOverlayOpacity(KmlGroundOverlay kmlGroundOverlay, double alphaValue) {
```
Rachael-E: I think we should make it even simpler with the following:

```
private void setGroundOverlayOpacity() {
    kmlGroundOverlay.setColor(ColorUtil.colorToArgb(new Color(0, 0, 0, slider.getValue())));
}
```

And then have in the start method:

```
   // add an event handler to the slider to apply opacity to the KML ground overlay based on the slider value
      slider.setOnMouseReleased(event -> setGroundOverlayOpacity());
      slider.setOnMouseDragged(event -> setGroundOverlayOpacity());
```

Rachael-E (approved):
thanks @JonLavi for making those changes, I've suggested a simplification for the slider event handler and for making the text and slider line up. Once the screenshot has also been updated to reflect these review changes, this is ready for second review from @tschie, thanks!

kml/edit-kml-ground-overlay/src/main/java/com/esri/samples/edit_kml_ground_overlay/EditKMLGroundOverlaySample.java
colinanderson: I don't think adding a space is going to be a very robust way to line up controls. Probably you can set some kind of alignment property on the `VBox` - https://openjfx.io/javadoc/11/javafx.graphics/javafx/scene/layout/VBox.html#setAlignment(javafx.geometry.Pos) maybe.

Do you test the samples on a right to left layout?

Rachael-E: that's true, I keep forgetting we have other platforms like Mac that can display text alignment differently. So a padding option or the above could be a better approach. If that still proves fiddly, perhaps centre align the label text above the slider to avoid the issue.

Rachael-E: and no, we don't usually test on a right to left layout, at least not that I'm aware of: not sure if this is something @hardikc07 tests for.

colinanderson: Using RTL is a good way to see if the controls are laid out properly e.g. the extra space would possibly end up on the wrong side.

JonLavi: The previous version had lefthand-alignment.
After playing around with it a little I've opted to just center both elements in the VBbox, so it's LH/RH agnostic.

JonLavi:
Thanks @Rachael-E , all suggestions implemented as far as possible.

kml/edit-kml-ground-overlay/src/main/java/com/esri/samples/edit_kml_ground_overlay/EditKMLGroundOverlaySample.java
```diff
@@ -48,7 +48,9 @@
 
 public class EditKMLGroundOverlaySample extends Application {
 
+  private KmlGroundOverlay kmlGroundOverlay;
   private SceneView sceneView;
+  private Slider slider;
 
   @Override
   public void start(Stage stage) {
```
Rachael-E, line 54: for the comment on line 64, we should clarify that the scene is an application cene to avoid confusion with the comment on line 71. Suggested: 
// set title, size and add applicationn scene to stage

Rachael-E (approved):
thanks @JonLavi, just one very minor clarification for a comment: this is ready for second review now

kml/edit-kml-ground-overlay/src/main/java/com/esri/samples/edit_kml_ground_overlay/EditKMLGroundOverlaySample.java
JonLavi: Thanks, I sometimes mix this up myself. Good spot!

kml/edit-kml-ground-overlay/src/main/java/com/esri/samples/edit_kml_ground_overlay/EditKMLGroundOverlaySample.java
```diff
@@ -0,0 +1,162 @@
...
+
+      // set the viewpoint to the ground overlay
+      sceneView.setViewpoint(new Viewpoint(kmlGroundOverlay.getExtent(), new Camera(kmlGroundOverlay.getGeometry().getExtent().getCenter(), 1250, 45, 60, 0)));
+
+      // create a slider for adjusting the overlay opacity
+      slider = new Slider(0, 1, 1);
+      slider.setMaxWidth(300);
+
+      // add an event handler to the slider to apply opacity to the KML ground overlay based on the slider value
+      slider.setOnMouseReleased(event -> setGroundOverlayOpacity());
```
tschie: I think this should work. Then you don't need the separate method or fields for the KmlGroundOverlay and Slider.
```java
slider.valueProperty().addListener(o ->
    kmlGroundOverlay.setColor(ColorUtil.colorToArgb(new Color(0, 0, 0, slider.getValue())))
);
```

kml/edit-kml-ground-overlay/src/main/java/com/esri/samples/edit_kml_ground_overlay/EditKMLGroundOverlaySample.java
```diff
@@ -0,0 +1,162 @@
...
+      Envelope overlayGeometry = new Envelope(-123.066227926904, 44.04736963555683, -123.0796942287304, 44.03878298600624, SpatialReferences.getWgs84());
+
+      // create a KML icon for the overlay image
+      String overlayImageURI = new File("./samples-data/images/1944.jpg").getAbsolutePath();
+      KmlIcon overlayImage = new KmlIcon(overlayImageURI);
+
+      // create the KML ground overlay
+      kmlGroundOverlay = new KmlGroundOverlay(overlayGeometry, overlayImage);
+
+      // set the rotation of the ground overlay
```
tschie: Might want to comment on why this is needed. To correct the overlays rotation with respect to the basemap?

kml/edit-kml-ground-overlay/src/main/java/com/esri/samples/edit_kml_ground_overlay/EditKMLGroundOverlaySample.java
```diff
@@ -0,0 +1,162 @@
...
+      KmlDataset kmlDataset = new KmlDataset(kmlGroundOverlay);
+
+      // create a KML layer using the KML dataset
+      KmlLayer kmlLayer = new KmlLayer(kmlDataset);
+
+      // add the KML layer to the scene view
+      sceneView.getArcGISScene().getOperationalLayers().add(kmlLayer);
+
+      // set the viewpoint to the ground overlay
+      sceneView.setViewpoint(new Viewpoint(kmlGroundOverlay.getExtent(), new Camera(kmlGroundOverlay.getGeometry().getExtent().getCenter(), 1250, 45, 60, 0)));
```
tschie: Can you reuse `overlayGeometry` instead of `kmlGroundOverlay.getExtent()` and `kmlGroundOverlay.getGeometry().getExtent()`?

kml/edit-kml-ground-overlay/README.md
```diff
@@ -0,0 +1,39 @@
...
+KML ground overlays are used for showing aerial imagery, symbology, or other images draped over a scene. Changing the geometry, rotation, and other attributes of a ground overlay after it has been loaded allows for live editing.  For example, editing the geometry and opacity of a historical image draped over present day satellite imagery makes it possible to view change over time.
+
+## How to use the sample
+
+Use the slider to adjust the opacity of the ground overlay.
+
+## How it works
+
+1. Create an `Envelope` defining the geometry of the overlay.
+2. Create a `KmlIcon` using a `Uri` linking to an image.
```
tschie: Don't think Uri should be in backticks since it's not a class.

kml/edit-kml-ground-overlay/README.md
```diff
@@ -0,0 +1,39 @@
...
+3. Create a `KmlGroundOverlay` using the envelope and icon.
+4. Use `kmlGroundOverlay.setRotation()` to set the rotation of the ground overlay.
```
tschie: Is this super common that it needs to be listed here? I would think the extent would usually get the correct placement.

kml/edit-kml-ground-overlay/README.md
```diff
@@ -0,0 +1,39 @@
...
+5. Create a `KmlDataset` using the ground overlay.
+6. Create a `KmlLayer` using the dataset.
+7. Add the KML layer to the scene.
+8. Listen for changes to the opacity slider and use `kmlGroundOverlay.setColor(value)` to change the value appropriately.
```
tschie: In other readmes, we've avoided using sample-specific elements in this section. Can we just say "Use `kmlGroundOverlay.setColor(value)` with a different alpha value to change the opacity."?

tschie (changes requested):
Nice simple sample. Looks really good. Just a few suggestions.

kml/edit-kml-ground-overlay/src/main/java/com/esri/samples/edit_kml_ground_overlay/EditKMLGroundOverlaySample.java
JonLavi: perfect, thanks!

kml/edit-kml-ground-overlay/README.md
JonLavi: Good point, it's just a small fix for the data used here, so it can go.

JonLavi:
Thanks for the suggestions @tschie , all applied.

tschie (approved):
Good to merge!

JonLavi:
Thanks @tschie , good catch on the local variables.

PR #413, opened 2019-10-21
Data parent directory

tschie:
This PR adds the parent directory argument to every File constructor. This allows one to use the `data.dir` system property to specify the location of the samples-data folder. This is useful for when you want to run a jar version of the sample from a directory which does not contain the samples-data folder.

### DIFF
--- a/analysis/viewshed-geoelement/src/main/java/com/esri/samples/viewshed_geoelement/ViewshedGeoElementSample.java
+++ b/analysis/viewshed-geoelement/src/main/java/com/esri/samples/viewshed_geoelement/ViewshedGeoElementSample.java
@@ -113,8 +113,10 @@ public class ViewshedGeoElementSample extends Application {
       graphicsOverlay.setRenderer(renderer3D);
 
       // create a graphic of a tank
-      String modelURI = new File("./samples-data/bradley_low_3ds/bradle.3ds").getAbsolutePath();
-      ModelSceneSymbol tankSymbol = new ModelSceneSymbol(modelURI, 10.0);
+      String modelURI =
+              new File(System.getProperty("data.dir"), "./samples-data/bradley_low_3ds/bradle.3ds").getAbsolutePath();
+      System.out.println(modelURI);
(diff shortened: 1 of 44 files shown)


analysis/viewshed-geoelement/src/main/java/com/esri/samples/viewshed_geoelement/ViewshedGeoElementSample.java
```diff
@@ -113,8 +113,10 @@ public void start(Stage stage) {
       graphicsOverlay.setRenderer(renderer3D);
 
       // create a graphic of a tank
-      String modelURI = new File("./samples-data/bradley_low_3ds/bradle.3ds").getAbsolutePath();
-      ModelSceneSymbol tankSymbol = new ModelSceneSymbol(modelURI, 10.0);
+      String modelURI =
+              new File(System.getProperty("data.dir"), "./samples-data/bradley_low_3ds/bradle.3ds").getAbsolutePath();
+      System.out.println(modelURI);
```
JonLavi: Probably don't need the sout here?

JonLavi (changes requested):
LGTM. Found one `System.out.println` that's probably not needed - leftover from testing this?
I'm also not sure why it's necessary to use `.getCanonicalPath()` in some cases dealing with tile caches/ geodatabases? Looks like replacing with `.getAbsolutePath()` works OK for some of the samples (Edit and Sync Features, Offline Geocode), but I haven't tried all of them. What's happening here?

Rachael-E (approved):
looks good @tschie, looks like there's just one instance where we're missing the parent directory argument in, if you could update this one too that would be great: 

**DictionaryRendererGraphicsOverlaySample.java**, line 105
`File mil2525dFile = new File("./samples-data/xml/Mil2525DMessages.xml");`

tschie:
@JonLavi `canonicalPath` isn't needed. It just fixes relative paths so you have "git/myProject/samples-data/something" instead of "git/myProject/./samples-data/something". I was using it to check if the relative path was causing a problem. I've removed use of it for now.

JonLavi (approved):
Good stuff!

PR #414, opened 2019-10-28
Create and save kml file (#406) [into master]

JonLavi:
Hi @tschie & @Rachael-E ,
I've noticed that during the transition between U6 and U7 development, we've targeted some new samples into the `dev` branch. Because these were only merged after we officially started working on U7, some samples ended up living in `dev` but never actually being merged into `master` - as a result not being in the Sample Viewer or on the developer webpage.
I've looked at a diff for direct merge of `dev` into `master`, but there's quite a few smaller changes.
So instead, I'm opening this PR (along with two others), where I've simply cherry-picked the exact commit that adds each sample.
This PR brings CreateAndSaveKMLFile into `master`, as it currently is in `dev`.
Mind that all this code has been reviewed and merged before, so I think the most we need to do is look at the file names, that nothing snuck in.
Thanks!


tschie (approved):

kml/create-and-save-kml-file/README.md
```diff
@@ -0,0 +1,48 @@
+# Create and save KML file
+
+Construct a KML document and save it as a KMZ file.
+
+![Create And Save KML File Sample](CreateAndSaveKMLFile.png)
```
Rachael-E, line 5: we should probably remove this description for now with the sample viewer in mind: same for the other PRs I meant to add.

PR #416, opened 2019-10-28
preplanned map (#405) [into master]

JonLavi:
Hi @tschie & @Rachael-E ,
I've noticed that during the transition between U6 and U7 development, we've targeted some new samples into the dev branch. Because these were only merged after we officially started working on U7, some samples ended up living in dev but never actually being merged into master - as a result not being in the Sample Viewer or on the developer webpage.
I've looked at a diff for direct merge of dev into master, but there's quite a few smaller changes.
So instead, I'm opening this PR (along with two others), where I've simply cherry-picked the exact commit that adds each sample.
This PR brings `DownloadPreplannedMap` into `master`, as it currently is in `dev`.
Mind that all this code has been reviewed and merged before, so I think the most we need to do is look at the file names, that nothing snuck in.
Thanks!


tschie (approved):

map/download-preplanned-map/src/main/java/com/esri/samples/download_preplanned_map/PreplannedMapAreaListCell.java
```diff
@@ -0,0 +1,40 @@
+/* Copyright 2019 Esri
```
Rachael-E, line 1: one comment for this: was the name of this class supposed to be
`DownloadPreplannedMapAreaListCell` ?

JonLavi: I think the name is correct as is `PreplannedMapAreaListCell`, because it just represents a cell item which holds a preplanned map area. It doesn't have anything to do directly with downloading the areas.

map/download-preplanned-map/README.md
```diff
@@ -0,0 +1,54 @@
+# Download a preplanned map area
+
+Take a map offline using a preplanned map area.
+
+![Download Preplanned Map Sample](DownloadPreplannedMap.png)
```
Rachael-E, line 5: should remove description with sample viewer in mind?

map/download-preplanned-map/src/main/java/com/esri/samples/download_preplanned_map/PreplannedMapAreaListCell.java
Rachael-E: nice one, just wanted to check thanks!

Rachael-E (approved):

PR #419, opened 2019-11-07
New Sample - Find connected features in utility network

JonLavi:
Hi both,

Here's the first sample for the Utility Network API. It's quite a big one, so no rush to review.
I followed the .NET sample design fairly closely, but of course there might be different ways to pull this off.

Note UtilityNetwork in U6 has a small bug causing the first trace to not return the complete set of results. This has been fixed for U7. The plan is to keep this sample in Dev until U7 ships, and it'll work just fine then.
~~Note there's a workaround in there (highlighted in the comments) for a bug that will be fixed in the next release, so we will remove the workaround again.~~

~~Also, I've originally branched this off `dev`, therefore targeting at `dev`, but shouldn't be a problem to rebase to `master`. Your call @tschie~~

Let me know what you think!

Thanks


network_analysis/find-connected-features-in-utility-network/README.md
```diff
@@ -0,0 +1,48 @@
...
+
+![](ConnectedTrace.JPG)
+
+## Use case
+
+This is useful to visualize and validate the network topology of a utility network for quality assurance. 
+
+## How to use the sample
+
+To add a starting point, select 'Add starting location(s)' and tap on one or more features. To add a barrier, select 'Add barrier(s)' and tap on one or more features. Depending on the type of feature, you may be prompted to select a terminal or the distance from the tapped location will be computed. Click 'Trace' to highlight all features connected to the specified starting locations and not positioned beyond the barriers. Click 'Reset' to clear parameters and start over.
```
Rachael-E: tap -> click (2x occurrences)
tapped -> clicked

Rachael-E: under what circumstances should the prompt for selecting a terminal occur? I think it would be helpful to have these listed. Looks like only when you click on a utility terminal to select if high or low?

network_analysis/find-connected-features-in-utility-network/README.md
```diff
@@ -0,0 +1,48 @@
...
+
+## How it works
+
+1. Create a `MapView` and subscribe to its `GeoViewTapped` event.
```
Rachael-E: I think this is .NET speak. Let's make it more java-y

network_analysis/find-connected-features-in-utility-network/README.md
```diff
@@ -0,0 +1,48 @@
...
+2. Create and load a `Map` that contains `FeatureLayer`(s) that are part of a utility network.
```
Rachael-E: we're not loading a map that contains a feature layer: we're getting a feature layer via a URL and adding it to the operational layers of the map

network_analysis/find-connected-features-in-utility-network/README.md
```diff
@@ -0,0 +1,48 @@
...
+3. Create and load a `UtilityNetwork` with the same feature service URL and map.
+4. Add a `GraphicsOverlay` with symbology that distinguishes starting points from barriers.
+5. Identify features on the map and add a `Graphic` that represents its purpose (starting point or barrier) at the location of each identified feature.
+6. Determine the type of the identified feature using `UtilityNetwork.Definition.GetNetworkSource` passing its table name.
```
Rachael-E: this looks like .NET doc, update for java

`UtilityNetwork.getDefinition.getNetworkSource`

network_analysis/find-connected-features-in-utility-network/README.md
```diff
@@ -0,0 +1,48 @@
...
+7. If a junction, display a terminal picker when more than one terminal is found and create a `UtilityElement` with the selected terminal or the single terminal if there is only one.
+8. If an edge, create a `UtilityElement` from the identified feature and compute its `FractionAlongLine` using `GeometryEngine.FractionAlong`.
```
Rachael-E: compute -> set
I can't see `FractionAlongLine`. Do you mean `FractionAlongEdge`?

Ensure you're using java specific doc `GeometryEngine.fractionAlong`

network_analysis/find-connected-features-in-utility-network/README.md
```diff
@@ -0,0 +1,48 @@
...
+9. Run a `UtilityNetwork.TraceAsync` with the specified parameters.
```
Rachael-E: `UtilityNetwork.traceAsync`

Can we elaborate on the specified parameters?

network_analysis/find-connected-features-in-utility-network/README.md
```diff
@@ -0,0 +1,48 @@
...
+10. Group the `UtilityElementTraceResult.Elements` by `NetworkSource.Name`
```
Rachael-E: Again, check for java naming not .NET
`UtilityElementTraceResults.getElements`
`UtilityNetworkSource.getName`

network_analysis/find-connected-features-in-utility-network/README.md
```diff
@@ -0,0 +1,48 @@
...
+11. For every `FeatureLayer` in this map with elements, select features by converting `UtilityElement`(s) to `ArcGISFeature`(s) using `UtilityNetwork.GetFeaturesFromElementsAsync`
```
Rachael-E: java naming

network_analysis/find-connected-features-in-utility-network/README.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Relevant API
+
+* UtilityNetworks.UtilityNetwork
```
Rachael-E: check java naming for all of these: We don't need the UtilityNetworks call before the class names.

network_analysis/find-connected-features-in-utility-network/README.md
```diff
@@ -0,0 +1,48 @@
...
+* UtilityNetworks.UtilityNetworkDefinition
+* UtilityNetworks.UtilityNetworkSource
+* UtilityNetworks.UtilityAssetType
+* UtilityNetworks.UtilityTerminal
+* UtilityNetworks.UtilityElement
+* GeometryEngine.FractionAlong
+
+## About the data
+
+The sample uses a dark vector basemap. It includes a subset of feature layers from a feature service that contains the same utility network used to run the connected trace.
```
Rachael-E: what is meant by "the same utility network": the same as what?

network_analysis/find-connected-features-in-utility-network/README.md
```diff
@@ -0,0 +1,48 @@
+# Find connected features in a utility network
+
+Find all features connected to a set of starting points in a utility network.
+
+![](ConnectedTrace.JPG)
```
Rachael-E: I think this should be .png

network_analysis/find-connected-features-in-utility-network/README.metadata.json
```diff
@@ -0,0 +1,21 @@
+{
+    "category": "Network analysis",
+    "description": "Find all features connected to a set of starting points in a utility network.",
+    "ignore": false,
+    "images": [
+        "ConnectedTrace.png"
```
Rachael-E: .png here is fine if updated in README

network_analysis/find-connected-features-in-utility-network/README.metadata.json
```diff
@@ -0,0 +1,21 @@
...
+    ],
+    "keywords": [
+        "connected trace",
+        "utility network",
+        "network analysis"
+    ],
+    "relevant_apis": [
```
Rachael-E: need to add in the full list of relevant apis

network_analysis/find-connected-features-in-utility-network/README.metadata.json
```diff
@@ -0,0 +1,21 @@
...
+        "UtilityNetworks",
+        "GeometryEngine"
+    ],
+    "snippets": [
```
Rachael-E: I think we include all snippets now aside from the launcher class?

network_analysis/find-connected-features-in-utility-network/src/main/java/com/esri/samples/find_connected_features_in_utility_network/FindConnectedFeaturesInUtilityNetworkController.java
```diff
@@ -0,0 +1,503 @@
...
+import com.esri.arcgisruntime.utilitynetworks.UtilityTraceType;
+
+public class FindConnectedFeaturesInUtilityNetworkController {
+
+  @FXML private Button resetButton;
+  @FXML private Button traceButton;
+  @FXML private Label statusLabel;
+  @FXML private MapView mapView;
+  @FXML private ProgressIndicator progressIndicator;
+  @FXML private RadioButton addingStartRadioButton;
```
Rachael-E: startingLocationsRadioButton is probably a better name

network_analysis/find-connected-features-in-utility-network/src/main/java/com/esri/samples/find_connected_features_in_utility_network/FindConnectedFeaturesInUtilityNetworkController.java
```diff
@@ -0,0 +1,503 @@
...
+      mapView.setMap(map);
+      mapView.setViewpointAsync(new Viewpoint(new Envelope(
+        -9813547.35557238,
+        5129980.36635111,
+        -9813185.0602376,
+        5130215.41254146,
+        SpatialReferences.getWebMercator())));
+
+      // create symbols for the starting point and barriers
+      startingPointSymbol = new SimpleMarkerSymbol(SimpleMarkerSymbol.Style.CROSS, ColorUtil.colorToArgb(Color.GREEN), 20);
```
Rachael-E: add comments to describe what each symbol is

e.g. green cross, red cross.

network_analysis/find-connected-features-in-utility-network/src/main/java/com/esri/samples/find_connected_features_in_utility_network/FindConnectedFeaturesInUtilityNetworkController.java
```diff
@@ -0,0 +1,503 @@
...
+   */
+  @FXML
+  private void handleMapViewClicked(MouseEvent e) {
+    if (e.getButton() == MouseButton.PRIMARY && e.isStillSincePress()) {
+
+      // show the progress indicator
+      progressIndicator.setVisible(true);
+
+      // set whether the user is adding a starting point or a barrier
+      boolean isAddingStart = addingStartRadioButton.isSelected();
```
Rachael-E: I think `isAddingStartingPoint` is probably clearer.

network_analysis/find-connected-features-in-utility-network/src/main/java/com/esri/samples/find_connected_features_in_utility_network/FindConnectedFeaturesInUtilityNetworkController.java
```diff
@@ -0,0 +1,503 @@
...
+
+      // identify the feature to be used
+      ListenableFuture<List<IdentifyLayerResult>> identifyLayerResultsFuture = mapView.identifyLayersAsync(screenPoint, 10, false);
+      identifyLayerResultsFuture.addDoneListener(() -> {
+        try {
+          // get the result of the query
+          List<IdentifyLayerResult> identifyLayerResults = identifyLayerResultsFuture.get();
+
+          // return if no features are identified
+          if (identifyLayerResults.isEmpty()) {
```
Rachael-E: can make this `if (!identifyLayerResults.isEmpty())` and that avoids having an unnecessary `return`

network_analysis/find-connected-features-in-utility-network/src/main/java/com/esri/samples/find_connected_features_in_utility_network/FindConnectedFeaturesInUtilityNetworkController.java
```diff
@@ -0,0 +1,503 @@
...
+            return;
+
+          } else {
+            // retrieve the first result and get it's contents
```
Rachael-E: it's -> its

network_analysis/find-connected-features-in-utility-network/src/main/java/com/esri/samples/find_connected_features_in_utility_network/FindConnectedFeaturesInUtilityNetworkController.java
```diff
@@ -0,0 +1,503 @@
...
+            // check that the result is a feature layer and has elements
+            if (layerContent instanceof FeatureLayer && !firstResult.getElements().isEmpty()) {
+              // retrieve the geoelements in the feature layer
+              GeoElement identifiedElement = firstResult.getElements().get(0);
+              if (identifiedElement instanceof ArcGISFeature) {
+                // get the feature
+                ArcGISFeature identifiedFeature = (ArcGISFeature) identifiedElement;
+
+                // get the network source of the identified feature
+                UtilityNetworkSource networkSource = utilityNetwork.getDefinition().getNetworkSource(identifiedFeature.getFeatureTable().getTableName());
```
Rachael-E: there are quite a lot of chained methods. Can we split it up into local variables for readability?

network_analysis/find-connected-features-in-utility-network/src/main/java/com/esri/samples/find_connected_features_in_utility_network/FindConnectedFeaturesInUtilityNetworkController.java
```diff
@@ -0,0 +1,503 @@
...
+    }
+  }
+
+  /**
+   * Uses the clicked map point to identify any utility elements in the utility network at the clicked location. Based
+   * on the selection mode, the clicked utility element is added either to the starting locations or barriers for the
+   * trace parameters. An appropriate graphic is created at the clicked location to mark the element as either a
+   * starting location or barrier.
+   *
+   * @param e mouse event registered when the map view is clicked on.
```
Rachael-E: don't need a full stop after a param description, only need punctuation in the description of the method.

network_analysis/find-connected-features-in-utility-network/src/main/java/com/esri/samples/find_connected_features_in_utility_network/FindConnectedFeaturesInUtilityNetworkController.java
```diff
@@ -0,0 +1,503 @@
...
+          progressIndicator.setVisible(false);
+        }
+      });
+    }
+  }
+
+  /**
+   * Uses a UtilityNetworkSource to create a UtilityElement object out of an ArcGISFeature.
+   *
+   * @param identifiedFeature an ArcGISFeature object that will be used to create a UtilityElement.
```
Rachael-E: don't need full stops for params or returns

network_analysis/find-connected-features-in-utility-network/src/main/java/com/esri/samples/find_connected_features_in_utility_network/FindConnectedFeaturesInUtilityNetworkController.java
```diff
@@ -0,0 +1,503 @@
...
+
+    // find the code matching the asset group name in the feature's attributes
+    int assetGroupCode = (int) attributes.get(assetGroupFieldName.toLowerCase());
+
+    // iterate through the network source's asset groups to find the group with the matching code
+    List<UtilityAssetGroup> assetGroups = networkSource.getAssetGroups();
+    for (UtilityAssetGroup assetGroup : assetGroups) {
+      if (assetGroup.getCode() == assetGroupCode) {
+
+        // get the code for the feature's asset type from it's attributes
```
Rachael-E: it's -> its

network_analysis/find-connected-features-in-utility-network/src/main/java/com/esri/samples/find_connected_features_in_utility_network/FindConnectedFeaturesInUtilityNetworkController.java
```diff
@@ -0,0 +1,503 @@
...
+                utilityElementGroups.put(networkSourceName, list);
+              } else {
+                utilityElementGroups.get(networkSourceName).add(utilityElement);
+              }
+            });
+
+            // get the feature layer for the utility element
+            utilityElementGroups.forEach((networkSourceName, utilityElements) -> {
+
+              // load each element's feature table manually. Note this is a workaround for some features that are returned by the trace not being selected properly
```
Rachael-E: I don't understand the workaround comment. What is meant by this? 
That some features returned by the trace don't select properly: in what way? They don't select at all?

network_analysis/find-connected-features-in-utility-network/src/main/java/com/esri/samples/find_connected_features_in_utility_network/FindConnectedFeaturesInUtilityNetworkController.java
```diff
@@ -0,0 +1,503 @@
...
+              FeatureTable utilityElementsFT = utilityElements.get(0).getNetworkSource().getFeatureTable();
+              utilityElementsFT.loadAsync();
+              utilityElementsFT.addDoneLoadingListener(()->{
```
Rachael-E:
```suggestion
              utilityElementsFT.addDoneLoadingListener(() -> {
```

network_analysis/find-connected-features-in-utility-network/src/main/java/com/esri/samples/find_connected_features_in_utility_network/FindConnectedFeaturesInUtilityNetworkController.java
```diff
@@ -0,0 +1,503 @@
...
+                          "Error fetching the corresponding features for the utility elements.").show();
+                      }
+                    });
+                  }
+                } else {
+                  new Alert(Alert.AlertType.ERROR, "Error loading Utility Element Feature Layer.").show();
+                }
+              });
+            });
+
```
Rachael-E: can remove new line

network_analysis/find-connected-features-in-utility-network/src/main/java/com/esri/samples/find_connected_features_in_utility_network/FindConnectedFeaturesInUtilityNetworkController.java
```diff
@@ -0,0 +1,503 @@
...
+    traceButton.setDisable(false);
+    resetButton.setDisable(false);
+
+    // hide the progress indicator
+    progressIndicator.setVisible(false);
+  }
+
+  /**
+   * Restores the sample to the startup-state by resetting the status text, hiding the progress indicator, clearing
+   * the trace parameters, de-selecting all features and removing any graphics
```
Rachael-E: full stop at the end of the method description

network_analysis/find-connected-features-in-utility-network/src/main/java/com/esri/samples/find_connected_features_in_utility_network/UtilityTerminalSelectionDialog.java
```diff
@@ -0,0 +1,76 @@
...
+    loader.setController(this);
+
+    setTitle("Select Utility Terminal:");
+
+    this.getDialogPane().setContent(utilityTerminalListView);
+
+    try {
+      loader.load();
+
+      // use a cell factory which shows the utility terminals's title
```
Rachael-E:
```suggestion
      // use a cell factory which shows the utility terminal's title
```

network_analysis/find-connected-features-in-utility-network/src/main/resources/find_connected_features_in_utility_network.fxml
```diff
@@ -0,0 +1,69 @@
...
+                </fills>
+            </Background>
+        </background>
+
+        <RadioButton fx:id="addingStartRadioButton" text="Add Starting Location(s)" selected="true">
+            <toggleGroup>
+                <ToggleGroup fx:id="toggleGroup"/>
+            </toggleGroup>
+        </RadioButton>
+        <RadioButton fx:id="addingBarrierRadioButton" toggleGroup="$toggleGroup" text="Add Barriers(s)"/>
```
Rachael-E: I must be missing something here, but I can't see the instantiation of the FXML object `addingBarrierRadioButton` in the Controller class in this PR?  Does it need to have an fx id?

network_analysis/find-connected-features-in-utility-network/src/main/resources/terminal_selection_dialog.fxml
```diff
@@ -0,0 +1,36 @@
...
+<?import javafx.scene.layout.StackPane?>
+<fx:root type="javafx.scene.control.Dialog" xmlns:fx="http://javafx.com/fxml">
+    <dialogPane>
+        <DialogPane>
+            <content>
+                <StackPane>
+                    <ListView fx:id="utilityTerminalListView" maxHeight="100"/>
+                </StackPane>
+            </content>
+            <ButtonType fx:id="cancelButton" buttonData="CANCEL_CLOSE" text="Cancel"/>
```
Rachael-E: same with cancelButton, can't see it's FXML object instantiation anywhere? Does it need to have an fx:id?

network_analysis/routing-around-barriers/src/main/resources/style.css
```diff
@@ -1,27 +1,3 @@
...
-.range-slider .axis {
-    -fx-tick-label-fill: white;
-}
-
-.panel-region .check-box {
-   -fx-text-fill: white;
-}
-
 .panel-region .radio-button {
    -fx-text-fill: white;
```
Rachael-E: is this change for Routing Around Barriers intended for this PR? We should probably stick to making changes only to one sample at a time.

network_analysis/find-connected-features-in-utility-network/src/main/resources/find_connected_features_in_utility_network.fxml
Rachael-E:
```suggestion
        <RadioButton fx:id="addingBarrierRadioButton" toggleGroup="$toggleGroup" text="Add Barrier(s)"/>
```

network_analysis/find-connected-features-in-utility-network/src/main/java/com/esri/samples/find_connected_features_in_utility_network/FindConnectedFeaturesInUtilityNetworkController.java
```diff
@@ -0,0 +1,503 @@
...
+                else if (networkSource.getSourceType() == UtilityNetworkSource.Type.EDGE && identifiedFeature.getGeometry().getGeometryType() == GeometryType.POLYLINE) {
+
+                  //  create a utility element with the identified feature
+                  utilityElement = utilityNetwork.createElement(identifiedFeature, null);
+
+                  // get the geometry of the identified feature as a polyline, and remove the z component
+                  Polyline polyline = (Polyline) GeometryEngine.removeZ(identifiedFeature.getGeometry());
+
+                  // compute how far the clicked location is along the edge feature
+                  utilityElement.setFractionAlongEdge(GeometryEngine.fractionAlong(polyline, mapPoint, -1));
```
Rachael-E: I've noticed when clicking on add a starting location, and then clicking in a location (shown on the screen shot) it throws an ArcGISRuntimeException error:
`Invalid argument: Fraction along value must be in [0,1] range`. It looks like for some reason this location returns a NaN which I think is causing the error. @colinanderson I think this could this be a Java bug if we're allowing NaN to be passed in?

Rachael-E: image
This is the location (it should be on the screen when loading the app the first time)

Rachael-E: Note: this happens at other joins too

network_analysis/find-connected-features-in-utility-network/src/main/java/com/esri/samples/find_connected_features_in_utility_network/FindConnectedFeaturesInUtilityNetworkController.java
```diff
@@ -0,0 +1,503 @@
...
+
+    } catch (Exception e) {
+      e.printStackTrace();
+    }
+  }
+
+  /**
+   * Uses the clicked map point to identify any utility elements in the utility network at the clicked location. Based
+   * on the selection mode, the clicked utility element is added either to the starting locations or barriers for the
+   * trace parameters. An appropriate graphic is created at the clicked location to mark the element as either a
```
Rachael-E:
```suggestion
   * trace parameters. The appropriate graphic is created at the clicked location to mark the element as either a
```

network_analysis/find-connected-features-in-utility-network/src/main/java/com/esri/samples/find_connected_features_in_utility_network/FindConnectedFeaturesInUtilityNetworkController.java
```diff
@@ -0,0 +1,503 @@
...
+                UtilityTerminal selectedTerminal = selectedTerminalOptional.get();
+                utilityElement = utilityNetwork.createElement(identifiedFeature, selectedTerminal);
+                showTerminalNameInStatusLabel(selectedTerminal);
+              }
+            }
+          }
+        }
+      }
+    }
+
```
Rachael-E: can remove new line here

network_analysis/find-connected-features-in-utility-network/src/main/java/com/esri/samples/find_connected_features_in_utility_network/FindConnectedFeaturesInUtilityNetworkController.java
```diff
@@ -0,0 +1,503 @@
...
+        statusLabel.setText("Trace failed.");
+        new Alert(Alert.AlertType.ERROR, "Error running utility network connected trace.").show();
+      }
+    });
+  }
+
+  /**
+   * Enables both buttons and hides the progress indicator.
+   */
+  private void enableUI() {
```
Rachael-E: this could be why the progress indicator disappears before the trace is complete? If you remove it from here then the method could be renamed "enableButtonInteraction" or similar, "enableUI" isn't too clear from its name what it does.

network_analysis/find-connected-features-in-utility-network/src/main/java/com/esri/samples/find_connected_features_in_utility_network/FindConnectedFeaturesInUtilityNetworkController.java
```diff
@@ -0,0 +1,503 @@
...
+    // enable the UI
+    traceButton.setDisable(false);
+    resetButton.setDisable(false);
+
+    // hide the progress indicator
+    progressIndicator.setVisible(false);
+  }
+
+  /**
+   * Restores the sample to the startup-state by resetting the status text, hiding the progress indicator, clearing
```
Rachael-E:
```suggestion
   * Resets the sample by resetting the status text, hiding the progress indicator, clearing
```

Rachael-E (changes requested):
@JonLavi thanks, looking good though some comments to address. Also we can chat about the following points if you like:

- I think we can make the HBox a bit shorter. Currently it has quite a lot of empty space at the bottom.
- When clicking "Trace" after setting a starting location and a barrier, a progress indicator comes up which is helpful. However it disappears before it has finished finding connected features. Can we fix this so the progress indicator remains on until the trace is complete? Also the "Trace" and "Reset" buttons should remain disabled until the trace is complete.
- Rather than clicking "Trace" with no starting locations or barriers set and getting an error coming up that no starting locations were added, we could disable the trace button until one is added?
- When starting the app or clicking "Reset", there is text which says "click on the network lines or points to add a utility element". With Java samples we don't normally include API related terms (i.e. referring to utility element) on the screen nor starting instructions, instead the user refers the read me to understand how to use the sample from the off. We should therefore probably remove this intro text and reword our "How to use this sample section" slightly. In any case, I think it's misleading because it's not clear from the UI if the user needs to add a starting location or a barrier.

network_analysis/find-connected-features-in-utility-network/src/main/java/com/esri/samples/find_connected_features_in_utility_network/FindConnectedFeaturesInUtilityNetworkController.java
JonLavi: I've removed the workaround again, after discussion with @ldanzinger . We're going to keep this sample in Dev until U7, then the bug will be fixed in the public release, and the sample will be published.

network_analysis/find-connected-features-in-utility-network/src/main/java/com/esri/samples/find_connected_features_in_utility_network/FindConnectedFeaturesInUtilityNetworkController.java
JonLavi: Thanks @Rachael-E , I've confirmed with @ZackAllen that this happens in .NET as well. We'd need to investigate?

network_analysis/find-connected-features-in-utility-network/README.md
JonLavi: Updated this section with the wording used in the upcoming sample 'subnetwork trace'.

JonLavi:
Thanks for the thorough review @Rachael-E . 

> 
> 
>     * I think we can make the HBox a bit shorter. Currently it has quite a lot of empty space at the bottom.
> 

Done.

>     * When clicking "Trace" after setting a starting location and a barrier, a progress indicator comes up which is helpful. However it disappears before it has finished finding connected features. Can we fix this so the progress indicator remains on until the trace is complete? Also the "Trace" and "Reset" buttons should remain disabled until the trace is complete.
> 

I've managed to fix this. It just needed to wait for 'selectFeaturesAsync' to finish.

>     * Rather than clicking "Trace" with no starting locations or barriers set and getting an error coming up that no starting locations were added, we could disable the trace button until one is added?
> 

I think this would be a really nice improvement. I think would require to do  `button.disableProperty().bind()` to the array of starting locations, but it would need to be an `ObservableList` for this to work. I've had a quick go but couldn't get a clean solution for this - perhaps @tschie can suggest something? Otherwise I find it starts adding a lot of unnecessary UI code which distracts from an already complex API. 

>     * When starting the app or clicking "Reset", there is text which says "click on the network lines or points to add a utility element". With Java samples we don't normally include API related terms (i.e. referring to utility element) on the screen nor starting instructions, instead the user refers the read me to understand how to use the sample from the off. We should therefore probably remove this intro text and reword our "How to use this sample section" slightly. In any case, I think it's misleading because it's not clear from the UI if the user needs to add a starting location or a barrier.

I've removed this promt and made some changes to the 'how it works' section. Let me know what you think.

network_analysis/find-connected-features-in-utility-network/src/main/java/com/esri/samples/find_connected_features_in_utility_network/FindConnectedFeaturesInUtilityNetworkController.java
Rachael-E: @colinanderson what do you think about this? Should we be raising an issue to investigate this? I'm unclear if it's a Java or a CAPI bug.

network_analysis/find-connected-features-in-utility-network/README.md
```diff
@@ -2,46 +2,47 @@
...
+![](ConnectedTrace.png)
 
 ## Use case
 
 This is useful to visualize and validate the network topology of a utility network for quality assurance. 
 
 ## How to use the sample
 
-To add a starting point, select 'Add starting location(s)' and tap on one or more features. To add a barrier, select 'Add barrier(s)' and tap on one or more features. Depending on the type of feature, you may be prompted to select a terminal or the distance from the tapped location will be computed. Click 'Trace' to highlight all features connected to the specified starting locations and not positioned beyond the barriers. Click 'Reset' to clear parameters and start over.
+Add starting locations and (optionally) barriers by clicking on one or more features while 'Add starting locations' or 'Add barriers' is selected. When a junction feature is identified, you may be prompted to select a terminal. When an edge feature is identified, the distance from the tapped location to the beginning of the edge feature will be computed. Click 'Trace' to highlight all features connected to the specified starting locations and not positioned beyond the barriers. Click 'Reset' to clear parameters and start over.
```
Rachael-E: this reads really nicely now

network_analysis/find-connected-features-in-utility-network/README.md
```diff
@@ -2,46 +2,47 @@
...
 4. Add a `GraphicsOverlay` with symbology that distinguishes starting points from barriers.
-5. Identify features on the map and add a `Graphic` that represents its purpose (starting point or barrier) at the location of each identified feature.
-6. Determine the type of the identified feature using `UtilityNetwork.Definition.GetNetworkSource` passing its table name.
-7. If a junction, display a terminal picker when more than one terminal is found and create a `UtilityElement` with the selected terminal or the single terminal if there is only one.
-8. If an edge, create a `UtilityElement` from the identified feature and compute its `FractionAlongLine` using `GeometryEngine.FractionAlong`.
-9. Run a `UtilityNetwork.TraceAsync` with the specified parameters.
-10. Group the `UtilityElementTraceResult.Elements` by `NetworkSource.Name`
-11. For every `FeatureLayer` in this map with elements, select features by converting `UtilityElement`(s) to `ArcGISFeature`(s) using `UtilityNetwork.GetFeaturesFromElementsAsync`
+5. Add a listener for clicks on the map view, and use `mapView.identifyLayersAsync()` to dentify clicked features.
```
Rachael-E: typo here for what should be "identify"

network_analysis/find-connected-features-in-utility-network/README.md
```diff
@@ -2,46 +2,47 @@
...
+6. Add a `Graphic` that represents its purpose (starting point or barrier) at the location of each identified feature.
+7. Determine the type of the identified feature using `utilityNetwork.getDefinition().getNetworkSource()` passing its table name.
+8. If a junction, display a terminal picker when more than one `UtilityTerminal` is found and create a `UtilityElement` with the selected terminal or the single terminal if there is only one.
+9. If an edge, create a utility element from the identified feature and set its `FractionAlongEdge` using `GeometryEngine.fractionAlong()`.
+10. Create `UtilityTraceParameters` for the trace, specifying the trace type, `UtilityTraceType.CONNECTED`, adding the starting location, and adding barriers (if applicable).
+11. Run a `utilityNetwork.traceAsync()` with the specified parameters, and get the `UtilityTraceResult`.
+12. From the utility trace result, get the `UtilityElementTraceResult`, and group the utility elements within by their network source name, which you can retrieve using `utilityElement.getNetworkSource().getName()`.
```
Rachael-E: we're not using first person throughout the rest of the readme, so I think here we could say:

> which can be retrieved using `utilityElement.getNEtworkSource().getName()`

network_analysis/find-connected-features-in-utility-network/README.md
```diff
@@ -2,46 +2,47 @@
...
-* UtilityNetworks.UtilityTraceParameters
-* UtilityNetworks.UtilityTraceResult
-* UtilityNetworks.UtilityElementTraceResult
-* UtilityNetworks.UtilityNetworkDefinition
-* UtilityNetworks.UtilityNetworkSource
-* UtilityNetworks.UtilityAssetType
-* UtilityNetworks.UtilityTerminal
-* UtilityNetworks.UtilityElement
-* GeometryEngine.FractionAlong
+* UtilityNetwork
```
Rachael-E: these need to be placed in alphabetical order

network_analysis/find-connected-features-in-utility-network/README.metadata.json
```diff
@@ -11,11 +11,23 @@
         "network analysis"
     ],
     "relevant_apis": [
-        "UtilityNetworks",
-        "GeometryEngine"
+        "GeometryEngine",
```
Rachael-E: I think these should be in alphabetical order here too?

Rachael-E (approved):
@JonLavi thanks for making those changes, it's looking really good. just a few minor things to address then ready for second review.

JonLavi:
Thanks @Rachael-E , good catch on those details!

network_analysis/find-connected-features-in-utility-network/src/main/java/com/esri/samples/find_connected_features_in_utility_network/FindConnectedFeaturesInUtilityNetworkSample.java
```diff
@@ -0,0 +1,64 @@
...
+  public void start(Stage stage) throws Exception {
+
+    // set up the scene
+    FXMLLoader loader = new FXMLLoader(getClass().getResource("/find_connected_features_in_utility_network.fxml"));
+    Parent root = loader.load();
+    controller = loader.getController();
+    Scene scene = new Scene(root);
+
+    // set title, size and add scene to stage
+    stage.setTitle("Find Connected Features in A Utility Network Sample");
```
tschie: `in A` -> 'in a'

network_analysis/find-connected-features-in-utility-network/src/main/java/com/esri/samples/find_connected_features_in_utility_network/FindConnectedFeaturesInUtilityNetworkController.java
```diff
@@ -0,0 +1,478 @@
...
+  private SimpleMarkerSymbol startingPointSymbol;
+  private UtilityNetwork utilityNetwork;
+  private UtilityTraceParameters utilityTraceParameters;
+
+  public void initialize() {
+    try {
+      // create a basemap and set it to the map view
+      ArcGISMap map = new ArcGISMap(Basemap.createStreetsNightVector());
+      mapView.setMap(map);
+      mapView.setViewpointAsync(new Viewpoint(
```
tschie: add a comment for why this viewpoint is chosen

network_analysis/find-connected-features-in-utility-network/src/main/java/com/esri/samples/find_connected_features_in_utility_network/FindConnectedFeaturesInUtilityNetworkController.java
```diff
@@ -0,0 +1,478 @@
...
+        new Envelope(-9813547.35557238, 5129980.36635111, -9813185.0602376, 5130215.41254146,
+          SpatialReferences.getWebMercator())));
+
+      // create symbols for the starting point and barriers: a green cross for starting locations, and a red X for barriers
```
tschie: ~: a green cross for starting locations, and a red X for barriers~

The code already says this

network_analysis/find-connected-features-in-utility-network/src/main/java/com/esri/samples/find_connected_features_in_utility_network/FindConnectedFeaturesInUtilityNetworkController.java
```diff
@@ -0,0 +1,478 @@
...
+
+      // create and apply a renderer for the electric distribution lines feature layer
+      distributionLineLayer.setRenderer(new SimpleRenderer(
+        new SimpleLineSymbol(SimpleLineSymbol.Style.SOLID, ColorUtil.colorToArgb(Color.DARKCYAN), 3)));
+
+      // add the feature layers to the map
+      map.getOperationalLayers().addAll(Arrays.asList(distributionLineLayer, electricDeviceLayer));
+
+      // create a graphics overlay and add it to the map view
+      graphicsOverlay = new GraphicsOverlay();
```
tschie: Looks like you could just have two graphics overlays -- one for starting locations, one for barriers with renders set to use their respective symbols. Then you don't have to keep the symbols as fields.

network_analysis/find-connected-features-in-utility-network/src/main/java/com/esri/samples/find_connected_features_in_utility_network/FindConnectedFeaturesInUtilityNetworkController.java
```diff
@@ -0,0 +1,478 @@
...
+
+    // disable the UI
+    traceButton.setDisable(true);
+    resetButton.setDisable(true);
+
+    // create utility trace parameters for a connected trace
+    utilityTraceParameters = new UtilityTraceParameters(UtilityTraceType.CONNECTED, startingLocations);
+
+    // if any barriers have been created, add them to the parameters
+    if (!barriers.isEmpty()) {
```
tschie: This check appears to be unnecessary.

network_analysis/find-connected-features-in-utility-network/src/main/java/com/esri/samples/find_connected_features_in_utility_network/FindConnectedFeaturesInUtilityNetworkController.java
```diff
@@ -0,0 +1,478 @@
...
+    // show the progress indicator and update the status text
+    progressIndicator.setVisible(true);
+    statusLabel.setText("Finding connected features...");
+
+    // disable the UI
+    traceButton.setDisable(true);
+    resetButton.setDisable(true);
+
+    // create utility trace parameters for a connected trace
+    utilityTraceParameters = new UtilityTraceParameters(UtilityTraceType.CONNECTED, startingLocations);
```
tschie: I think you should create `UtilityTraceParameters` at the start (and after a reset) and then add starting locations and barriers on subsequent clicks. Then you don't have to keep track of two redundant array lists. It's an unfortunate design choice that you can't create parameters with empty starting locations and then add them later, so you'll have to create it the first time a starting location is added.

network_analysis/find-connected-features-in-utility-network/src/main/java/com/esri/samples/find_connected_features_in_utility_network/FindConnectedFeaturesInUtilityNetworkController.java
```diff
@@ -0,0 +1,478 @@
...
+    utilityTraceResultsFuture.addDoneListener(() -> {
+      try {
+        List<UtilityTraceResult> utilityTraceResults = utilityTraceResultsFuture.get();
+
+          if (utilityTraceResults.get(0) instanceof UtilityElementTraceResult) {
+            UtilityElementTraceResult utilityElementTraceResult = (UtilityElementTraceResult) utilityTraceResults.get(0);
+
+            if (!utilityElementTraceResult.getElements().isEmpty()) {
+              // clear the previous selection from the layer
+              mapView.getMap().getOperationalLayers().forEach(layer -> {
```
tschie: Seems like this should be cleared right before you run the trace so that even if the trace fails, the previous result is cleared.

network_analysis/find-connected-features-in-utility-network/src/main/java/com/esri/samples/find_connected_features_in_utility_network/UtilityTerminalSelectionDialog.java
```diff
@@ -0,0 +1,76 @@
...
+    } catch (Exception e) {
+      e.printStackTrace();
+    }
+
+    // on 'continue', return the selected terminal
+    setResultConverter(dialogButton -> {
+      if (dialogButton == continueButton) {
+        if (utilityTerminalListView.getSelectionModel().getSelectedItem() != null) {
+          try {
+            return utilityTerminalListView.getSelectionModel().getSelectedItem();
```
tschie: why would this throw?

network_analysis/find-connected-features-in-utility-network/src/main/java/com/esri/samples/find_connected_features_in_utility_network/UtilityTerminalListCell.java
```diff
@@ -0,0 +1,40 @@
...
+package com.esri.samples.find_connected_features_in_utility_network;
+
+import javafx.scene.control.ListCell;
+
+import com.esri.arcgisruntime.utilitynetworks.UtilityTerminal;
+
+/**
+ * Shows the title of the UtilityTerminal in the selection list view.
+ */
+class UtilityTerminalListCell extends ListCell<UtilityTerminal> {
```
tschie: Since all this class does is show the name, you might consider using a `StringConverter` instead of a custom `ListCell`.

network_analysis/find-connected-features-in-utility-network/src/main/resources/find_connected_features_in_utility_network.fxml
```diff
@@ -0,0 +1,69 @@
...
+                            <Color opacity="0.75"/>
+                        </fill>
+                    </BackgroundFill>
+                </fills>
+            </Background>
+        </background>
+
+        <RadioButton fx:id="startingLocationsRadioButton" text="Add Starting Location(s)" selected="true">
+            <toggleGroup>
+                <ToggleGroup fx:id="toggleGroup"/>
```
tschie:
```xml
<fx:define>
  <ToggleGroup fx:id="toggleGroup"/>
</fx:define>
<RadioButton toggleGroup="$toggleGroup" fx:id="startingLocationsRadioButton" text="Add Starting Location(s)" selected="true"/>
<RadioButton toggleGroup="$toggleGroup" text="Add Barrier(s)"/>
```

network_analysis/find-connected-features-in-utility-network/src/main/java/com/esri/samples/find_connected_features_in_utility_network/UtilityTerminalSelectionDialog.java
```diff
@@ -0,0 +1,76 @@
...
+import javafx.fxml.FXML;
+import javafx.fxml.FXMLLoader;
+import javafx.scene.control.Alert;
+import javafx.scene.control.ButtonType;
+import javafx.scene.control.Dialog;
+import javafx.scene.control.ListView;
+
+import com.esri.arcgisruntime.utilitynetworks.UtilityTerminal;
+
+class UtilityTerminalSelectionDialog extends Dialog<UtilityTerminal> {
```
tschie: Since I think you can only choose one, you could use the built in `ChoiceDialog` instead of a custom dialog.

tschie (changes requested):
Good start. I added some suggestions. Please request a re-review after you've applied them.

network_analysis/find-connected-features-in-utility-network/src/main/java/com/esri/samples/find_connected_features_in_utility_network/FindConnectedFeaturesInUtilityNetworkController.java
tschie: I've created an issue for this.

network_analysis/find-connected-features-in-utility-network/src/main/java/com/esri/samples/find_connected_features_in_utility_network/FindConnectedFeaturesInUtilityNetworkController.java
JonLavi: Yep, this is a great suggestion

network_analysis/find-connected-features-in-utility-network/src/main/java/com/esri/samples/find_connected_features_in_utility_network/UtilityTerminalSelectionDialog.java
JonLavi: Not sure, I think I've blindly followed a pattern from another similar interface that may have run an async task or something similar. I've remove the try/catch

network_analysis/find-connected-features-in-utility-network/src/main/resources/find_connected_features_in_utility_network.fxml
JonLavi: Thanks, I've not seen <fx:define> before!

JonLavi:
Hi both,
This sample has been combined with the 'subnetwork trace' sample, so the design has changed slightly. I've gone through and updated this, and applied most of @tschie 's suggestions.

Main changes:
- Add a combobox or selection prompt for choosing the trace type.
- Add a unique value renderer for the line layer (to differentiate the types)
- update the colors and sizes of the start and barrier symbols
- There is a safety check on the geometry engine result to account for an edge case. This is related to a bug that core is now aware of, but may not be fixed before 100.7 release.
- Terminal selection is unchanged
- Trace code now builds utility trace parameters using the selected type of trace.

Could you both please re-review? Thanks!

tschie (changes requested):
- After adding a few starting locations and barriers, the only trace type that doesn't error is CONNECTED.

A few small things:
- I see "fraction along edge: ..." with no number (may be cut off)
- Terminal chooser can show up on other screen (use `alert.initOwner`)

network_analysis/trace-a-utility-network/src/main/java/com/esri/samples/trace_a_utility_network/TraceAUtilityNetworkController.java
```diff
@@ -0,0 +1,516 @@
...
+          enableButtonInteraction();
+
+          // hide the progress indicator
+          progressIndicator.setVisible(false);
+
+          // update the status text
+          statusLabel.setText("");
+
+          // listen to clicks on the map view
+          mapView.setOnMouseClicked(this::handleMapViewClicked);
```
tschie: Define in FXML. If you only want it enabled when the utility network is loaded, add a loaded check in the handler.

network_analysis/trace-a-utility-network/src/main/java/com/esri/samples/trace_a_utility_network/TraceAUtilityNetworkController.java
```diff
@@ -0,0 +1,516 @@
...
+   */
+  @FXML
+  private void handleMapViewClicked(MouseEvent e) {
+    if (e.getButton() == MouseButton.PRIMARY && e.isStillSincePress()) {
+
+      // show the progress indicator
+      progressIndicator.setVisible(true);
+
+      // set whether the user is adding a starting point or a barrier
+      boolean isAddingStartingPoint = startingLocationsRadioButton.isSelected();
```
tschie: I think this could be inlined in the `if` since it's only used once

network_analysis/trace-a-utility-network/src/main/java/com/esri/samples/trace_a_utility_network/TraceAUtilityNetworkController.java
```diff
@@ -0,0 +1,516 @@
...
+                UtilityNetworkSource networkSource = utilityNetwork.getDefinition().getNetworkSource(featureTableName);
+
+                UtilityElement utilityElement = null;
+
+                // check if the network source is a junction or an edge
+                if (networkSource.getSourceType() == UtilityNetworkSource.Type.JUNCTION) {
+                  //  create a utility element with the identified feature
+                  utilityElement = createUtilityElement(identifiedFeature, networkSource);
+                }
+                // check if the network source is an edge
```
tschie: This comment is already implied by the comment above and the condition. The `else` should be on the same line as the preceding `{` too.

network_analysis/trace-a-utility-network/src/main/java/com/esri/samples/trace_a_utility_network/TraceAUtilityNetworkController.java
```diff
@@ -0,0 +1,516 @@
...
+                  //  create a utility element with the identified feature
+                  utilityElement = utilityNetwork.createElement(identifiedFeature, null);
+
+                  // get the geometry of the identified feature as a polyline, and remove the z component
+                  Polyline polyline = (Polyline) GeometryEngine.removeZ(identifiedFeature.getGeometry());
+
+                  // compute how far the clicked location is along the edge feature
+                  double fractionAlongEdge = GeometryEngine.fractionAlong(polyline, mapPoint, -1);
+                  if (Double.isNaN(fractionAlongEdge)) {
+                    return;
```
tschie, line 272: This seems extreme for not having a comment or alert explaining what happened. Is it desired that the utility element doesn't get added in this case? If so, an alert should be shown.

network_analysis/trace-a-utility-network/src/main/java/com/esri/samples/trace_a_utility_network/TraceAUtilityNetworkController.java
```diff
@@ -0,0 +1,516 @@
...
+                UtilityElement utilityElement = null;
+
+                // check if the network source is a junction or an edge
+                if (networkSource.getSourceType() == UtilityNetworkSource.Type.JUNCTION) {
+                  //  create a utility element with the identified feature
+                  utilityElement = createUtilityElement(identifiedFeature, networkSource);
+                }
+                // check if the network source is an edge
+                else if (networkSource.getSourceType() == UtilityNetworkSource.Type.EDGE &&
+                  identifiedFeature.getGeometry().getGeometryType() == GeometryType.POLYLINE) {
```
tschie: Will this ever not be `POLYLINE` with type `EDGE`?

network_analysis/trace-a-utility-network/src/main/java/com/esri/samples/trace_a_utility_network/TraceAUtilityNetworkController.java
```diff
@@ -0,0 +1,516 @@
...
+                  }
+
+                  // set the fraction along edge
+                  utilityElement.setFractionAlongEdge(fractionAlongEdge);
+
+                  // update the status label text
+                  statusLabel.setText("Fraction along edge: " + utilityElement.getFractionAlongEdge());
+                }
+
+                if (utilityElement != null) {
```
tschie: Can this ever occur in this sample? Is there any way to avoid it?

network_analysis/trace-a-utility-network/src/main/java/com/esri/samples/trace_a_utility_network/TraceAUtilityNetworkController.java
```diff
@@ -0,0 +1,516 @@
...
+    // find the code matching the asset group name in the feature's attributes
+    int assetGroupCode = (int) attributes.get(assetGroupFieldName.toLowerCase());
+
+    // iterate through the network source's asset groups to find the group with the matching code
+    List<UtilityAssetGroup> assetGroups = networkSource.getAssetGroups();
+    for (UtilityAssetGroup assetGroup : assetGroups) {
+      if (assetGroup.getCode() == assetGroupCode) {
+
+        // get the code for the feature's asset type from its attributes
+        String assetTypeCode = attributes.get("assettype").toString();
```
tschie: This attribute appears to be type `short`.  You could do:
```java
if (assetType.getCode() == (short) attributes.get("assettype"))
```

network_analysis/trace-a-utility-network/src/main/java/com/esri/samples/trace_a_utility_network/TraceAUtilityNetworkController.java
```diff
@@ -0,0 +1,516 @@
...
+            // if there is only one terminal, use it to create a utility element
+            if (terminals.size() == 1) {
+              utilityElement = utilityNetwork.createElement(identifiedFeature, terminals.get(0));
+              // show the name of the terminal in the status label
+              showTerminalNameInStatusLabel(terminals.get(0));
+
+              // if there is more than one terminal, prompt the user to select one
+            } else if (terminals.size() > 1) {
+              // create a dialog for terminal selection
+              UtilityTerminalSelectionDialog utilityTerminalSelectionDialog =
```
tschie: Instead of creating a custom dialog, you could do:

```java
ChoiceDialog<UtilityTerminal> utilityTerminalSelectionDialog = new ChoiceDialog<>(terminals.get(0), terminals);
utilityTerminalSelectionDialog.setTitle("Select Utility Terminal:");
              
// override the list cell in the dialog's combo box to show the terminal name
@SuppressWarnings("unchecked")
ComboBox<UtilityTerminal> comboBox = (ComboBox<UtilityTerminal>) ((GridPane) utilityTerminalSelectionDialog.getDialogPane().getContent()).getChildren().get(1);
comboBox.setCellFactory(param -> new UtilityTerminalListCell());
comboBox.setButtonCell(new UtilityTerminalListCell());
```

I admit it's a bit of a hack to work around `UtilityTerminal.toString` not being helpful, but I think it's a little better than creating a whole dialog for it.

Here's the `ListCell`:
```java
public class UtilityTerminalListCell extends ListCell<UtilityTerminal> {
    @Override
    protected void updateItem(UtilityTerminal item, boolean empty) {
        super.updateItem(item, empty);
        if (empty) {
            setText(null);
        } else {
            setText(item.getName());
        }
    }
}
```

network_analysis/trace-a-utility-network/src/main/java/com/esri/samples/trace_a_utility_network/TraceAUtilityNetworkController.java
```diff
@@ -0,0 +1,516 @@
...
+            progressIndicator.setVisible(false);
+            new Alert(Alert.AlertType.ERROR, "Trace result not a utility element.").show();
+          }
+
+      } catch (Exception e) {
+        statusLabel.setText("Trace failed.");
+        progressIndicator.setVisible(false);
+        new Alert(Alert.AlertType.ERROR, "Error running utility network connected trace.").show();
+      } finally {
+        enableButtonInteraction();
```
tschie, line 439: This is enabling the trace button before the trace is completed.

tschie (changes requested):
Found a few more things. See code comments.

network_analysis/trace-a-utility-network/src/main/java/com/esri/samples/trace_a_utility_network/TraceAUtilityNetworkController.java
JonLavi: This is currently a workaround for `fractionAlong` returning `NaN`. That issue should be fixed for 100.7, I believe. Let me double check and we might be able to remove this anyway.

network_analysis/trace-a-utility-network/src/main/java/com/esri/samples/trace_a_utility_network/TraceAUtilityNetworkController.java
JonLavi: Probably another guard to do with the above case
`
Will this ever not be POLYLINE with type EDGE?
`
Will have to look it up and double check the sample design.

network_analysis/trace-a-utility-network/src/main/java/com/esri/samples/trace_a_utility_network/TraceAUtilityNetworkController.java
JonLavi: I suppose it is one less class so it keeps the focus on the main point of the sample, so happy to implement this!

network_analysis/trace-a-utility-network/src/main/java/com/esri/samples/trace_a_utility_network/TraceAUtilityNetworkController.java
JonLavi: I don't actuall know. The UtilityNetwork team have created the sample design so I am guessing there is some thought behind it, but I'll double check!

JonLavi:
> 
> 
>     * After adding a few starting locations and barriers, the only trace type that doesn't error is CONNECTED.
> 
> 
> A few small things:
> 
>     * I see "fraction along edge: ..." with no number (may be cut off)
> 
>     * Terminal chooser can show up on other screen (use `alert.initOwner`)

Thanks @tschie , I've adressed most things for now.
I've not been getting any errors running any type of trace. The sample can be a bit iffy sometimes if the input isn't exaclty correct. Can you provide a screenshot of the starting points/barriers that cause errors, and let me know which error is thrown? Would need some more debugging info to get to the bottom of this.

I believe the fraction along edge bug has to do with the workaround against it showing NaN, so there's a bit more investigating to do there.

Rachael-E (changes requested):
@JonLavi this is another one of those tricky samples isn't it! After test driving the sample I've noticed a few other behaviours that I wanted to run by you before going back in for a full review. 

When finding a junction in the sample, I click on one and up comes an option to change to High or Low.  Two questions/thoughts:

1. If I click either high or low, depending on what I have selected in the UI box (a starting location or a barrier), that starter or barrier will be placed directly in the junction's location after I've confirmed a selection from High or Low. I think this is a bug in the sample, a mouse click handler issue?
2. After I click on the junction's box, and select Low, and then go back to the box, it's set back to High according to the drop down selection. As a user I'm unclear if my selection to Low has been executed. Is this feature integral to showing how to trace a utility network?

network_analysis/trace-a-utility-network/src/main/java/com/esri/samples/trace_a_utility_network/TraceAUtilityNetworkController.java
tschie: It's fine to have a catch for it. I'm just saying that a `return` without telling the user why nothing happened sounds extreme.

JonLavi:
Thanks @Rachael-E. I'm not sure if I'm understanding the behaviours you are describing correctly? Might be best to discuss either in person or record a video.
With some of the recent comments from @tschie as well, I'm thinking it might be good to get the UtilityNetwork team involved and perhaps re-visit this sample design, as I'm unsure about what technical reasons are behind some of the design decisions. Any thoughts?

network_analysis/find-connected-features-in-utility-network/src/main/java/com/esri/samples/find_connected_features_in_utility_network/FindConnectedFeaturesInUtilityNetworkController.java
JonLavi: While this is being fixed in core I've added an error message to inform the user that the barrier/starting point cant be created at that particular map point.

network_analysis/trace-a-utility-network/src/main/java/com/esri/samples/trace_a_utility_network/TraceAUtilityNetworkController.java
JonLavi: Gotcha, I've added an error message for now. This should be fixed in core at some point, so we'd be able to remove it afterwards.

tschie:
> After I click on the junction's box, and select Low, and then go back to the box, it's set back to High according to the drop down selection. As a user I'm unclear if my selection to Low has been executed.

Each time you make a selection, you're adding some new utility element, right? Can you add a barrier to Low & High? What would you expect then?

network_analysis/trace-a-utility-network/src/main/java/com/esri/samples/trace_a_utility_network/TraceAUtilityNetworkController.java
JonLavi: Removed the check for `GeometryType.POLYLINE`. As you said, the geometry type will always match

JonLavi:
> 
> 
> > After I click on the junction's box, and select Low, and then go back to the box, it's set back to High according to the drop down selection. As a user I'm unclear if my selection to Low has been executed.
> 
> Each time you make a selection, you're adding some new utility element, right? Can you add a barrier to Low & High? What would you expect then?

Yes, there's a UtilityElement being addeed every time. You could add a barrier both on the Low & High 'terminals' of a node on the network, but visually they sit in the same location, so the two crosses would overlap each other. On a code level, there will be two barriers added to the trace. 

Meanwhile @Rachael-E and I clarified the previous questions in person. There was some confusion with the previous UI and the input confirmation, but should be improved now.

Would you both mind having another look? Thanks!

network_analysis/trace-a-utility-network/README.md
```diff
@@ -0,0 +1,54 @@
...
+3. Create and load a `UtilityNetwork` with the same feature service URL and map.
+4. Add a `GraphicsOverlay` with symbology that distinguishes starting points from barriers.
+5. Add a listener for clicks on the map view, and use `mapView.identifyLayersAsync()` to identify clicked features.
+6. Create a `UtilityElement` that represents its purpose (starting point or barrier) at the location of each identified feature, and save it to a list.
+7. Determine the type of the identified feature using `utilityNetwork.getDefinition().getNetworkSource()` passing its table name.
+8. If a junction, display a terminal picker when more than one `UtilityTerminal` is found and create a `UtilityElement` with the selected terminal or the single terminal if there is only one.
+9. If an edge, create a utility element from the identified feature and set its `FractionAlongEdge` using `GeometryEngine.fractionAlong()`.
+10. Create `TraceParameters` with the selected trace type along with the collected starting locations and barriers (if applicable).
```
Rachael-E: should be `UtilityTraceParameters` here, and then camel case in references below

network_analysis/trace-a-utility-network/README.md
```diff
@@ -0,0 +1,54 @@
...
+11. Set the `TraceParameters.TraceConfiguration` with the utility tier's `TraceConfiguration` property.
+12. Run a `utilityNetwork.traceAsync()` with the specified parameters, and get the `UtilityTraceResult`.
+13. From the utility trace result, get the `UtilityElementTraceResult`, and group the utility elements within by their network source name, which can be retrieved using `utilityElement.getNetworkSource().getName()`.
+13. For every feature layer in this map with elements, select features by converting utility elements to `ArcGISFeature`(s) using `UtilityNetwork.fetchFeaturesForElementsAsync()`
```
Rachael-E: I can't follow this in the code, not seeing a `fetchFeaturesForElementsAsync()` method?

network_analysis/trace-a-utility-network/build.gradle
```diff
@@ -0,0 +1,77 @@
+plugins {
+    id 'application'
+    id 'org.openjfx.javafxplugin' version '0.0.5'
+}
+
+group = 'com.esri.samples'
+
+ext {
```
Rachael-E, line 8: this can now be updated to 100.7.0

network_analysis/trace-a-utility-network/src/main/java/com/esri/samples/trace_a_utility_network/TraceAUtilityNetworkController.java
```diff
@@ -0,0 +1,493 @@
...
+import javafx.scene.paint.Color;
+
+import com.esri.arcgisruntime.concurrent.ListenableFuture;
+import com.esri.arcgisruntime.data.ArcGISFeature;
+import com.esri.arcgisruntime.data.FeatureQueryResult;
+import com.esri.arcgisruntime.data.QueryParameters;
+import com.esri.arcgisruntime.data.ServiceFeatureTable;
+import com.esri.arcgisruntime.geometry.Envelope;
+import com.esri.arcgisruntime.geometry.GeometryEngine;
+import com.esri.arcgisruntime.geometry.GeometryType;
```
Rachael-E, line 47: unused import

network_analysis/trace-a-utility-network/src/main/java/com/esri/samples/trace_a_utility_network/TraceAUtilityNetworkController.java
```diff
@@ -0,0 +1,493 @@
...
+          progressIndicator.setVisible(false);
+          new Alert(Alert.AlertType.ERROR, "Trace result not a utility element.").show();
+          enableButtonInteraction();
+        }
+
+      } catch (Exception e) {
+        statusLabel.setText("Trace failed.");
+        progressIndicator.setVisible(false);
+        // Note: this sample server may return a generic message in some circumstances when incompatible trace
+        // parameters are specified. This will be fixed in ArcGIS server 10.8
```
Rachael-E: is this a certain it will be fixed? I think it's probably ok not to state a time for when it will be fixed (the concern being it won't be fixed with 10.8 for whatever reason, or we forget to update the comment in the sample). Just stating that it may return a generic message is good enough for me.

Rachael-E (commented):
just a query on the readme to look at and some minor comments. Otherwise looks good @JonLavi !

network_analysis/trace-a-utility-network/README.md
JonLavi: Yes, I've tidied this up a little

network_analysis/trace-a-utility-network/README.md
JonLavi: This was lost in translation from .NET, and not up to date anyway. We've used a different approach in Java so that the UI is disabled until the selection is done. Updated to reflect what's going on in the code.

network_analysis/trace-a-utility-network/src/main/java/com/esri/samples/trace_a_utility_network/TraceAUtilityNetworkController.java
JonLavi: 🎉

network_analysis/trace-a-utility-network/src/main/java/com/esri/samples/trace_a_utility_network/TraceAUtilityNetworkController.java
JonLavi: Good point, I've removed the reference to any upcoming fix

network_analysis/trace-a-utility-network/build.gradle
JonLavi: 🎉

network_analysis/trace-a-utility-network/README.md
```diff
@@ -0,0 +1,54 @@
...
+7. Determine the type of the identified feature using `utilityNetwork.getDefinition().getNetworkSource()` passing its table name.
+8. If a junction, display a terminal picker when more than one `UtilityTerminal` is found and create a `UtilityElement` with the selected terminal or the single terminal if there is only one.
+9. If an edge, create a utility element from the identified feature and set its `FractionAlongEdge` using `GeometryEngine.fractionAlong()`.
+10. Create `UtilityTraceParameters` with the selected trace type along with the collected starting locations and barriers (if applicable).
+11. Set the `TraceConfiguration` of the utility trace parameters to the the utility tier's trace configuration property.
+12. Run a `utilityNetwork.traceAsync()` with the specified parameters, and get the `UtilityTraceResult`.
+13. From the utility trace result, get the `UtilityElementTraceResult`.
+14. For each feature layer in the map, create `QueryParameters` to find features from the result who's network source name matches the layer's feature table name, and use `FeatureLayer.selectFeaturesAsync()` to select these features.
```
tschie: ~who's~ whose

network_analysis/trace-a-utility-network/src/main/java/com/esri/samples/trace_a_utility_network/TraceAUtilityNetworkController.java
```diff
@@ -0,0 +1,493 @@
...
+        if (utilityTraceResults.get(0) instanceof UtilityElementTraceResult) {
+          UtilityElementTraceResult utilityElementTraceResult = (UtilityElementTraceResult) utilityTraceResults.get(0);
+
+          if (!utilityElementTraceResult.getElements().isEmpty()) {
+
+            // iterate through the map's feature layers
+            mapView.getMap().getOperationalLayers().forEach(layer -> {
+              if (layer instanceof FeatureLayer) {
+
+                // create query parameters to find features who's network source name matches the layer's feature
```
tschie: ~who's~ whose

network_analysis/trace-a-utility-network/src/main/java/com/esri/samples/trace_a_utility_network/TraceAUtilityNetworkController.java
```diff
@@ -0,0 +1,493 @@
...
+                  if (Double.isNaN(fractionAlongEdge)) {
+                    new Alert(Alert.AlertType.ERROR, "Cannot add starting location / barrier here.");
+                    return;
+                  }
+
+                  // set the fraction along edge
+                  utilityElement.setFractionAlongEdge(fractionAlongEdge);
+
+                  // update the status label text
+                  statusLabel.setText("Fraction along edge: " + utilityElement.getFractionAlongEdge());
```
tschie: Can we format the fraction so it stops at 3 or so decimal places?

tschie (approved):
Looks okay. Just a couple minor things

Rachael-E (approved):

Rachael-E (approved):
@JonLavi great! Please also merge with master once this has been checked into dev too.

network_analysis/trace-a-utility-network/src/main/java/com/esri/samples/trace_a_utility_network/TraceAUtilityNetworkController.java
JonLavi: Done!

JonLavi:
Thanks both, going to merge into `dev` first

PR #425, opened 2019-12-03
100.7.0

tschie:
New and updated samples for the 100.7.0 release.


markjdugger (approved):
Looks good to me. Couple of things:  (1) I can't verify whether the code changes in the samples are accurate or not, but most of them concern addition of  `System.getProperty("data.dir")` as a param to `File` constructor and they're all done the same way, so I hope that's good.
(2) I verified that `arcgisVersion` in Gradle build scripts was properly changed to 100.7.0

Rachael-E (approved):
@tschie this is looking good to me!

JonLavi (approved):
Looks good. There's a rouge gradle-wrapper.jar file in there:
`map/download-preplanned-map/gradle/wrapper/gradle-wrapper.jar`
not sure if it's necessary or not.

PR #426, opened 2019-12-05
remove reference to toggling coordinates

JonLavi:
Hi both,

small change needed to this readme that @hardikc07 made me aware of.
The original design for this sample has a toggle box to swap the x-y-coordinates of the data. I vaguely remember the consensus being that this distracted from the purpose of the sample, and that functionality must have been removed (at least in the Java implementation). As a result, just need to remove the reference to this toggle box from the readme.

Let me know if you spot anything out of the order.

Thanks

### DIFF
--- a/ogc/browse-wfs-layers/README.md
+++ b/ogc/browse-wfs-layers/README.md
@@ -12,8 +12,6 @@ Services often have multiple layers available for display. For example, a featur
 
 A list of layers in the WFS service will be shown. Select a layer to display it.
 
-Some WFS services return coordinates in X,Y order, while others return coordinates in lat/long (Y,X) order. If you don't see features rendered or you see features in the wrong location, use the checkbox to swap the coordinate order and reload. Note: the WFS service in this sample does not require coordinate swapping.
-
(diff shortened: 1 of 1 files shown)


Rachael-E (approved):
thanks @JonLavi , I remember this. As I recall when building this sample, we agreed as a team to remove the co-ordinate switching. But then a few months later decided it should remain. I didn't get round to adding that control back in again, I think there might be an issue open for this in the samples repo. For now though yes we should remove that text, can always add it back in again when the functionality is put back in. Might want to check with @tschie if he'd like this change to go into the 100.7.0 branch as well.

tschie (approved):
Yes, please pull into 100.7.0 as well.

hardikc07:
looks like all other SDKs have that checkbox option here : https://devtopia.esri.com/runtime/common-samples/issues/1246

and chat thread on slack too - https://esri-runtime.slack.com/archives/CCD6VTY0L/p1559920096025100?thread_ts=1559844598.011700&cid=CCD6VTY0L

PR #427, opened 2019-12-06
Display subtype feature layer

Rachael-E:
This is a new sample for display subtype feature layer. The design shows a random renderer being applied to the sublayer, this would have added in a lot of unnecessary code so I've opted for a toggle button to switch between the original renderer and one single new one. @JonLavi could you take a first look please?


tschie:
Don't think we really need to be able to change the minimum scale in the sample to demonstrate it.

Either way, I'd propose a minor UI change:
UI

feature_layers/display-subtype-feature-layer/README.md
```diff
@@ -0,0 +1,36 @@
...
+## Use case
+
+This is useful for controlling labeling, visibility and symbology of a given subtype as though they are distinct layers on the map.
+
+## How to use the sample
+
+The sample loads with the sublayer visible on the map. Toggle its visibility by clicking the "Show sublayer" checkbox. To change the sublayer's renderer, click the "Toggle Renderer" button, and click the "Set Minimum Scale" button to set its minimum scale to that of the current map scale. Zoom in and out to see the sublayer become visible based on its new scale range.
+
+## How it works
+1.  Create a `SubtypeFeatureLayer` from a `ServiceFeatureTable` that defines a subtype, and add it to the `ArcGISMap`.
```
JonLavi: extra little sneaky space here after the bullet point

feature_layers/display-subtype-feature-layer/README.md
```diff
@@ -0,0 +1,36 @@
...
+2. Get a `SubtypeSublayer` from the subtype feature layer using its name.
+3. Enable the sublayer's labels and define them with `getLabelDefinitions()`.
```
JonLavi, line 21: Should we be using `.` before method calls? i.e. `.getLabelDefinitions()`? Would apply to a few of them below as well.

feature_layers/display-subtype-feature-layer/src/main/java/com/esri/samples/feature_layers/DisplaySubtypeFeatureLayerController.java
```diff
@@ -0,0 +1,143 @@
...
+import com.esri.arcgisruntime.layers.SubtypeFeatureLayer;
+import com.esri.arcgisruntime.layers.SubtypeSublayer;
+import com.esri.arcgisruntime.mapping.ArcGISMap;
+import com.esri.arcgisruntime.mapping.Basemap;
+import com.esri.arcgisruntime.mapping.Viewpoint;
+import com.esri.arcgisruntime.symbology.Renderer;
+import com.esri.arcgisruntime.symbology.SimpleMarkerSymbol;
+import com.esri.arcgisruntime.symbology.SimpleRenderer;
+import com.esri.arcgisruntime.symbology.Symbol;
+import javafx.fxml.FXML;
```
JonLavi: This import would belong with the group below? Also 2 lines below is a `com.esri.arcgisruntime` import that would belong in this group up here

feature_layers/display-subtype-feature-layer/src/main/java/com/esri/samples/feature_layers/DisplaySubtypeFeatureLayerController.java
```diff
@@ -0,0 +1,143 @@
...
+import javafx.scene.control.Label;
+import javafx.scene.control.ToggleButton;
+
+public class DisplaySubtypeFeatureLayerController {
+
+  @FXML private ToggleButton toggleRendererButton;
+  @FXML private MapView mapView;
+  @FXML private Label currentMapScaleLabel;
+  @FXML private Label minScaleLabel;
+  @FXML private CheckBox checkBox;
```
JonLavi: I'd prefer giving this a more meaningful name for the function it serves, e.g. `showSublayerCheckBox`

feature_layers/display-subtype-feature-layer/src/main/java/com/esri/samples/feature_layers/DisplaySubtypeFeatureLayerController.java
```diff
@@ -0,0 +1,143 @@
...
+      mapView.addMapScaleChangedListener(mapScaleChangedEvent ->
+        currentMapScaleLabel.setText("Current Map Scale: 1:" + Math.round(mapView.getMapScale())));
+
+      // set the viewpoint to Naperville, Illinois
+      Viewpoint initialViewpoint = new Viewpoint(new Envelope(-9812691.11079696, 5128687.20710657,
+        -9812377.9447607, 5128865.36767282, SpatialReferences.getWebMercator()));
+      map.setInitialViewpoint(initialViewpoint);
+
+      // create a subtype feature layer from the service feature table, and add it to the map
+      final String serviceFeatureTableUrl = "https://sampleserver7.arcgisonline" +
```
JonLavi: I suppose the String is only ever being used once, so you could pop it straight into the `new ServiceFeatureTable()` below?

feature_layers/display-subtype-feature-layer/src/main/java/com/esri/samples/feature_layers/DisplaySubtypeFeatureLayerController.java
```diff
@@ -0,0 +1,143 @@
...
+      // access the json required for sublayer label definitions
+      File jsonFile = new File(System.getProperty("data.dir"),
+          Objects.requireNonNull(getClass().getClassLoader().getResource("label_definition.json")).getFile());
+      final String json;
+      // read in the complete file as a string
+      try (Scanner scanner = new Scanner(jsonFile)) {
+        json = scanner.useDelimiter("\\A").next();
+      }
+
+      subtypeFeatureLayer.loadAsync();
```
JonLavi, line 80: just needs a quick comment here along the lines of 
`// load the subtype feature layer`

feature_layers/display-subtype-feature-layer/src/main/java/com/esri/samples/feature_layers/DisplaySubtypeFeatureLayerController.java
```diff
@@ -0,0 +1,143 @@
...
+        sublayer.setLabelsEnabled(true);
+        sublayer.getLabelDefinitions().add(LabelDefinition.fromJson(json));
+
+        // get the original renderer of the sublayer
+        originalRenderer = sublayer.getRenderer();
+      });
+
+    } catch (Exception e) {
+      // on any error, display the stack trace.
+      e.printStackTrace();
```
JonLavi, line 99: I think we've moved to not just printing the stack trace, but showing an alert with a more meaningful error message. Something like:
`new Alert(Alert.AlertType.ERROR, "Error loading the Subtype Feature Layer").show`

feature_layers/display-subtype-feature-layer/src/main/java/com/esri/samples/feature_layers/DisplaySubtypeFeatureLayerController.java
```diff
@@ -0,0 +1,143 @@
...
+  }
+  
+  /**
+   * Switches between a renderer with a pink diamond symbol and the sublayer's original renderer.
+   */
+  @FXML
+  private void handleToggleRendererButtonClicked() {
+    if (toggleRendererButton.isSelected()) {
+      Symbol symbol = new SimpleMarkerSymbol(SimpleMarkerSymbol.Style.DIAMOND, 0xfff58f84, 20);
+      Renderer sublayerRenderer = new SimpleRenderer(symbol);
```
JonLavi: To be a bit more resource efficient and avoid creating a new `Renderer` every time this is called, you could create the Symbol and Renderer in the `initialize()` and then simply apply it in this method.

feature_layers/display-subtype-feature-layer/src/main/resources/style.css
```diff
@@ -0,0 +1,27 @@
...
+
+.slider .axis {
+    -fx-tick-label-fill: white;
+}
+
+.range-slider .axis {
+    -fx-tick-label-fill: white;
+}
+
+.panel-region .check-box {
```
JonLavi, line 5: I think you're only using the `.panel-region *` rules in this style.css, so would be worth getting rid of the rest. I think the idea was to keep the style.css files limited to only the used rules going forward, and some day go back and tidy up the old ones.

JonLavi (changes requested):
Nice work @Rachael-E, only a few small suggestions for improvements

feature_layers/display-subtype-feature-layer/README.md
Rachael-E: good question. I looked at some of our other readmes, and they don't use the `.` before the method. As I recall, we don't use it since it should be obvious to the reader which object is calling that method (e.g. in step 3 above it should be obvious that it's the SubtypeSublayer that the getLabelDefinitions() would be called on). If we're using the full object with method reference then that should be referred to as `SubtypeSublayer.getLabelDefinitions()`.

feature_layers/display-subtype-feature-layer/src/main/java/com/esri/samples/feature_layers/DisplaySubtypeFeatureLayerController.java
Rachael-E: in this case, the try and catch are covering the whole of the initialize method. If anything goes wrong it could be for a number of reasons, not only related to the loading of the subtype feature layer. So I think in this case it's ok to leave as `e.printStackTrace()` since we can't be certain it would only ever be a loading error.

Rachael-E:
thanks @tschie, @JonLavi, I've addressed the comments and updated the UI. Could you take another look please? Thanks!

feature_layers/display-subtype-feature-layer/README.md
```diff
@@ -10,10 +10,10 @@ This is useful for controlling labeling, visibility and symbology of a given sub
 
 ## How to use the sample
 
-The sample loads with the sublayer visible on the map. Toggle its visibility by clicking the "Show sublayer" checkbox. To change the sublayer's renderer, click the "Toggle Renderer" button, and click the "Set Minimum Scale" button to set its minimum scale to that of the current map scale. Zoom in and out to see the sublayer become visible based on its new scale range.
+The sample loads with the sublayer visible on the map. Toggle its visibility by clicking the "Show sublayer" checkbox. To change the sublayer's renderer, click the "Show alternative renderer" button, and to return to its original renderer, click the "Show alternative renderer" button. Click the "Set sublayer minimum scale" button to set the sublayer's minimum scale to that of the current map scale (this is labelled in the UI and updates when the scale is changed). Zoom in and out to see the sublayer become visible based on its new scale range.
```
JonLavi: I think the second reference to the button here is wrong: ` To change the sublayer's renderer, click the "Show alternative renderer" button, and to return to its original renderer, click the "Show original renderer" button.`

JonLavi (approved):
Nice work @Rachael-E , only one small mistake in the updated 'how to use' section.

feature_layers/display-subtype-feature-layer/src/main/java/com/esri/samples/feature_layers/DisplaySubtypeFeatureLayerController.java
```diff
@@ -0,0 +1,148 @@
...
+      map.setInitialViewpoint(initialViewpoint);
+
+      // create a subtype feature layer from the service feature table, and add it to the map
+      ServiceFeatureTable serviceFeatureTable = new ServiceFeatureTable("https://sampleserver7.arcgisonline" +
+        ".com/arcgis/rest/services/UtilityNetwork/NapervilleElectric/FeatureServer/100");
+      SubtypeFeatureLayer subtypeFeatureLayer = new SubtypeFeatureLayer(serviceFeatureTable);
+      map.getOperationalLayers().add(subtypeFeatureLayer);
+      
+      // access the json required for sublayer label definitions
+      File jsonFile = new File(System.getProperty("data.dir"),
```
tschie: Since we use Apache's IOUtils in other samples, you could do:
```groovy
compile 'commons-io:commons-io:2.4'
```
```java
final String json = IOUtils.toString(getClass().getResourceAsStream("/label_definition.json"), StandardCharsets.UTF_8);
```

feature_layers/display-subtype-feature-layer/src/main/java/com/esri/samples/feature_layers/DisplaySubtypeFeatureLayerController.java
```diff
@@ -0,0 +1,148 @@
...
+  private void handleMinScaleButtonClicked() {
+    sublayer.setMinScale(mapView.getMapScale());
+    minScaleLabel.setText("Sublayer min scale: 1:" + Math.round(sublayer.getMinScale()));
+  }
+
+  /**
+   * Sets the visibility of the sublayer.
+   */
+  @FXML
+  private void controlSublayerVisibility() {
```
tschie: The naming convention should probably be similar to the other handlers.

feature_layers/display-subtype-feature-layer/src/main/java/com/esri/samples/feature_layers/DisplaySubtypeFeatureLayerController.java
```diff
@@ -0,0 +1,148 @@
...
+      e.printStackTrace();
+    }
+  }
+
+  /**
+   * Sets the minimum scale of the labels for the sublayer.
+   */
+  @FXML
+  private void handleMinScaleButtonClicked() {
+    sublayer.setMinScale(mapView.getMapScale());
```
tschie, line 108: The sublayer might not be initialized before these handlers are called.

feature_layers/display-subtype-feature-layer/src/main/java/com/esri/samples/feature_layers/DisplaySubtypeFeatureLayerController.java
```diff
@@ -0,0 +1,148 @@
...
+  @FXML
+  private void controlSublayerVisibility() {
+    sublayer.setVisible(sublayerVisibilityCheckbox.isSelected());
+  }
+  
+  /**
+   * Sets the renderer of the sublayer to its original format (a white and black circular icon). 
+   */
+  @FXML
+  private void handleOriginalRendererButtonClicked() {
```
tschie, line 124: This is okay for two radio buttons. With more, you'd probably set `userData` on the radio button, and then watch the toggle group for changes for the selected user data.

tschie (commented):
Just a few more suggestions

feature_layers/display-subtype-feature-layer/src/main/java/com/esri/samples/feature_layers/DisplaySubtypeFeatureLayerController.java
Rachael-E: Good to know thanks! I was wondering if there was a better way to handle this.

feature_layers/display-subtype-feature-layer/src/main/java/com/esri/samples/feature_layers/DisplaySubtypeFeatureLayerController.java
Rachael-E: I've handled this by disabling the UI until the layer has loaded: otherwise I could put in a load check for the sublayer in each method?

Rachael-E:
@tschie @JonLavi thanks, all changes updated. Feel free to merge if you think this looks good, otherwise I'll do it when back in the office.

feature_layers/display-subtype-feature-layer/src/main/java/com/esri/samples/feature_layers/DisplaySubtypeFeatureLayerController.java
tschie: nvm, what you're doing is fine

PR #430, opened 2019-12-12
Patch/iwa self signed

JonLavi:
Following the issue pointed out by @gayl0000 and the advice given by @XuemingRocks, I've updated this sample to return the correct response when encountering `AuthenticationChallenge.Type.SELF_SIGNED_CHALLENGE`.

I've also noticed that it would be beneficial to check if the entered URL has the 'https://' prefix, and add it if it isn't in there already.


tschie (approved):

Rachael-E (approved):
looks good! Might want to check with @tschie whether this should first be merged to dev, and then merge dev into 100.7.0. Just in case we miss this being added into dev later, though maybe master then gets merged into dev anyway at the beginning of a new release, I'm not too sure.

JonLavi:
Good point @Rachael-E .
@tschie can you please advise on the merging order with the new release?

PR #437, opened 2020-01-13
Testfx find place

JonLavi:
Hi Tyler,
As discussed, here’s a draft PR for the testfx experiment.

### DIFF
--- a/search/find-place/src/test/java/com/esri/samples/find_place/FindPlaceTest.java
+++ b/search/find-place/src/test/java/com/esri/samples/find_place/FindPlaceTest.java
@@ -108,7 +108,45 @@ public class FindPlaceTest extends FxRobot {
     sleep(1000);
   }
 
+  /**
+   * Test Case 2: When using a custom search query, the app returns the expected number of results
+   */
+  @Test
+  public void customSearchInput() {
 
+    clickOn(placeBox).write("esri").sleep(1000);
+
+    Node esriEntry = lookup("Esri, 380 New York St, Redlands, CA, 92373, USA").query();
+    if (esriEntry != null) {
+      clickOn(esriEntry);
+    } else {
+      fail("No search result found for custom query \"esri\" ");
+    }
+
+    sleep(1000);
+
+    clickOn(locationBox).write("Redlands").sleep(1000);
+    
+    Node redlandsEntry = lookup("Redlands, CA, USA").query();
+    if (redlandsEntry != null) {
+      clickOn(redlandsEntry);
+    } else {
+      fail("No search result found for custom query \"Redlands\" ");
+    }
+    sleep(1000);
+    clickOn("Search");
+
+    // wait for the map view to zoom to the location
+    if (mapView instanceof MapView) {
+      ((MapView) mapView).addDrawStatusChangedListener((drawStatusChangedEvent) -> {
+        if (drawStatusChangedEvent.getDrawStatus() == DrawStatus.COMPLETED) {
+          // get the graphics overlay and assert that the expected number of graphics (location pins) is displayed
+          GraphicsOverlay graphicsOverlay = ((MapView) mapView).getGraphicsOverlays().get(0);
+          assertEquals("Unexpected number of graphics (location pins) found",4, graphicsOverlay.getGraphics().size());
+        }
+      });
+    }
+  }
 
   // test cases:
   // TODO: find callout node in map view and confirm text (probably something you'll be doing in our unit tests)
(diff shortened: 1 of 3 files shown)


search/find-place/src/test/java/com/esri/samples/find_place/FindPlaceTest.java
```diff
@@ -0,0 +1,206 @@
...
+    }
+
+    // get a handle on the ComboBoxes
+    placeBox = lookup("#placeBox").queryComboBox();
+    locationBox = lookup("#locationBox").queryComboBox();
+
+    // get a handle on the ComboBoxes' arrow buttons
+    for (Node child : locationBox.getChildrenUnmodifiable()) {
+      if (child.getStyleClass().contains("arrow-button")) {
+        locationBoxArrowRegion = ((Pane) child).getChildren().get(0);
```
tschie, line 63: not sure why this is getting set in a loop

search/find-place/src/test/java/com/esri/samples/find_place/FindPlaceTest.java
```diff
@@ -0,0 +1,206 @@
...
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+
+public class FindPlaceTest extends FxRobot {
+
+  public static final int SLEEP_500MS = 500;
```
tschie, line 32: why are these needed?

search/find-place/src/test/java/com/esri/samples/find_place/FindPlaceTest.java
```diff
@@ -0,0 +1,206 @@
...
+  public void panEnablesRedoSearchButton() {
+
+    // get a handle on the Redo Search button
+    Button redoButton = lookup("#redoButton").queryButton();
+
+    // perform a search with pre-defined search inputs
+    clickOn(placeBoxArrowRegion);
+    clickOn("Starbucks");
+
+    clickOn(locationBoxArrowRegion);
```
tschie: Chain methods that are related. Probably separate out the sleep calls for readability.

search/find-place/src/test/java/com/esri/samples/find_place/FindPlaceTest.java
```diff
@@ -0,0 +1,206 @@
...
+    FxToolkit.cleanupStages();
+    // wait before the next test
+    sleep(SLEEP_1000MS);
+  }
+
+  /**
+   * Test Case 1: When using one of the pre-defined search inputs, the app returns the expected number of results
+   */
+  @Test
+  public void defaultSearchInput() {
```
tschie, line 91: In automated tests, we never actually want to make real network requests. Instead, we should mock them.

Simply verifying the default options for the dropdowns are present is good enough for this kind of "end-to-end" test. To test that the `search` method on the controller is called, you'd want to use TestFX at the controller level instead of the application level. Then you could use something like Mockito to mock the controller.

search/find-place/src/test/java/com/esri/samples/find_place/FindPlaceTest.java
```diff
@@ -0,0 +1,206 @@
...
+
+    // assert that the expected number of graphics (location pins) is displayed
+    assertEquals("Unexpected number of graphics (location pins) found", 50, graphicsOverlay.getGraphics().size());
+  }
+
+  /**
+   * Test Case 2: When using a custom search query, the app returns the expected number of results
+   */
+  @Test
+  public void customSearchInput() {
```
tschie, line 112: I have noticed a bug in this sample where if you clear the place input (or edit it significantly), the default options are removed. Would be good to have a test for that and use it to fix the bug and verify.

tschie (commented):
Good stuff. The comments I made in your other PR are also mostly relevant here.

search/find-place/src/test/java/com/esri/samples/find_place/FindPlaceTest.java
JonLavi: Since there's two ComboBoxes, I needed to find a way to get a handle on each 'arrow-button' and know which is which. Since I think if I simply do `lookup(".arrow-button").queryAll()` I'll get two nodes and not know which ComboBox they belong to? Might be a more elegant way to do this though

search/find-place/src/test/java/com/esri/samples/find_place/FindPlaceTest.java
JonLavi: As you pointed out below, we're still doing live server requests here, so I had these sleep intervals put in to wait for the response. They are defined as constants just so they can be changed at one point and it'd apply everywhere

search/find-place/src/test/java/com/esri/samples/find_place/FindPlaceTest.java
JonLavi: Yep, I've noticed the tests sometimes tried to proceed before the server response came back. I'll look into mocking!

JonLavi:
closing as there will not likely be any progress on this in the forseable future

PR #439, opened 2020-01-30
New sample: Configure subnetwork trace

JonLavi:
Hi both,
I've finally come around to finishing this sample.

It's a fairly complex one, so I imagine there's still a lot of room for improvement. Take your time as it's also a big one! Hopefully there aren't too many inconsistencies, but apologies if I missed out anything regarding the readme/metadata/screenshot. 

If you two have any suggestions about refactoring/increasing readibility, I'll gladly apply them. So if you have any ideas for improving `convertToDataType()` and `expressionToString()` that would also be great.

Other than that, limiting the input in `comparisonValuesTextField` to numerical perhaps isn't the ideal way of doing things? I've used a listener and regex to basically 'disallow' entering text, and only accept numbers and decimal places. The alternative may have been declaring a class `NumericalField` that inherits from `TextField` and does that internally, but I felt that may be a bit too much for a sample?

Thanks both!


network_analysis/configure-subnetwork-trace/README.md
```diff
@@ -0,0 +1,69 @@
...
+* 'Transformer Load' Equal '15'
+* 'Phases Current' DoesNotIncludeTheValues 'A'
+* 'Generation KW' LessThan '50'
+
+## How it works
+
+1. Create and load a `UtilityNetwork` with a feature service URL, then get an asset type and a tier by their names.
+2. Populate the choice list for the comparison source with the non-system defined `utilityNetwork.getDefinition().getNetworkAttributes()`. Populate the choice list for the comparison operator with the enum values from `UtilityAttributeComparisonOperator`.
+3. Create a `UtilityElement` from this asset type to use as the starting location for the trace.
+4. Update the selected barrier expression and the checked options in the UI using this tier's `TraceConfiguration`.
```
Rachael-E: spotted this with the Android implementation: some of the class names need updated from the original design 
```suggestion
4. Update the selected barrier expression and the checked options in the UI using this tier's `UtilityTraceConfiguration`.
```

network_analysis/configure-subnetwork-trace/README.md
```diff
@@ -0,0 +1,69 @@
...
+5. When 'Network Attribute' is selected, if its `Domain` is a `CodedValueDomain`, populate the choice list for the comparison value with its `CodedValues`.  Otherwise, display a free-form textbox for entering an attribute value.
+6. When 'Add' is clicked, create a new `UtilityNetworkAttributeComparison` using the selected comparison source, operator, and selected or typed value. Use the selected source's `UtilityNetworkAttribute.DataType` to convert the comparison value to the correct data type.
+7. If the Traversability's list of `Barriers` is not empty, create a `UtilityTraceOrCondition` with the existing `Barriers` and the new comparison from step 7.
+8. When 'Trace' is clicked, create `UtilityTraceParameters` passing in `UtilityTraceType.SUBNETWORK` for the trace type and the default starting location. Set its `TraceConfiguration` with the modified options, selections, and expression; then run a `UtilityNetwork.traceAsync()`.
```
Rachael-E:
```suggestion
8. When 'Trace' is clicked, create `UtilityTraceParameters` passing in `UtilityTraceType.SUBNETWORK` for the trace type and the default starting location. Set its `UtilityTraceConfiguration` with the modified options, selections, and expression; then run a `UtilityNetwork.traceAsync()`.
```

network_analysis/configure-subnetwork-trace/README.metadata.json
```diff
@@ -0,0 +1,55 @@
+{
+  "category": "Network analysis",
```
Rachael-E: I think Utility Network now has it's own category after Jenn requested it. Probably not something for this PR, but just to make a note this is something we'll have to migrate in the future along with the other UN samples.

network_analysis/configure-subnetwork-trace/src/main/java/com/esri/samples/configure_subnetwork_trace/CodedValueListCell.java
```diff
@@ -0,0 +1,37 @@
+/*
+ * Copyright 2019 Esri.
```
Rachael-E:
```suggestion
 * Copyright 2020 Esri.
```

network_analysis/configure-subnetwork-trace/src/main/java/com/esri/samples/configure_subnetwork_trace/ComparisonSourceListCell.java
```diff
@@ -0,0 +1,36 @@
+/*
+ * Copyright 2019 Esri.
```
Rachael-E:
```suggestion
 * Copyright 2020 Esri.
```

network_analysis/configure-subnetwork-trace/src/main/java/com/esri/samples/configure_subnetwork_trace/ConfigureSubnetworkTraceLauncher.java
```diff
@@ -0,0 +1,27 @@
+/*
+ * Copyright 2019 Esri.
```
Rachael-E:
```suggestion
 * Copyright 2020 Esri.
```

network_analysis/configure-subnetwork-trace/src/main/java/com/esri/samples/configure_subnetwork_trace/ConfigureSubnetworkTraceSample.java
```diff
@@ -0,0 +1,63 @@
+/*
+ * Copyright 2019 Esri.
```
Rachael-E:
```suggestion
 * Copyright 2020 Esri.
```

network_analysis/configure-subnetwork-trace/src/main/resources/configure_subnetwork_trace.fxml
```diff
@@ -0,0 +1,70 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<!--
+  ~ Copyright 2019 Esri.
```
Rachael-E:
```suggestion
  ~ Copyright 2020 Esri.
```

network_analysis/configure-subnetwork-trace/src/main/java/com/esri/samples/configure_subnetwork_trace/ConfigureSubnetworkTraceController.java
```diff
@@ -0,0 +1,419 @@
+/*
+ * Copyright 2019 Esri.
```
Rachael-E:
```suggestion
 * Copyright 2020 Esri.
```

network_analysis/configure-subnetwork-trace/src/main/java/com/esri/samples/configure_subnetwork_trace/ConfigureSubnetworkTraceController.java
```diff
@@ -0,0 +1,419 @@
...
+
+    // check if an expression is already defined for the traversability barriers
+    if (utilityTraceConfiguration.getTraversability().getBarriers() instanceof UtilityTraceConditionalExpression) {
+      UtilityTraceConditionalExpression otherExpression =
+          (UtilityTraceConditionalExpression) utilityTraceConfiguration.getTraversability().getBarriers();
+      // use the existing expression to create an `or` expression with the user-defined expression
+      expression = new UtilityTraceOrCondition(otherExpression, expression);
+    }
+
+    // set the new expression to the traversability
```
Rachael-E: I'm a bit unsure of this comment: from the code it reads as if the expression is being set to the barriers?

network_analysis/configure-subnetwork-trace/src/main/java/com/esri/samples/configure_subnetwork_trace/ConfigureSubnetworkTraceController.java
```diff
@@ -0,0 +1,419 @@
...
+
+      // toggle the selection combo box to be visible if it has any items
+      comparisonValuesComboBox.setVisible(!comparisonValuesComboBox.getItems().isEmpty());
+      // toggle the text field to be hidden if the combo box is visible, or show it if the combo box is invisible
+      comparisonValuesTextField.setVisible(!comparisonValuesComboBox.isVisible());
+    }
+  }
+
+  /**
+   * Converts an object representing a value into the data type specified.
```
Rachael-E: it's as you've said before: this is a confusing workflow! I'm also unsure about the method description. Digging into it a little deeper, this method always returns a value. The javadoc makes it sound like a data type is returned. 

Also, are there ever any booleans passed on? I couldn't find any but I admit I am totally in the dark with this sample and how it's meant to be used.

I'd be tempted to get some more experienced eyes on this just in case there's something deeper going on that we're not handling well, @tschie or @colinanderson. Should we be handling this conversion at a Java or Core level, or is this a pattern we'd expect a customer to handle?

network_analysis/configure-subnetwork-trace/src/main/java/com/esri/samples/configure_subnetwork_trace/ConfigureSubnetworkTraceController.java
```diff
@@ -0,0 +1,419 @@
...
+    UtilityAttributeComparisonOperator selectedOperator =
+        comparisonOperatorsComboBox.getSelectionModel().getSelectedItem();
+
+    // check if a comparison value was specified, and capture it to use as the last parameter of the
+    // UtilityNetworkAttributeComparison
+    Object otherValue;
+    // if a comparison value is selected from the ComboBox, use it as the third parameter
+    if (selectedAttribute.getDomain() instanceof CodedValueDomain &&
+        comparisonValuesComboBox.getSelectionModel().getSelectedItem() != null) {
+      // convert the selected comparison value to the data type defined by the selected attribute
```
Rachael-E: again I'm confused by this comment. we're taking the data type, and converting it from it's type to a value, no? as the description stands, it looks like we're just retrieving if it's an integer/float etc which I think is misleading, whereas we're returning it's actual value e.g. 4?

network_analysis/configure-subnetwork-trace/src/main/java/com/esri/samples/configure_subnetwork_trace/ConfigureSubnetworkTraceController.java
```diff
@@ -0,0 +1,419 @@
...
+  private UtilityTraceConfiguration initialUtilityTraceConfiguration;
+  private UtilityTraceConfiguration utilityTraceConfiguration;
+
+  @FXML
+  public void initialize() {
+
+    try {
+
+      // add a listener to the comparison value text field, so that it only accepts numerical input separated by a decimal
+      comparisonValuesTextField.textProperty().addListener((observable, oldValue, newValue) -> {
```
Rachael-E: this is cool! though I noticed when selecting e.g. "Is subnetwork controller EQUAL" and add 15.0 or 15000000000 that two exceptions are reported

`java.lang.NumberFormatException: For input string: "15000000000"`
`Caused by: java.lang.reflect.InvocationTargetException`

again I'm probably not using the values a utility networker would use, and in other fields these entries seem to be ok.

network_analysis/configure-subnetwork-trace/src/main/java/com/esri/samples/configure_subnetwork_trace/ConfigureSubnetworkTraceController.java
```diff
@@ -0,0 +1,419 @@
...
+
+          // save the initial expression
+          initialExpression =
+              (UtilityTraceConditionalExpression) utilityTier.getTraceConfiguration().getTraversability().getBarriers();
+
+          // show the initial expression in the text area
+          traceConditionsTextArea.setText(expressionToString(initialExpression));
+
+          // set the traversability scope
+          utilityTier.getTraceConfiguration().getTraversability().setScope(UtilityTraversabilityScope.JUNCTIONS);
```
Rachael-E: I think you can replace this with the following since you stored it in a variable earlier
```suggestion
          utilityTraceConfiguration().getTraversability().setScope(UtilityTraversabilityScope.JUNCTIONS);
```

network_analysis/configure-subnetwork-trace/src/main/java/com/esri/samples/configure_subnetwork_trace/ConfigureSubnetworkTraceController.java
```diff
@@ -0,0 +1,419 @@
...
+  private void onAddConditionClick() {
+
+    // get the selected utility network attribute and attribute comparison operator
+    UtilityNetworkAttribute selectedAttribute = comparisonSourcesComboBox.getSelectionModel().getSelectedItem();
+    UtilityAttributeComparisonOperator selectedOperator =
+        comparisonOperatorsComboBox.getSelectionModel().getSelectedItem();
+
+    // check if a comparison value was specified, and capture it to use as the last parameter of the
+    // UtilityNetworkAttributeComparison
+    Object otherValue;
```
Rachael-E: I think a comment to explain this Object reference would be helpful

network_analysis/configure-subnetwork-trace/src/main/java/com/esri/samples/configure_subnetwork_trace/ConfigureSubnetworkTraceController.java
```diff
@@ -0,0 +1,419 @@
...
+      comparisonValuesComboBox.setVisible(!comparisonValuesComboBox.getItems().isEmpty());
+      // toggle the text field to be hidden if the combo box is visible, or show it if the combo box is invisible
+      comparisonValuesTextField.setVisible(!comparisonValuesComboBox.isVisible());
+    }
+  }
+
+  /**
+   * Converts an object representing a value into the data type specified.
+   *
+   * @param value    the value to convert
```
Rachael-E: excess space here between param and description

network_analysis/configure-subnetwork-trace/src/main/java/com/esri/samples/configure_subnetwork_trace/ConfigureSubnetworkTraceController.java
```diff
@@ -0,0 +1,419 @@
...
+      stringBuilder.append(
+          String.format("'%1$s' %2$s", categoryComparison.getCategory().getName(),
+              categoryComparison.getComparisonOperator().name()));
+    }
+
+    // for network attribute comparison expressions, add the network attribute name and comparison operator
+    else if (expression instanceof UtilityNetworkAttributeComparison) {
+      UtilityNetworkAttributeComparison attributeComparison = (UtilityNetworkAttributeComparison) expression;
+      stringBuilder.append(
+          String.format("'%1$s' %2$s", attributeComparison.getNetworkAttribute().getName(),
```
Rachael-E: to avoid repeated use of `attributeComparison.getNetworkAttribute` you could instead store it in a variable for re-use: up to you though, I'm ok either way. 

`UtilityNetworkAttribute utilityNetworkAttribute = attributeComparison.getNetworkAttribute();`

network_analysis/configure-subnetwork-trace/src/main/java/com/esri/samples/configure_subnetwork_trace/CodedValueListCell.java
```diff
@@ -0,0 +1,37 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.esri.samples.configure_subnetwork_trace;
+
+import javafx.scene.control.ListCell;
+
+import com.esri.arcgisruntime.data.CodedValue;
+import com.esri.arcgisruntime.utilitynetworks.UtilityNetworkAttribute;
```
Rachael-E: unused import

network_analysis/configure-subnetwork-trace/src/main/java/com/esri/samples/configure_subnetwork_trace/CodedValueListCell.java
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * Shows the name of the UtilityNetworkAttribute in the attribute selection ComboBox.
```
Rachael-E: Is this description correct? it looks like the type the ListCall takes is `CodedValue` and not UtilityNetworkAttribute.

network_analysis/configure-subnetwork-trace/README.md
```diff
@@ -0,0 +1,69 @@
...
+
+Get a server-defined trace configuration for a given tier and modify its traversability scope, add new condition barriers and control what is included in the subnetwork trace result.
+
+![Configure Subnetwork Trace Sample Screenshot](ConfigureSubnetworkTrace.png)
+
+## Use case
+
+While some traces are built from an ad-hoc group of parameters, many are based on a variation of the trace configuration taken from the subnetwork definition. For example, an electrical trace will be based on the trace configuration of the subnetwork, but may add additional clauses to constrain the trace along a single phase. Similarly, a trace in a gas or electric design application may include features with a status of "In Design" that are normally excluded from trace results.
+
+## How to use the sample
```
Rachael-E: I think there should be a note in here that you have to use the "Add" button in order to add it and then run using the trace button.

Rachael-E (changes requested):
this is quite the sample, good work @JonLavi ! I've made some suggestions and asked some questions throughout. Also I reckon you probably started working on this in 2019 but we should update to 2020 for the copyrights.

network_analysis/configure-subnetwork-trace/README.md
```diff
@@ -0,0 +1,69 @@
...
+* 'Generation KW' LessThan '50'
+
+## How it works
+
+1. Create and load a `UtilityNetwork` with a feature service URL, then get an asset type and a tier by their names.
+2. Populate the choice list for the comparison source with the non-system defined `utilityNetwork.getDefinition().getNetworkAttributes()`. Populate the choice list for the comparison operator with the enum values from `UtilityAttributeComparisonOperator`.
+3. Create a `UtilityElement` from this asset type to use as the starting location for the trace.
+4. Update the selected barrier expression and the checked options in the UI using this tier's `TraceConfiguration`.
+5. When 'Network Attribute' is selected, if its `Domain` is a `CodedValueDomain`, populate the choice list for the comparison value with its `CodedValues`.  Otherwise, display a free-form textbox for entering an attribute value.
+6. When 'Add' is clicked, create a new `UtilityNetworkAttributeComparison` using the selected comparison source, operator, and selected or typed value. Use the selected source's `UtilityNetworkAttribute.DataType` to convert the comparison value to the correct data type.
```
Rachael-E: just spotted, mixing UI with "How it Works"! Which I don't think we're meant to do, but if it must stay then we should make it obvious it's UI. Our How it works section shouldn't really have UI in them so it's not as obvious as it would be in say the "How to use section". 
```suggestion
6. When the "Add" button is clicked from the UI, create a new `UtilityNetworkAttributeComparison` using the selected comparison source, operator, and selected or typed value. Use the selected source's `UtilityNetworkAttribute.DataType` to convert the comparison value to the correct data type.
```

Rachael-E (changes requested):

network_analysis/configure-subnetwork-trace/README.metadata.json
JonLavi: Yes, I'll confirm if this change is going live just yet.

network_analysis/configure-subnetwork-trace/src/main/java/com/esri/samples/configure_subnetwork_trace/CodedValueListCell.java
JonLavi: Good find, thanks

network_analysis/configure-subnetwork-trace/src/main/java/com/esri/samples/configure_subnetwork_trace/ConfigureSubnetworkTraceController.java
JonLavi: Yep, that will look a bit cleaner. Thanks!

network_analysis/configure-subnetwork-trace/src/main/java/com/esri/samples/configure_subnetwork_trace/CodedValueListCell.java
JonLavi: Good spot, it was wrong! Updated to be `CodedValue`

network_analysis/configure-subnetwork-trace/README.metadata.json
JonLavi: Made a PR for a new category, and once that's merged, this sample here will also go in there 👍

network_analysis/configure-subnetwork-trace/src/main/java/com/esri/samples/configure_subnetwork_trace/ConfigureSubnetworkTraceController.java
JonLavi: Perhaps 
```
// set the new expression to the traversability's barriers
```
will be a bit more specific?

network_analysis/configure-subnetwork-trace/src/main/java/com/esri/samples/configure_subnetwork_trace/ConfigureSubnetworkTraceController.java
JonLavi: I agree that this is quite confusing.
I've found that this particular utility network accepts `0` and `1` as a 'boolean', hence the text field only accepts numbers. 
I think this sample has a little more functionality than the dataset actually supports. I think the UN team designed it particularly this way so that users have a more general starting point rather than limiting it to the abilities of the demo dataset.
Maybe @jNery could also help clarify?

network_analysis/configure-subnetwork-trace/src/main/java/com/esri/samples/configure_subnetwork_trace/ConfigureSubnetworkTraceController.java
JonLavi: I think this is the particular case is where an int input of either 0 or 1 is used to represent 'false' or 'true', so the values can only be one of those two. We might have to rig the sample so that when selecting the comparison source `is network controller`, we'd get a drop-down with true/false.

JonLavi: Adding a try/catch to deal with the exception is a possible fix. However it still doesn't resolve the fact that the expected input for 'is subnetwork controller' is either 0 or 1

network_analysis/configure-subnetwork-trace/src/main/java/com/esri/samples/configure_subnetwork_trace/ConfigureSubnetworkTraceController.java
JonLavi: The input box for the `comparison value` at the moment only accepts 'string' inputs (limited to numbers via the code from the beginning of the initialize method:
```
comparisonValuesTextField.textProperty().addListener((observable, oldValue, newValue) -> {if (!newValue.matches("\\d*([.]\\d*)?")) {
          comparisonValuesTextField.setText(oldValue);
        }
```

When it comes to constructing the `UtilityTraceConditionalExpression`, the `comparison value` has to be a particular data type, depending on which `comparison attribute` is selected. So basically what we are doing is checking 'does this need to be a double/int/float/bool?' and convert the input to the appropriate type using `convertToDataType()`. That method returns an `Object` that represents the value, and it's type is not clear at the begining, hence it's type is just a generic `Object`.
Hopefully this explanation helps clarify a little? I'll scratch my head a bit on how to improve the comments so it's more obvious what's going on.

network_analysis/configure-subnetwork-trace/README.md
Rachael-E: this still needs updated to UtilityTraceConfiguration :)

utility_network/configure-subnetwork-trace/README.md
```diff
@@ -0,0 +1,69 @@
...
+
+![Configure Subnetwork Trace Sample Screenshot](ConfigureSubnetworkTrace.png)
+
+## Use case
+
+While some traces are built from an ad-hoc group of parameters, many are based on a variation of the trace configuration taken from the subnetwork definition. For example, an electrical trace will be based on the trace configuration of the subnetwork, but may add additional clauses to constrain the trace along a single phase. Similarly, a trace in a gas or electric design application may include features with a status of "In Design" that are normally excluded from trace results.
+
+## How to use the sample
+
+The sample loads with a server-defined trace configuration from a tier. Check or uncheck which options to include in the trace - such as containers or barriers. Use the selection boxes to define a new condition network attribute comparison. Click 'Trace' to run a subnetwork trace with this modified configuration from a default starting location.
```
Rachael-E: could still do with a reference to the "Add" button in this section

utility_network/configure-subnetwork-trace/README.md
```diff
@@ -0,0 +1,69 @@
...
+* 'Phases Current' DoesNotIncludeTheValues 'A'
+* 'Generation KW' LessThan '50'
+
+## How it works
+
+1. Create and load a `UtilityNetwork` with a feature service URL, then get an asset type and a tier by their names.
+2. Populate the choice list for the comparison source with the non-system defined `utilityNetwork.getDefinition().getNetworkAttributes()`. Populate the choice list for the comparison operator with the enum values from `UtilityAttributeComparisonOperator`.
+3. Create a `UtilityElement` from this asset type to use as the starting location for the trace.
+4. Update the selected barrier expression and the checked options in the UI using this tier's `UtilityTraceConfiguration`.
+5. When 'Network Attribute' is selected, if its `Domain` is a `CodedValueDomain`, populate the choice list for the comparison value with its `CodedValues`.  Otherwise, display a free-form textbox for entering an attribute value.
```
Rachael-E:
```suggestion
5. When 'Network Attribute' is selected, if its `Domain` is a `CodedValueDomain`, populate the choice list for the comparison value with its `CodedValue`s.  Otherwise, display a free-form textbox for entering an attribute value.
```

utility_network/configure-subnetwork-trace/README.md
```diff
@@ -0,0 +1,69 @@
...
+6. When the "Add" button is clicked from the UI, create a new `UtilityNetworkAttributeComparison` using the selected comparison source, operator, and selected or typed value. Use the selected source's `UtilityNetworkAttribute.DataType` to convert the comparison value to the correct data type.
+7. If the Traversability's list of `Barriers` is not empty, create a `UtilityTraceOrCondition` with the existing `Barriers` and the new comparison from step 7.
```
Rachael-E:
```suggestion
7. If the traversability's list of `Barriers` is not empty, create a `UtilityTraceOrCondition` with the existing `Barriers` and the new comparison from step 7.
```

utility_network/configure-subnetwork-trace/README.md
```diff
@@ -0,0 +1,69 @@
...
+8. When 'Trace' is clicked, create `UtilityTraceParameters` passing in `UtilityTraceType.SUBNETWORK` for the trace type and the default starting location. Set its `UtilityTraceConfiguration` with the modified options, selections, and expression; then run a `UtilityNetwork.traceAsync()`.
+9. When `Reset` is clicked, set the trace configurations expression back to its original value.
```
Rachael-E:
```suggestion
9. When the "Reset" button is clicked, set the trace configurations expression back to its original value.
```

utility_network/configure-subnetwork-trace/README.md
```diff
@@ -0,0 +1,69 @@
...
+3. Create a `UtilityElement` from this asset type to use as the starting location for the trace.
+4. Update the selected barrier expression and the checked options in the UI using this tier's `UtilityTraceConfiguration`.
+5. When 'Network Attribute' is selected, if its `Domain` is a `CodedValueDomain`, populate the choice list for the comparison value with its `CodedValues`.  Otherwise, display a free-form textbox for entering an attribute value.
+6. When the "Add" button is clicked from the UI, create a new `UtilityNetworkAttributeComparison` using the selected comparison source, operator, and selected or typed value. Use the selected source's `UtilityNetworkAttribute.DataType` to convert the comparison value to the correct data type.
+7. If the Traversability's list of `Barriers` is not empty, create a `UtilityTraceOrCondition` with the existing `Barriers` and the new comparison from step 7.
+8. When 'Trace' is clicked, create `UtilityTraceParameters` passing in `UtilityTraceType.SUBNETWORK` for the trace type and the default starting location. Set its `UtilityTraceConfiguration` with the modified options, selections, and expression; then run a `UtilityNetwork.traceAsync()`.
```
Rachael-E: Didn't spot these button references before in here: still not 100% happy with UI references in this section but I think with UN being a complex workflow it does help to have it split up. Again we just need to ensure the button names are tagged within quotation marks rather than as code. 
```suggestion
8. When the "Trace" button is clicked, create `UtilityTraceParameters` passing in `UtilityTraceType.SUBNETWORK` for the trace type and the default starting location. Set its `UtilityTraceConfiguration` with the modified options, selections, and expression; then run a `UtilityNetwork.traceAsync()`.
```

Rachael-E (commented):
spotted some additional minor readme things: will hold off re-reviewing the code until you've had the change to push up the comments we were commenting on ;)

utility_network/configure-subnetwork-trace/README.md
JonLavi: Thanks for spotting these. I think the reference to specific UI elements in this section makes it a little bit easier to keep an overview, since this is quite a complex workflow. I agree though that it's deviating from the standard!

utility_network/configure-subnetwork-trace/README.md
```diff
@@ -0,0 +1,69 @@
...
+
+![Image of configure subnetwork trace](ConfigureSubnetworkTrace.png)
+
+## Use case
+
+While some traces are built from an ad-hoc group of parameters, many are based on a variation of the trace configuration taken from the subnetwork definition. For example, an electrical trace will be based on the trace configuration of the subnetwork, but may add additional clauses to constrain the trace along a single phase. Similarly, a trace in a gas or electric design application may include features with a status of "In Design" that are normally excluded from trace results.
+
+## How to use the sample
+
+The sample loads with a server-defined trace configuration from a tier. Check or uncheck which options to include in the trace - such as containers or barriers. Use the selection boxes to define a new condition network attribute comparison, and then use 'Add' to add the it to the trace configuration. Click 'Trace' to run a subnetwork trace with this modified configuration from a default starting location.
```
Rachael-E: rogue "the" hanging around!
```suggestion
The sample loads with a server-defined trace configuration from a tier. Check or uncheck which options to include in the trace - such as containers or barriers. Use the selection boxes to define a new condition network attribute comparison, and then use 'Add' to add it to the trace configuration. Click 'Trace' to run a subnetwork trace with this modified configuration from a default starting location.
```

utility_network/configure-subnetwork-trace/README.md
JonLavi: Yep, this reference was added.

Rachael-E (approved):
just one wee typo to catch then it's good for 2nd review, thanks @JonLavi !

JonLavi:
Thanks @Rachael-E !
@tschie ready for your review when you have some time, thanks!

tschie:
I think a table view would really work well for this.

Configure-Subnetwork-Trace

You could have the first two columns use cells with combo boxes. Might be able to make a custom cell for the third columns which picks an appropriate picker for the value type instead of trying to interpret all of them as strings.

JonLavi:
That does look very sleek @tschie ! 
I don't think I have the scope to do a revamp of the design just now. Perhaps create an enhancement issue in the common repo?

JonLavi:
@tschie can I double check with you how you'd like to proceed here? Are you happy to review the sample as is, or should we put in an issue for a design enhancement? Note that most platforms have already implemented this as-is, so it would be more straightforward to merge this sample the way it is, for now, and keep an enhancement in mind for later. Cheers!

JonLavi:
@mbcoder / @colinanderson , would either of you be able to review this at some point?

utility_network/configure-subnetwork-trace/README.md
```diff
@@ -0,0 +1,69 @@
...
+Example barrier conditions for the default dataset:
+
+* 'Transformer Load' Equal '15'
+* 'Phases Current' DoesNotIncludeTheValues 'A'
+* 'Generation KW' LessThan '50'
+
+## How it works
+
+1. Create and load a `UtilityNetwork` with a feature service URL, then get an asset type and a tier by their names.
+2. Populate the choice list for the comparison source with the non-system defined `utilityNetwork.getDefinition().getNetworkAttributes()`. Populate the choice list for the comparison operator with the enum values from `UtilityAttributeComparisonOperator`.
```
alan-edi: Swap steps 1 and 2, to match their order in the code and also to make the reference to "this asset type" in step 3 make sense.

utility_network/configure-subnetwork-trace/README.md
```diff
@@ -0,0 +1,69 @@
...
+3. Create a `UtilityElement` from this asset type to use as the starting location for the trace.
+4. Update the selected barrier expression and the checked options in the UI using this tier's `UtilityTraceConfiguration`.
+5. When 'Network Attribute' is selected, if its `Domain` is a `CodedValueDomain`, populate the choice list for the comparison value with its `CodedValue`s. Otherwise, display a free-form textbox for entering an attribute value.
```
alan-edi: There seems to be an unhelpful mixture of terminology being used here. This doc talks of 'Network Attribute' being selected, whereas the code uses the term "comparison sources", as in `comparisonSourcesComboBox` and `onComparisonSourceChanged`.

Perhaps a simple way of improving that might be to say:

> When a 'Network Attribute' comparison source is selected...

utility_network/configure-subnetwork-trace/README.md
```diff
@@ -0,0 +1,69 @@
...
+6. When the 'Add' button is clicked from the UI, create a new `UtilityNetworkAttributeComparison` using the selected comparison source, operator, and selected or typed value. Use the selected source's `UtilityNetworkAttribute.DataType` to convert the comparison value to the correct data type.
+7. If the Traversability's list of `Barriers` is not empty, create a `UtilityTraceOrCondition` with the existing `Barriers` and the new comparison from step 7.
+8. When the 'Trace' button is clicked, create `UtilityTraceParameters` passing in `UtilityTraceType.SUBNETWORK` for the trace type and the default starting location. Set its `UtilityTraceConfiguration` with the modified options, selections, and expression; then run a `UtilityNetwork.traceAsync()`.
+9. When the 'Reset' button is clicked, set the trace configurations expression back to its original value.
+10. Display the count elements returned within the `UtilityElementTraceResult`.
```
alan-edi: I'd swap steps 9 and 10, because 10 relates to the trace that's run in step 8.

Also: the count elements -> the count **of** elements

utility_network/configure-subnetwork-trace/build.gradle
```diff
@@ -0,0 +1,74 @@
+plugins {
+    id 'application'
+    id 'org.openjfx.javafxplugin' version '0.0.5'
+}
+
+group = 'com.esri.samples'
+
+ext {
+    arcgisVersion = '100.8.0'
```
alan-edi: 100.9.0

utility_network/configure-subnetwork-trace/src/main/java/com/esri/samples/configure_subnetwork_trace/ConfigureSubnetworkTraceController.java
```diff
@@ -0,0 +1,424 @@
...
+import java.util.Collections;
+import java.util.List;
+import java.util.UUID;
+import java.util.stream.Collectors;
+
+import com.esri.arcgisruntime.concurrent.ListenableFuture;
+import com.esri.arcgisruntime.data.CodedValue;
+import com.esri.arcgisruntime.data.CodedValueDomain;
+import com.esri.arcgisruntime.loadable.LoadStatus;
+import com.esri.arcgisruntime.mapping.view.MapView;
```
alan-edi: Unused import

utility_network/configure-subnetwork-trace/src/main/java/com/esri/samples/configure_subnetwork_trace/ConfigureSubnetworkTraceController.java
```diff
@@ -0,0 +1,424 @@
...
+      UtilityNetworkAttributeComparison attributeComparison = (UtilityNetworkAttributeComparison) expression;
+
+      UtilityNetworkAttribute utilityNetworkAttribute = attributeComparison.getNetworkAttribute();
+
+      stringBuilder.append(
+          String.format("'%1$s' %2$s", utilityNetworkAttribute.getName(),
+              attributeComparison.getComparisonOperator().name()));
+
+      if (utilityNetworkAttribute.getDomain() instanceof CodedValueDomain) {
+        CodedValueDomain codedValueDomain = (CodedValueDomain) utilityNetworkAttribute.getDomain();
```
alan-edi, line 244: Extra comments about how we handle the Domain types might help. For example, here we could say something like:
> For CodedValueDomains, add the name of a CodedValue that matches the network attribute's comparison value (if any)

and below, at the start of the associated `else` block, something like:
> For other types of domain, add the name of the "other" attribute in the comparison, or the value in the comparison if there's no "other" attribute

alan-edi (commented):
@JonLavi Looks very good. My comments are mainly on the README file again.

JonLavi:
Thanks for the review @alan-edi ! I've updated with your suggestions.
@Rachael-E do you want the `ListCell` classes also removed here? You said in a different review that going forward you want to avoid using those to reduce the file count.

Rachael-E:
thanks @JonLavi, with this being such a long complex sample already, I think in this case it's good to have the ListCells abstracted out if you're ok with that. In the previous example we had, the sample code was less complex and smaller and made more sense to keep it "in class".

alan-edi (approved):
@JonLavi Looks great!

PR #442, opened 2020-02-14
Display annotation

JonLavi:
A new sample for the `AnnotationLayer` API.


JonLavi:
Hi @Rachael-E , would you please do a first review of this sample, thanks!

display_information/display-annotation/build.gradle
```diff
@@ -0,0 +1,78 @@
+plugins {
+    id 'application'
+    id 'org.openjfx.javafxplugin' version '0.0.5'
+}
+
+group = 'com.esri.samples'
+
+ext {
+    arcgisVersion = '100.8.0-2694'
```
Rachael-E: is this functionality released for 100.7.0, or should this be being merged into the dev branch?

display_information/display-annotation/README.metadata.json
```diff
@@ -0,0 +1,28 @@
+{
+    "category": "Layers",
```
Rachael-E: little rusty here, but doesn't look like we have a category for "Layers": looking at the package where this samples is saved, this should be "Display information"

Rachael-E (commented):
@JonLavi looks good to me thanks, just two things to double check

display_information/display-annotation/README.metadata.json
JonLavi: You're right, I've missed on changing this. Amended

display_information/display-annotation/build.gradle
JonLavi: I originally thought it's 100.8.0, but you're right, it's 100.7.0. Changed!

JonLavi:
Thanks for catching those details, @Rachael-E. Care to have another look please?

Rachael-E (approved):
thanks @JonLavi LGTM!

JonLavi:
@tschie could you please do a second review? Thanks!

display_information/display-annotation/src/main/java/com/esri/samples/display_annotation/DisplayAnnotationSample.java
```diff
@@ -0,0 +1,95 @@
...
+            map.getOperationalLayers().add(riverFeatureLayer);
+
+            // create an annotation layer from a feature service
+            AnnotationLayer annotationLayer = new AnnotationLayer("https://sampleserver6.arcgisonline.com/arcgis/rest/services/RiversAnnotation/FeatureServer/0");
+
+            // add the annotation layer to the map
+            map.getOperationalLayers().add(annotationLayer);
+
+            // add the map view to stack pane
+            stackPane.getChildren().addAll(mapView);
```
tschie:
```suggestion
            stackPane.getChildren().add(mapView);
```

display_information/display-annotation/README.md
```diff
@@ -0,0 +1,41 @@
+# Display annotation
+
+Display annotation from a feature service URL.
+
+![Image showing the Display Annotation sample](DisplayAnnotation.png)
+
+## Use case
+
+Annotation is useful for displaying text that you don't want to move or resize when the map is panned or zoomed (unlike labels which will move and resize). You can use annotation to place text at a fixed size, position, orientation, font, and so on. You may choose to do this for cartographic reasons or because the exact placement of the text is important.
```
tschie, line 9: > for cartographic reasons or because the exact placement of the text is important.

A real example would help this use case.

tschie (approved):
Looks good. Just a couple of small things

display_information/display-annotation/src/main/java/com/esri/samples/display_annotation/DisplayAnnotationSample.java
JonLavi: Good catch, thanks!

display_information/display-annotation/README.md
JonLavi: Very good point, I'll have a think and perhaps suggest an addition to the common readme.

PR #448, opened 2020-03-11
Update readmes to match common design (Display Information category) 2/2

JonLavi:


display_information/control-annotation-sublayer-visibility/README.md
```diff
@@ -1,33 +1,45 @@
...
+
+Use the checkboxes to manually set "Open" and "Closed" annotation sublayers visibility to on or off.
 
 ## How it works
 
-1. Load a `MobileMapPackage` containing annotation layers.
-2. Get the `AnnotationSublayer` from the `Annotation Layer` of the mobile map package.
-3. Set the annotation sublayer's visibility as required.
+1. Load the `MobileMapPackage`.
+2. Populate checkbox text with the `AnnotationSublayer` names (including maximum and minimum scales, where relevant).
```
Rachael-E: this whole how it works section isn't consistent with our other how it works sections :( it focusses on UI and not the API development workflow. It's unlikely this would go through a re-review in common samples. For Java we should at least pepper through some of the logic code for this if you don't feel comfortable restructuring this.  There are things I would remove though, such as the level of detail for changing the text to black and light gray: this is not something we should be setting out in the How it Works section since it has nothing to do with our API.

Rachael-E (changes requested):
@JonLavi all good bar one how it works section: I've left some comments there, thanks

display_information/control-annotation-sublayer-visibility/README.md
JonLavi: Totally agree @Rachael-E . I've revamped with a suggestion. Can you let me know what you think? Cheers!

Rachael-E (approved):
cool, looks good thanks!

PR #450, opened 2020-03-13
Update readmes to match common design (Editing category)

JonLavi:


editing/add-features/README.md
```diff
@@ -1,30 +1,34 @@
...
 
 ## How to use the sample
 
-Click on a location in the MapView to add a Feature at that location.
+Click on a location on the map to add a feature at that location.
 
 ## How it works
 
-To add a feature to a `ServiceFeatureTable` and update it's server with that `Feature`:
+A `Feature` instance is added to a `ServiceFeatureTable` which then pushes that new feature to the server.
```
Rachael-E: we can remove this since since the how it works section isn't meant to have an intro: all this info is covered in the steps below anyway.

editing/edit-feature-attachments/README.md
```diff
@@ -1,35 +1,42 @@
...
-To add an attachment.
-  - click on a feature and then click add attachment button in panel
+Attachments provide a flexible way to manage additional information that is related to your features. Attachments allow you to add files to individual features, including: PDFs, text documents, or any other type of file. For example, if you have a feature representing a building, you could use attachments to add multiple photographs of the building taken from several angles, along with PDF files containing the building's deed and tax information.
 
-To delete an attachment.
-  - click on a feature, select an attachment from attachment list, and click delete attachment button
+## How to use the sample
 
-## How it works
+Click a feature on the map to select it. Click 'Add attachment' to add an attachment to the selected feature. The list view will update to show the attachments of the selected feature (if any). To delete an attachment, first select the feature, and then select an attachment from the list view. Then click 'Delete Attachment' to delete the selected attachment.
```
Rachael-E: Match case from button text here, as is done the "Delete Attachment" reference
```suggestion
Click a feature on the map to select it. Click 'Add Attachment' to add an attachment to the selected feature. The list view will update to show the attachments of the selected feature (if any). To delete an attachment, first select the feature, and then select an attachment from the list view. Then click 'Delete Attachment' to delete the selected attachment.
```

editing/edit-feature-attachments/README.md
```diff
@@ -1,35 +1,42 @@
...
+
+Attachments can only be added to and accessed on service feature tables when their hasAttachments property is true.
+
 ## Relevant API
 
-* ArcGISFeature
-* ArcGISMap
-* Attachment
-* Feature
+* ApplyEditsAsync
```
Rachael-E: looks like there are a mixture of classes and methods in here. Mike and Tyler confirmed during the Android readme effort that "java se was the only platform for which api should be restricted to classes", so we should make this effort going forward just to strip this to relevant classes. This is because we have many more classes than the other SDKs and so the classes are therefore more relevant than they might be in other SDKs. 

So for this list, let's remove the method mentions, but add in ArcGISFeature (since at least one of the listed methods listed belongs to that). The others belong to FeatureLayer and ServiceFeatureTable as far as I can see.

editing/update-attributes/README.md
```diff
@@ -1,29 +1,33 @@
...
 
-To get a `Feature` from a `ServiceFeatureTable` and update its attributes:
-
-1. Create a service feature table from a URL.
-2. Create a `FeatureLayer` from the service feature table.
-3. Select features from the feature layer, `FeatureLayer.selectFeatures`.
-4. To update the feature's attribute use the `ArcGISFeature.getAttributes().put()`.
-5. After a change, apply the changes on the server using `ServiceFeatureTable.applyEditsAsync()`.
+1. Create a `ServiceFeatureTable` object from a URL.
+    * When the table loads, you can get the domain to determine which options to present in your UI.
```
Rachael-E: this line doesn't make sense for our Java sample

editing/update-attributes/README.md
```diff
@@ -1,29 +1,33 @@
...
-* ArcGISMap
-* Feature
+* ArcGISFeature
 * FeatureLayer
-* MapView
 * ServiceFeatureTable
+
+## Tags
+
+attribute, coded value, coded value domain, domain, editing, value
```
Rachael-E: these tags don't make sense for this sample? at least, we have no mention of domains, coded values or coded value domains throughout. I know in other sample we do. Might be worth checking with the team.

Rachael-E (commented):
thanks @JonLavi , a few things to look at but not much! Nice job!

editing/edit-feature-attachments/README.md
JonLavi: Not quite sure where this came from, sorry! Must have been at the end of my attention span

editing/update-attributes/README.md
JonLavi: Good spot, I'll double check what's going on here, but these definitely don't belong

editing/add-features/README.md
JonLavi: Yep, I've removed it

editing/edit-feature-attachments/README.md
JonLavi: updated!

editing/update-attributes/README.md
JonLavi: Removed!

JonLavi:
Thanks for the keen eye @Rachael-E. Updated with these small changes. Could you have another look please?

editing/update-attributes/README.md
```diff
@@ -30,4 +29,4 @@ To change the feature's damage property, click on the feature to select it, and
 
 ## Tags
 
-attribute, coded value, coded value domain, domain, editing, value
+amend, attribute, details, edit, editing, information, value
```
Rachael-E: maybe "update" here as well?

Rachael-E (approved):
cool, thanks @JonLavi ! Just an additional suggestion for the tags in update-attribute if you agree with it.

editing/update-attributes/README.md
JonLavi: Added :)

PR #451, opened 2020-03-13
Update readmes to match common design (Feature Layers category) 1/x

JonLavi:


feature_layers/change-feature-layer-renderer/README.md
```diff
@@ -1,27 +1,34 @@
-# Change Feature Layer Renderer
+# Change feature layer renderer
 
-Change how a feature layer looks with a renderer.
+Change the appearance of a feature layer with a renderer.
 
-![](ChangeFeatureLayerRenderer.gif)
+![Image of change feature layer renderer](ChangeFeatureLayerRenderer.png)
+
+## Use case
```
Rachael-E, line 7: again, another example of a sample where the readme design uses different data to our Java sample (ours still uses the pool permits, not the trees). I logged an issue for this in the Common Samples repo (no. 1471). The updated service is the following if you wanted to fix it up as part of this, otherwise let's revert this to the original description until we update the data. 

https://services.arcgis.com/V6ZHFr6zdgNZuVG0/arcgis/rest/services/Landscape_Trees/FeatureServer/0

feature_layers/feature-collection-layer/README.md
```diff
@@ -1,30 +1,37 @@
-# Feature Collection Layer
+# Feature collection layer
 
-Combine feature tables with different geometries into a single layer.
+Create a Feature Collection Layer from a Feature Collection Table, and add it to a map.
```
Rachael-E: the reference to these can all be lower case
```suggestion
Create a feature collection layer from a feature collection table, and add it to a map.
```

feature_layers/feature-layer-definition-expression/README.md
```diff
@@ -1,24 +1,33 @@
...
-1. Create a `ServiceFeatureTable` from a URL.
+1. Create a service feature table from a URL.
 2. Create a feature layer from the service feature table.
-3. Set the limit of the features on your feature layer using `FeatureLayer.setDefinitionExpression(Expression)`.
+3. Set the limit of the features on your feature layer using the `DefinitionExpression`.
 
 ## Relevant API
 
-* ArcGISMap
+* DefinitionExpression
```
Rachael-E: this isn't a class in our API, so we can remove (it's a setter method though)

feature_layers/feature-layer-definition-expression/README.md
```diff
@@ -1,24 +1,33 @@
...
 
 ## How it works
 
-To limit the `Feature`s in your `FeatureLayer`:
-
-1. Create a `ServiceFeatureTable` from a URL.
+1. Create a service feature table from a URL.
 2. Create a feature layer from the service feature table.
-3. Set the limit of the features on your feature layer using `FeatureLayer.setDefinitionExpression(Expression)`.
+3. Set the limit of the features on your feature layer using the `DefinitionExpression`.
```
Rachael-E: let's use our original instruction for this since this isn't a class in our API 

FeatureLayer.setDefinitionExpression("Expression")

Rachael-E (commented):
thanks @JonLavi, good that these readme's have highlighted another case of mismatching data. Otherwise just a removal of a non-class mention and some casing to fix up.

feature_layers/change-feature-layer-renderer/README.md
JonLavi: Of course, totally happy to update the data!

feature_layers/feature-layer-definition-expression/README.md
JonLavi: Yes, removed

feature_layers/feature-layer-definition-expression/README.md
JonLavi: Yes, I've reverted this

JonLavi:
Thanks @Rachael-E! I've updated the data on `ChangeFeatureLayerRenderer`, and also added a DoneLoadingListener to promote good practice with the loadable pattern. Could you have a look at the updated code please? Cheers!

feature_layers/change-feature-layer-renderer/src/main/java/com/esri/samples/change_feature_layer_renderer/ChangeFeatureLayerRendererSample.java
```diff
@@ -61,13 +61,13 @@ public void start(Stage stage) {
...
-      SimpleLineSymbol lineSymbol = new SimpleLineSymbol(SimpleLineSymbol.Style.SOLID, 0xFF0000FF, 2);
-      SimpleRenderer blueRenderer = new SimpleRenderer(lineSymbol);
+      // create a marker symbol renderer with a green circle
+      SimpleMarkerSymbol markerSymbol = new SimpleMarkerSymbol(SimpleMarkerSymbol.Style.CIRCLE, 0xFF317D11, 5 );
+      SimpleRenderer blueRenderer = new SimpleRenderer(markerSymbol);
 
       // create renderer toggle switch
       ToggleButton rendererSwitch = new ToggleButton();
-      rendererSwitch.setText("blue renderer");
+      rendererSwitch.setText("green renderer");
```
Rachael-E: we should also set the text to display "original renderer" when the green renderer is displaying (so the user knows when clicking the button it will go back to original renderer)

Rachael-E (commented):
ah thanks for updating that @JonLavi ! Can you also update it so that the text changes when the renderer changes? I've made a comment, thanks.

feature_layers/change-feature-layer-renderer/src/main/java/com/esri/samples/change_feature_layer_renderer/ChangeFeatureLayerRendererSample.java
JonLavi: Changing the toggle button's text gives a bit of a strange UI, I find, so I swapped it with a tickbox that says 'apply green renderer'. Also updated the screenshot/gif to match. What do you think?

Rachael-E:
I like the checkbox @JonLavi, although ofc it means it differs to the other SDKs implementations since they'll all be using the toggle button. I think the toggle button worked just as well as the checkbox, since if it was showing e.g. the green renderer then the button would say "show original renderer" so that is clear. It might be less faff in the long run for us to stick with the toggle button rather than ask every other SDK to use a checkbox. What do you think?

feature_layers/change-feature-layer-renderer/README.md
```diff
@@ -1,27 +1,34 @@
...
+
+A feature layer hosted on ArcGIS Online has a preset renderer and will display in an ArcGIS Runtime application with that renderer. However, for example, the color associated with the original renderer may be unsuitable for a company with staff or clients who are color blind, or for presentation with a different set of basemap and operational layers. In these cases, the renderer on the feature layer's data can be set to a more suitable color.
 
 ## How to use the sample
 
-Use the buttons in the control panel to change the renderer.
+Use the button to change the renderer on the feature layer. The original renderer displays orange circles, the diameters of which are proportional to carbon storage of each tree. When the blue renderer in this sample is applied, it displays the location of the trees simply as blue points.
```
Rachael-E: with the recent changes, there is no blue renderer: it switches between being green dots and orange circles. So we should change the sample to be blue rather than green me thinks

feature_layers/change-feature-layer-renderer/src/main/java/com/esri/samples/change_feature_layer_renderer/ChangeFeatureLayerRendererSample.java
```diff
@@ -61,53 +61,59 @@ public void start(Stage stage) {
       stage.setScene(scene);
       stage.show();
 
-      // create a blue (0xFF0000FF) line symbol renderer
-      SimpleLineSymbol lineSymbol = new SimpleLineSymbol(SimpleLineSymbol.Style.SOLID, 0xFF0000FF, 2);
-      SimpleRenderer blueRenderer = new SimpleRenderer(lineSymbol);
+      // create a marker symbol renderer with a green circle
```
Rachael-E: we should change this to blue

feature_layers/change-feature-layer-renderer/src/main/java/com/esri/samples/change_feature_layer_renderer/ChangeFeatureLayerRendererSample.java
```diff
@@ -61,53 +61,59 @@ public void start(Stage stage) {
...
+      SimpleMarkerSymbol markerSymbol = new SimpleMarkerSymbol(SimpleMarkerSymbol.Style.CIRCLE, 0xFF317D11, 5 );
```
Rachael-E: make the colour blue

Rachael-E (commented):
thanks @JonLavi, I left a comment in the conversation about the checkbox vs toggle button and also we need to ensure we're showing blue renderer as per the design. Thanks!

JonLavi:
> 
> 
> I like the checkbox @JonLavi, although ofc it means it differs to the other SDKs implementations since they'll all be using the toggle button. I think the toggle button worked just as well as the checkbox, since if it was showing e.g. the green renderer then the button would say "show original renderer" so that is clear. It might be less faff in the long run for us to stick with the toggle button rather than ask every other SDK to use a checkbox. What do you think?

Thanks @Rachael-E, I've updated as you suggested. I personally think that the 'pressed' state of the toggle button should be enough to inform the users about what happens when they 'de-press' it, that being that the renderer is removed. But, let's just move forward with this to get this merged. 
I wasn't able to create a gif with my machine (some problems with the software), but I don't think the gif is of much value here anyway, so I removed it. If you think we should still have one, perhaps you could spin one up? Thanks!

Rachael-E (approved):

PR #455, opened 2020-03-16
Update readmes to match common design (Geometry category) 1/x

JonLavi:


geometry/buffer/README.md
```diff
@@ -1,29 +1,35 @@
...
 
 ## How to use the sample
 
-Tap on the map to create planar (green) and geodesic (purple) buffers around the tapped location. Enter a value in the spinner between 500 and 2000 (in miles) to set the buffer distance (this range is appropriate to show the difference between geodesic and planar buffers).
+Click on the map. A planar and a geodesic buffer will be created at the click location using the distance (miles) specified in the text box. Continue clicking to create additional buffers. Notice that buffers closer to the equator appear similar in size. As you move north or south from the equator, however, the geodesic polygons become much larger. Geodesic polygons are in fact a better representation of the true shape and size of the buffer. Click `Clear` to remove all buffers and start again.
```
Rachael-E: We don't have a Clear button, but we do have a text input to determine the size of the buffer that should be mentioned here

geometry/buffer/README.md
```diff
@@ -1,29 +1,35 @@
...
 
 ## How it works
 
-To create a buffer around a point:
-
-1. Call `GeometryEngine.buffer` passing in a `Point` and a distance in meters. This returns a `Polygon` which can be displayed using a `Graphic`.
-2. For the geodesic buffer, call `GeometryEngine.bufferGeodetic` passing in the point, distance, linear unit, max deviation, and `GeodeticCurveType.GEODESIC`.
+1. The `Point` for a click on the map is captured.
```
Rachael-E: passive voice again which personally i think we should avoid.

geometry/buffer/README.md
```diff
@@ -1,29 +1,35 @@
...
+2. The static method `GeometryEngine.buffer()` is called to create a planar buffer polygon from the map location and distance.
+3. Another static method, `GeometryEngine.bufferGeodetic()` is called to create a geodesic buffer polygon using the same inputs.
+4. The polygon results (and click location) are displayed in the map view with different symbols in order to highlight the difference between the buffer techniques due to the spatial reference used in the planar calculation.
```
Rachael-E: this step is already described in the how to use section

Rachael-E (commented):
just some comments to think about here @JonLavi including an update to the UI, thanks

geometry/buffer/README.md
JonLavi: Yep, I've updated the whole section accordingly!

geometry/buffer/README.md
JonLavi: Good catch, I've removed the mention of the 'clear' box. The input field is mentioned in the beginning so I think we're ok?

JonLavi:
Thanks @Rachael-E , could you have a second look at the revised sections please?

geometry/buffer/README.md
```diff
@@ -10,14 +10,13 @@ Creating buffers is a core concept in GIS proximity analysis that allows you to
...
+Click on the map. A planar and a geodesic buffer will be created at the click location using the distance (miles) specified in the text box. Continue clicking to create additional buffers. Notice that buffers closer to the equator appear similar in size. As you move north or south from the equator, however, the geodesic polygons become much larger. Geodesic polygons are in fact a better representation of the true shape and size of the buffer.
 
 ## How it works
 
-1. The `Point` for a click on the map is captured.
-2. The static method `GeometryEngine.buffer()` is called to create a planar buffer polygon from the map location and distance.
-3. Another static method, `GeometryEngine.bufferGeodetic()` is called to create a geodesic buffer polygon using the same inputs.
-4. The polygon results (and click location) are displayed in the map view with different symbols in order to highlight the difference between the buffer techniques due to the spatial reference used in the planar calculation.
+1. Capture `Point`s when the `MapView` is clicked.
+2. Use The static method `GeometryEngine.buffer()` to create a planar buffer polygon from the map location and distance.
```
Rachael-E: lower case "the"
```suggestion
2. Use the static method `GeometryEngine.buffer()` to create a planar buffer polygon from the map location and distance.
```

Rachael-E (approved):
thanks @JonLavi just a wee typo to fix up then good to merge!

PR #456, opened 2020-03-16
Update readmes to match common design (Geometry category) 2/x

JonLavi:


geometry/densify-and-generalize/README.md
```diff
@@ -1,35 +1,35 @@
...
+The sample shows a polyline representing a ship's location at irregular intervals. The density of vertices along the ship's route is appropriate to represent the path of the ship at the sample map view's initial scale. However, that level of detail may be too great if you wanted to show a polyline of the ship's movement down the whole of the Willamette river. Then, you might consider generalizing the polyline to still faithfully represent the ship's passage on the river without having an overly complicated geometry.
 
-Use the sliders to control the parameters of the densify and generalize methods. You can deselect the checkboxes for either method to remove its effect from the result polyline. You can also hide the result to only see the original by deselecting the "Show result" checkbox.
+Densifying a multipart geometry can be used to more accurately represent curved lines or to add more regularity to the vertices making up a multipart geometry.
 
-## How it works
+## How to use the sample 
+
+Use the sliders to control the parameters of the densify and generalize methods. You can deselect the checkboxes for either method to remove its effect from the result polyline.
```
Rachael-E: We should keep in the part about the "Show result" part of the UI

geometry/geodesic-operations/README.md
```diff
@@ -1,35 +1,33 @@
...
 
 ## How it works
 
-To create a geodesic path between two points:
-
-1. Create a `Polyline` using two points.
-2. Pass this polyline to: `GeometryEngine.densifyGeodetic(polyline, segmentLength, unitOfMeasurement, GeodeticCurveType.GEODESIC)`. This will create a new polyline with segments of length `segmentLength` and `LinearUnit` set to `unitOfMeasurement`. The curve will be geodesic.
-3. You can set this geometry to a `Graphic` to display the curve in a `GraphicsOverlay`.
-4. To get the distance, use `GeometryEngine.lengthGeodetic(pathGeometry, unitOfMeasurement, GeodeticCurveType.GEODESIC)`
+1. Create a `Point` in New York City and display it as a `Graphic`.
```
Rachael-E: I made a comment here for Android too that we don't need to be so specific about where the point is: we removed the reference to NYC.

Rachael-E (approved):
this looks good, just 2 minor comments to consider then ok to merge!

geometry/densify-and-generalize/README.md
JonLavi: Agreed, I've added it back in.

geometry/geodesic-operations/README.md
JonLavi: That's a good point, I also found that a bit strange. It's now removed.

PR #458, opened 2020-03-16
Update readmes to match common design (Group Layers & Hydrography categories) 

JonLavi:


group_layers/group-layers/README.md
```diff
@@ -2,22 +2,27 @@
...
 
-The layers in the map will be displayed in a table of contents. Toggle the checkbox next to a layer's name to change its visibility.
+The layers in the map will be displayed in a table of contents. Toggle the checkbox next to a layer's name to change its visibility. Turning a group layer's visibility off will override the visibility of its child layers.
 
 ## How it works
 
 1. Create an empty `GroupLayer`.
 2. Add a child layer to the group layer's layers collection.
-3. To toggle the visibility of the group, simply change the group layer's visibility property.
+3. To build the table of contents:
```
Rachael-E: we do this quite differently in Java and the only step we follow here is checking if `isShowChildrenInLegend` is true. I'm not sure how best to write up our steps when a lot of it is Java logic to recursively populate the table of contents with layer names?

group_layers/group-layers/README.md
```diff
@@ -27,12 +32,8 @@ The layers in the map will be displayed in a table of contents. Toggle the check
 
 The full extent of a group layer may change when child layers are added/removed. Group layers do not have a spatial reference, but the full extent will have the spatial reference of the first child layer.
 
-Group layers can be saved to web scenes. In web maps, group layers will be flattened in the web map's operational layers.
-
-The implementation shown here makes use of a custom tree cell for displaying layers in a tree view. In the custom
```
Rachael-E, line 32: With the above comment in mind, we should keep this information in here for the Java sample.

Rachael-E (changes requested):
with this one we just need to move back in some of the removed Java specific instruction, thanks @JonLavi

group_layers/group-layers/README.md
JonLavi: Thanks for pointing this out. Considering most of the building of the 'tree' is UI focused, I've reverted back to the old java 'how it works' section, with a few additions. Looks like the main point is to show how to use the `GroupLayer` class and toggling visibility, so we don't need to do much more than that. What do you think?

group_layers/group-layers/README.md
JonLavi: Yep, I've added this back in. Thanks!

JonLavi:
Thanks @Rachael-E , could you have a second look at the updated sections please?

Rachael-E (approved):
cool thanks @JonLavi , and yup, I agree with just showing the main API call in the how it works section.

PR #459, opened 2020-03-16
Update readmes to match common design (Image Layers + KML category) 1/2

JonLavi:


image_layers/map-image-layer-sublayer-visibility/README.md
```diff
@@ -1,27 +1,28 @@
...
-To change visibility of a `ArcGISSubLayer` from your `ArcGISMap`:
-
-1. Create an `ArcGISMapImageLayer` from its URL.
-2. Add it to `ArcGISMap.getOperationalLayers().add()`.
-3. Display the ArcGISMap by adding it to the `MapView`.
-4. Gain access to the sub-layers from the `ArcGISMapImageLayer.getSubLayers()` method which returns a `SubLayerList`. The sub layer list is a modifiable list of ArcGISSubLayers.
-5. Determine if the layer is visible or not by turning on / off the ArcGIS sub layers visibility in the sub layer list.
+1. Create an `ArcGISMapImageLayer` object with the URL to a map image service.
+2. Get the `SubLayerList` with `mapImageLayer.getSubLayers()`.
+3. For each layer in the sublayer list, set its visible property to true or false.
```
Rachael-E: let's add a code instruction here "with .setVisible()" since the main point of the sample is to show how to set the visibility

image_layers/map-image-layer-tables/README.md
```diff
@@ -36,5 +37,8 @@ To query a map image layer's tables and find related features:
 
 ## Additional information
 
-You can use `ArcGISMapImageLayer.LoadTablesAndLayersAsync()` to recursively load all sublayers and tables
-associated with a map image layer.
+You can use `arcGISMapImageLayer.loadTablesAndLayersAsync()` to recursively load all sublayers and tables associated with a map image layer.
```
Rachael-E: Let's use the class format for this 
```suggestion
You can use `ArcGISMapImageLayer.loadTablesAndLayersAsync()` to recursively load all sublayers and tables associated with a map image layer.
```

image_layers/map-image-layer/README.md
```diff
@@ -1,21 +1,27 @@
...
 
-To add an `ArcGISMapImageLayer` to your `ArcGISMap` using its URL:
+## How it works
 
-1. Create an ArcGIS map image layer from its URL.
-2. Add it to `ArcGISMap.getOperationalLayers().add()`.
-3. Display the ArcGIS map by adding it to the `MapView`.
+1. Create an instance of `ArcGISMap`.
+2. Create an `ArcGISMapImageLayer` with the URL to a map image service.
+3. Add it to the map's operational layers collection.
```
Rachael-E: We should keep code instruction in here: it's helpful to the reader to see how to add it `ArcGISMap.getOperationalLayers().add(mapImageLayer)`

kml/create-and-save-kml-file/README.md
```diff
@@ -45,4 +44,4 @@ Click anywhere on the map to sketch. Press ENTER to commit the sketch or ESCAPE
```
Rachael-E, line 43: for the existing relevant API above there is a class referenced that we don't use in Java:

KmlNode.SaveAsAsync -> instead we use KmlDocument.saveAsAsync. So we can just remove this line since we already reference KmlDocument

kml/display-kml-network-links/README.md
```diff
@@ -1,17 +1,19 @@
...
+![Image of display KML network links](DisplayKmlNetworkLinks.png)
+
+## Use case
+
+KML files can reference other KML files on the network and support automatically refreshing content. For example, survey workers will benefit from KML data shown on their devices automatically refreshing to show the most up-to-date state. Additionally, discovering KML files linked to the data they are currently viewing provides additional information to make better decisions in the field.
 
 ## How to use the sample
 
-The data shown should refresh automatically every few seconds. A network message will be displayed when the dataset is loaded.
+The sample will load the KML file automatically. The data shown should refresh automatically every few seconds. Pan and zoom to explore the map.
```
Rachael-E: I think we should keep the info about showing a network message in: otherwise when you start the sample and see a message in Dutch it's a bit random! 

"A network message will be displayed when the dataset is loaded."

kml/display-kml-network-links/README.md
```diff
@@ -20,12 +22,12 @@ To show KML with network links and display network messages:
 
 * KmlDataset
 * KmlLayer
-* KmlNetworkLinkMessageReceivedEvent
+* KmlNetworkLinkControlMessageEventArgs
```
Rachael-E: this isn't in our API: instead use KmlNetworkLinkMessageReceivedEvent

kml/edit-kml-ground-overlay/README.md
```diff
@@ -15,12 +15,13 @@ Use the slider to adjust the opacity of the ground overlay.
 ## How it works
 
 1. Create an `Envelope` defining the geometry of the overlay.
-2. Create a `KmlIcon` using a  URI linking to an image.
+2. Create a `KmlIcon` using a `Uri` linking to an image.
```
Rachael-E: Since URI is an acronym and not a class in our API, we should keep it capitalized
```suggestion
2. Create a `KmlIcon` using a URI linking to an image.
```

kml/edit-kml-ground-overlay/README.md
```diff
@@ -15,12 +15,13 @@ Use the slider to adjust the opacity of the ground overlay.
...
 3. Create a `KmlGroundOverlay` using the envelope and icon.
-4. Create a `KmlDataset` using the ground overlay.
-5. Create a `KmlLayer` using the dataset.
-6. Add the KML layer to the scene.
-7. Use `kmlGroundOverlay.setColor(value)` with a different alpha value to change the opacity.
+4. Set the value of`KmlGroundOverlay.setRotation()`.
```
Rachael-E: space needed here
```suggestion
4. Set the value of `KmlGroundOverlay.setRotation()`.
```

kml/edit-kml-ground-overlay/README.md
```diff
@@ -35,4 +36,4 @@ The [image](https://libapps.s3.amazonaws.com/accounts/55937/images/1944.jpg) is
```
Rachael-E, line 36: In about the data, it's possible our alt text could do with updating: was this included in Trevor's updates he did in the common samples? If not might be worth checking with him if there's any room for improvement here.

kml/identify-kml-features/README.md
```diff
@@ -1,26 +1,46 @@
...
-2. Add a clicked listener to the map view with `mapView.setOnMouseClicked` to get the clicked screen point.
-3. Identify features at the screen location with `mapView.identifyLayerAsync(kmlLayer, screenPoint, tolerance, false)`.
-4. Use the returned future to get the `IdentifyLayerResult`.
-5. Loop through `result.getElements()`, looking for `GeoElements` which are instances of `KmlPlacemark`.
+1. Listen to clicks on the `MapView` using `setOnMouseClicked()`.
+2. When the map view is clicked:
+  * Dismiss the `Callout`, if one is showing.
+  * Call `identifyLayerAsync(...)` passing in the `KmlLayer`, screen point and tolerance.
+  * Await the result of the identify and then get the `KmlPlacemark` from the result.
+  * Create a callout at the calculated map point and populate the callout content with text from the placemark's `BalloonContent`. NOTE: KML supports defining HTML for balloon content and may need to be converted from HTML to text.
```
Rachael-E: `BalloonContent` is not a class in our API: however in our sample we call `getBalloonContent` on a KmlNode which returns a string. So I think the following small change is more accurate: 
```suggestion
  * Create a callout at the calculated map point and populate the callout content with text from the placemark's "balloon content". NOTE: KML supports defining HTML for balloon content and may need to be converted from HTML to text.
```

Rachael-E (changes requested):
some subtle things to look at here and removal of class references which aren't in our API, thanks

kml/display-kml-network-links/README.md
JonLavi: That's a very good point. I've added the info back in 😄

kml/edit-kml-ground-overlay/README.md
JonLavi: Good catch. I've had a go at updating following the pattern used in the other samples. Might have to change slightly if we get a different feedback.

JonLavi:
Thanks @Rachael-E ! I've updated the sections you pointed out. Would you be able to take a second look please?

Rachael-E (approved):

PR #460, opened 2020-03-16
Update readmes to match common design (KML category) 2/2

JonLavi:


kml/list-kml-contents/README.md
```diff
@@ -1,21 +1,23 @@
...
-To list the nodes in a KML file:
-
-1. Create a `KmlDataset` pointing to the KML file.
-2. Start with a list of the rood nodes with `kmlDataset.getRootNodes()`.
-3. For each node, check if it is a `KmlContainer` or `KmlNetworkLink`. These types can have child nodes. If it is one of these, cast to the appropriate type and call `getChildNodes()`.
-4. Recursively search these child nodes for more nodes.
+1. Add the KML file to the scene as a layer.
+2. Explore the root nodes of the `KmlDataset` recursively explored to create a view model.
+  * Each node is enabled for display at this step. KML files may include nodes that are turned off by default.
+3. When a node is selected, use the node's `Extent` to determine a viewpoint and set the `SceneView` object's viewpoint do it.
```
Rachael-E: We have no Extent class in our API, it's a method we call instead. Also a wee typo and an opportunity to format viewpoint as a class: 
```suggestion
3. When a node is selected, get its extent with selectedNode.getExtent(). Use that to create a `Viewpoint` and set the `SceneView` object's viewpoint to it.
```

kml/play-a-kml-tour/README.md
```diff
@@ -1,42 +1,37 @@
...
 
-KML, the file format used by Google Earth, supports creating tours, which can control the viewpoint of the scene, hide and show content, and play audio. Tours allow you to easily share tours of geographic locations, which can be augmented with rich multimedia.
+KML, the file format used by Google Earth, supports creating tours, which can control the viewpoint of the scene, hide and show content, and play audio. Tours allow you to easily share tours of geographic locations, which can be augmented with rich multimedia. Runtime allows you to consume these tours using a simple API.
 
 ## How to use the sample
 
-Click the play button to start the tour. The narration audio will start and then the viewpoint will animate. Press the button again to pause the tour.
-
-To start again from the beginning, click the reset button to reset the tour. Then press play.
+The sample will load the KMZ file from ArcGIS Online. When a tour is found, the _Play_ button will be enabled. Use _Play_ and _Pause_ to control the tour. When you're ready to show the tour, use the reset button to return the tour to the unplayed state.
```
Rachael-E: we need to tweak this for our sample, though I've noticed a UI ?bug/behaviour difference. To replay, first we have to hit refresh, and then play.

```suggestion
The sample will load the KMZ file from ArcGIS Online. Click the play button to start the tour. The narration audio will start and then the viewpoint will animate. Press the button again to pause the tour. To restart the tour, hit the refresh button and then the play button.
```

kml/play-a-kml-tour/README.md
```diff
@@ -1,42 +1,37 @@
...
 
 ## How it works
 
-1. Create a `KmlDataset` with the path to a local KML file with a KML tour.
-2. Create and load a `KmlLayer` with the dataset.
-3. When the layer has loaded, search its `KmlNode`s by recursing through `kmlLayer.getRootNodes()` to find a `KmlTour` node.
-4. Create a `KmlTourController` and set the tour with `kmlTourController.setTour(kmlTour)`.
-5. Use `kmltourController.play()`, `kmltourController.pause()`, and `kmltourController.reset()` to control the tour.
+1. Load the KML file and add it to a layer.
```
Rachael-E:
```suggestion
1. Create and load a `KmlLayer` with a `KmlDataset` and add it to the scene's operational layers.
```

kml/play-a-kml-tour/README.md
```diff
@@ -1,42 +1,37 @@
...
+2. Create the KML tour controller. Wire up the buttons to the `Play`, `Pause`, and `Reset` methods.
```
Rachael-E: We should remove the UI references here since they don't match Java anyway and we should refrain from UI wiring up in this section
```suggestion
2. Create a `KMLTourController` and set the tour with `kmlTourController.setTour(kmlTour)`.
```

kml/play-a-kml-tour/README.md
```diff
@@ -1,42 +1,37 @@
...
+3. Explore the tree of KML content to find the first KML tour. Once a tour is found, provide it to the KML tour controller.
```
Rachael-E: Since this isn't the order we do this in Java, I think we should use what we had previously for this section. It doesn't differ from the design fully and it's more important that our readme match how we write the sample in Java. 

I'd recommend the previous version for ordering, we could re-word it to match more closely the words used in the common design though if you felt it deviated too much. We just need to ensure the ordering is correct (e.g. the recursing through Kmlnode step should come before the above mention of setting up the KmlTourController)

kml/play-a-kml-tour/README.md
```diff
@@ -1,42 +1,37 @@
...
+4. Enable the buttons to allow the user to play, pause, and reset the tour.
```
Rachael-E:
```suggestion
4. Use `kmltourController.play()`, `kmltourController.pause()`, and `kmltourController.reset()` to control the tour.
```

Rachael-E (changes requested):
we need some tweaking here for the readmes to match what the samples are doing, as it differs slightly in Java compared to the common readme, thanks @JonLavi !

kml/play-a-kml-tour/README.md
JonLavi: No worries, happy to revert to the old descriptions and ordering. The contents are quite similar.

JonLavi:
Thanks @Rachael-E! I've implemented your suggestions, and reverted the 'how it works' section for the Play KML Tour sample. Would you mind having a second look?

Rachael-E (approved):
cool, LGTM, thanks!

PR #465, opened 2020-03-25
Update sample name in application

Rachael-E:
Just a small patch to update the name from "Find Connected Features in a Utility Network Sample" to "Trace a Utility Network Sample". I kept it all capitalized, though I'm rusty on format now for our JavaFX application naming. @JonLavi  please could you take a look?


JonLavi (approved):
Good catch! Yep, I think capitalized is good here.

PR #476, opened 2020-04-21
Update readmes to match common design (OGC category) 2/2

JonLavi:


ogc/wms-layer-url/README.md
```diff
@@ -1,18 +1,33 @@
-# WMS Layer URL
+# WMS layer URL
```
Rachael-E: should we add brackets round the URL here like we have in other readme updates? I'm easy either way.

ogc/wfs-xml-query/README.md
```diff
@@ -2,18 +2,21 @@
 
 Load a WFS feature table using an XML query.
 
-![](WfsXmlQuery.png)
+![Image of load WFS with XML query](WfsXmlQuery.png)
 
 ## Use case
 
-ArcGIS Runtime `QueryParameters` objects can't represent all possible queries that can be made against a WFS feature service. For example, Runtime query parameters don't support wildcard searches. However, queries can be provided as raw XML strings, allowing access to query functionality not available with `QueryParameters`.
+Runtime `QueryParameters` objects can't represent all possible queries that can be made against a WFS feature service. For example, Runtime query parameters don't support wildcard searches. You can provide queries as raw XML strings, allowing you to access query functionality not available with `QueryParameters`.
```
Rachael-E: We should specify ArcGIS Runtime when first referring to runtime: it's probably obvious but since it's an Esri product, using the full terminology on it's first use is best practice (in the same way ArcGIS Pro, ArcGIS Online are used etc). Probably best to use it the second time too
```suggestion
ArcGIS Runtime `QueryParameters` objects can't represent all possible queries that can be made against a WFS feature service. For example, ArcGIS Runtime query parameters don't support wildcard searches. You can provide queries as raw XML strings, allowing you to access query functionality not available with `QueryParameters`.
```

ogc/wms-layer-url/README.md
```diff
@@ -1,18 +1,33 @@
...
+
+## How to use the sample
+
+The map will load automatically when the sample starts.
 
 ## How it works
 
-To create a `WmsLayer` from a URL and display it:
+1. Create a `WmsLayer` specifying the URL of the service and the names of layers you want to display.
+    * **Note**: The name comes from the `Name` property, not the `Title` property. On many services, the title is human-readable while the name is a numeric identifier.
```
Rachael-E: we should avoid using backticks on non API methods/classes, suggest the following
```suggestion
    * **Note**: The name comes from the 'Name' property, not the 'Title' property. On many services, the title is human-readable while the name is a numeric identifier.
```

ogc/wmts-layer/README.md
```diff
@@ -1,23 +1,35 @@
...
-Dsiplay tiles from a Web Map Tile Service.
+Display a layer from a Web Map Tile Service.
+ 
+![Image of WMTS layer](WmtsLayer.png)
 
-![](WmtsLayer.png)
+## Use case
 
-## How it works
+WMTS services can have several layers. You can use Runtime to explore the layers available from a service. This would commonly be used to enable a browsing experience where users can choose which layers they want to display at run time.
```
Rachael-E:
```suggestion
WMTS services can have several layers. You can use ArcGIS Runtime to explore the layers available from a service. This would commonly be used to enable a browsing experience where users can choose which layers they want to display at run time.
```

ogc/wmts-layer/README.md
```diff
@@ -1,23 +1,35 @@
...
+## How it works
+ 
 1. Create a `WmtsService` using the URL of the WMTS Service.
-2. After loading the WmtsService, get the list of `WmtsLayerInfo`s from the service info: `service.getServiceInfo().getLayerInfos()`
-3. For the layer you want to display, get the layer ID using `getLayerInfos().get(0).getId()`
-4. Use the LayerInfo to create the WMTSLayer: `new WmtsLayer(layerInfos.get(0))`
-5. Set it as the maps' basemap with `map.setBasemap(new Basemap(wmtsLayer))`
+1. After loading the WMTS service, get the list of `WmtsLayerInfos` from the service info: `service.getServiceInfo().getLayerInfos()`
+1. Use one of the layer infos to create a new `WmtsLayer(layerInfos.get(0))`
+1. Set it as the maps' basemap with `map.setBasemap(new Basemap(wmtsLayer))`.
```
Rachael-E:
```suggestion
1. Set it as the map's basemap with `map.setBasemap(new Basemap(wmtsLayer))`.
```

Rachael-E (approved):
pretty minor suggestions here to look at, so will just approve so as not to hold up!

PR #477, opened 2020-04-21
Update readmes to match common design (OGC category) 1/2

JonLavi:


ogc/browse-wfs-layers/README.md
```diff
@@ -2,25 +2,23 @@
...
+![Image of browse WFS layers](BrowseWfsLayers.png)
 
 ## Use case
 
 Services often have multiple layers available for display. For example, a feature service for a city might have layers representing roads, land masses, building footprints, parks, and facilities. A user can choose to only show the road network and parks for a park accessibility analysis.
 
 ## How to use the sample
 
-A list of layers in the WFS service will be shown. Select a layer to display it.
+A list of layers in the WFS service will be shown. Select a layer to display.
```
Rachael-E: tiny suggestion but I think it improves readability
```suggestion
A list of layers in the WFS service will be shown. Select a layer to display it.
```

ogc/display-wfs-layer/README.md
```diff
@@ -16,7 +16,7 @@ Pan and zoom to see features within the current map extent.
 
 1. Create a `WfsFeatureTable` with a URL.
 2. Create a `FeatureLayer` from the feature table and add it to the map.
-3. Add a `NavigationChangedListener` to the map view and listen for a `NavigationChangedEvent`. Check if it `!isNavigating()` to detect when the user has stopped navigating the map.
+3. Add a `NavigationChangedListener` to the map view to listen for a `NavigationChangedEvent`. Check if it `isNavigating()` to detect when the user has stopped navigating the map.
```
Rachael-E: think it should be the following (as the original)?
```suggestion
3. Add a `NavigationChangedListener` to the map view to listen for a `NavigationChangedEvent`. Check if it `!isNavigating()` to detect when the user has stopped navigating the map.
```

ogc/open-street-map-layer/README.md
```diff
@@ -1,23 +1,35 @@
...
-* Basemap
-* Basemap.Type
+* Map
 * MapView
 
+## Additional information
+
+The attribution text will be set to the required OpenStreetMap attribution automatically. 
+
+Apps that expect to make many requests to OpenStreetMap should consider using an alternative tile server via the `WebTiledLayer` class. See [layer types described](https://developers.arcgis.com/android/latest/guide/layer-types-described.htm#ESRI_SECTION1_B995CCAB20584F91890B3614CF16CF43) in the *ArcGIS Runtime SDK for Android* documentation for more information on OpenStreetMap usage restrictions and alternatives.
```
Rachael-E: updating to Java doc
```suggestion
Apps that expect to make many requests to OpenStreetMap should consider using an alternative tile server via the `WebTiledLayer` class. See [layer types described](https://developers.arcgis.com/java/latest/guide/layer-types-described.htm#ESRI_SECTION1_B995CCAB20584F91890B3614CF16CF43) in the *ArcGIS Runtime SDK for Java* documentation for more information on OpenStreetMap usage restrictions and alternatives.
```

Rachael-E (commented):
just a few wee suggestions to consider

ogc/open-street-map-layer/README.md
JonLavi: Good catch, thank you!

ogc/display-wfs-layer/README.md
JonLavi: I personally think the `!` operator is just easy to miss here and can be confusing. If you check `isNavigating()` you either get true or false, and then the user can deduct whether naviation has stopped or is still going. Therefore it might not be necessary to do `!` even though we do it in the code?

Rachael-E (approved):

PR #481, opened 2020-04-21
Update readmes to match common design (Scene category) 2/5

JonLavi:


scene/create-terrain-surface-from-local-tile-package/README.md
```diff
@@ -1,22 +1,32 @@
-# Create Terrain from a Local Tile Package
+# Create terrain from a local tile package
 
 Set the terrain surface with elevation described by a local tile package.
 
-The terrain surface is what the basemap, operational layers, and graphics are draped on. The tile package must be a LERC (limited error raster compression) encoded TPK. Details on creating these are in the [ArcGIS Pro documentation](https://pro.arcgis.com/en/pro-app/help/sharing/overview/tile-package.htm).
+![Image of create terrain from a local tile package](CreateTerrainSurfaceFromLocalTilePackage.png)
 
-![](CreateTerrainSurfaceFromLocalTilePackage.png)
+## Use case
```
Rachael-E, line 7: this is the same use case as for terrain surface: just checking this is intentional? I would have thought the use case would be more specific to the local tile package.

scene/create-terrain-surface-from-local-tile-package/README.md
```diff
@@ -1,22 +1,32 @@
...
+
+In a scene view, the terrain surface is what the basemap, operational layers, and graphics are draped onto. For example, when viewing a scene in a mountainous region, applying a terrain surface to the scene will help in recognizing the slopes, valleys, and elevated areas.
+
+## How to use the sample
+
+When loaded, the sample will show a scene with a terrain surface applied. Pan and zoom to explore the scene and observe how the terrain surface allows visualizing elevation differences.
```
Rachael-E, line 13: Same with this section: feels like it should be tweaked slightly to allude to viewing the terrain surface from the local tile package applied.

Rachael-E (commented):
just a query about the use case and how it works on create-terrain-surface-from-local-tile-package to double check

scene/create-terrain-surface-from-local-tile-package/README.md
JonLavi: I think it's pretty much the same, just a different data input source?

Rachael-E: that's fine, just wanted to double check

Rachael-E (approved):

PR #485, opened 2020-04-21
Raster category readmes update 1

Rachael-E:
First batch of 5 sample readme updates for the Raster category.

### DIFF
--- a/raster/blend-renderer/README.md
+++ b/raster/blend-renderer/README.md
@@ -1,33 +1,34 @@
-# Blend Renderer
+# Blend renderer
 
-Apply a blend renderer to a raster.
+Blend a hillshade with a raster by specifying the elevation data. The resulting raster looks similar to the original raster, but with some terrain shading, giving it a textured look.
 
-Blend renderers can be used to blend elevation data with imagery, creating a 3D effect.
+![Image of blend renderer](BlendRenderer.png)
 
-![](BlendRenderer.png)
+## Use case
+
+A blend renderer can be used to apply a color ramp to a hillshade to emphasize areas of high or low elevation. A blend renderer can also be used to add a hillshade effect to aerial or satellite imagery, thereby making changes in elevation more visible.
 
 ## How to use the sample
 
-Choose and adjust the settings to update the blend renderer on the raster layer. To use a color ramp instead of the satellite imagery, choose the color ramp type NONE.
+Choose and adjust the altitude, azimuth, slope type and color ramp type settings to update the image.
 
 ## How it works
 
-To apply a `BlendRenderer` to a `RasterLayer`:
-
-1. Create a `Raster` from a raster file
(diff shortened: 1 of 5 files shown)


raster/blend-renderer/README.md
```diff
@@ -1,33 +1,34 @@
...
-4. Create a `Raster` for elevation from a grayscale raster file
-5. Create a `BlendRenderer`, specifying the elevation raster, color ramp, and other properties
-    * If you specify a non-null color ramp, use the elevation raster as the base raster in addition to the elevation raster parameter. That way the color ramp is used instead of the satellite imagery
-6. Set the renderer on the raster layer with `rasterLayer.setRenderer(renderer)`.
+1. Create a `Raster` object from a raster file.
+2. Create a `RasterLayer` object from the raster.
+3. Create a `Basemap` object from the raster layer and set it to the map.
+4. Create another `Raster` object for elevation from a grayscale raster file.
+5. Create a `BlendRenderer` object, specifying the elevation raster, color ramp, and other properties.
+	- If you specify a non-null color ramp, use the elevation raster as the base raster in addition to the elevation raster parameter. That way, the color ramp is used instead of the satellite imagery.
```
JonLavi: Lets keep this consistent by using four spaces and an `*`
```suggestion
   * If you specify a non-null color ramp, use the elevation raster as the base raster in addition to the elevation raster parameter. That way, the color ramp is used instead of the satellite imagery.
```

raster/raster-layer-file/README.md
```diff
@@ -1,21 +1,32 @@
-# Raster Layer File
+# Raster layer (file)
```
JonLavi, line 1: I prefer the `(file)` in parentheses, and even better would be naming this sample `raster layer from file`.
However it now diverges from the original sample name. Do we want to go ahead with renaming anyway?

raster/raster-layer-file/README.md
```diff
@@ -1,21 +1,32 @@
...
 * Raster
 * RasterLayer
+
+## Additional information
+
+See the topic [What is raster data?](http://desktop.arcgis.com/en/arcmap/10.3/manage-data/raster-and-images/what-is-raster-data.htm) in the *ArcMap* documentation for more information about raster images.
+
+## Tags
+
+data, image, import, layer, raster, visualization
```
JonLavi, line 32: Same with this PR, we need to make sure there and empty newline at the end of each file

JonLavi (approved):
Looks good @Rachael-E . Just a few minor things, so I'll approve so that you can merge when you're happy with it

raster/raster-layer-file/README.md
Rachael-E: after discussions about similar from the Android team, the renaming of samples will be another effort. So let's stick with this for now and then when the big name overhaul is done this can be updated then (along with our sample names in code).

PR #489, opened 2020-04-21
Raster category readme updates 2

Rachael-E:
Second batch of readme updates for the Raster category.

### DIFF
--- a/raster/raster-layer-geopackage/README.md
+++ b/raster/raster-layer-geopackage/README.md
@@ -1,23 +1,34 @@
-# Raster Layer GeoPackage
+# Raster layer (GeoPackage)
 
-Display raster data from a geopackage.
+Display a raster contained in a GeoPackage.
 
-![](RasterLayerGeopackage.png)
+![Image of raster layer GeoPackage](RasterLayerGeopackage.png)
+
+## Use case
+
+GeoPackages are designed to simplify file management and transfer. They can store raster files (as well as other types of data). An end-user wishing to transfer rasters from ArcGIS Pro or between runtime apps might need to import raster files from GeoPackages into their map to view and analyze the data.
+
+## How to use the sample
+
+When the sample starts, a raster will be loaded from a GeoPackage and displayed in the map view.
 
 ## How it works
 
-To add a`RasterLayer` as an operational layer from a `GeoPackage`:
-
 1. Create and load a `GeoPackage`, specifying the path to the local .gpkg file.
-2. When it is done loading, get the `GeoPackageRaster`s inside with `geoPackage.getGeoPackageRasters()`.
-3. Construct a `RasterLayer` with the `GeoPackageRaster` in the list you want to use.
-4. Add the raster layer to the map as an operational layer `map.getOperationalLayers().add(rasterLayer)`.
+2. When it is done loading, get the `GeoPackageRaster`s from the geopackage.
+3. Construct a `RasterLayer` with the geopackage raster you want to use.
+4. Add the raster layer to the map.
 
 ## Relevant API
 
-* ArcGISMap
-* Basemap
 * GeoPackage
 * GeoPackageRaster
-* MapView
 * RasterLayer
+
+## About the data
+
+The [Aurora Colorado GeoPackage](https://www.arcgis.com/home/item.html?id=68ec42517cdd439e81b036210483e8e7) holds datasets that cover Aurora, Colorado. The raster that is opened and displayed shows airport noise levels.
+
+## Tags
+
+container, data, image, import, layer, OGC, package, raster, visualization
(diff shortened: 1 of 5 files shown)


raster/raster-layer-geopackage/README.md
```diff
@@ -1,23 +1,34 @@
-# Raster Layer GeoPackage
+# Raster layer (GeoPackage)
```
JonLavi, line 1: Should we be adding the parentheses here if they weren't there before?

raster/raster-layer-geopackage/README.md
```diff
@@ -1,23 +1,34 @@
...
+When the sample starts, a raster will be loaded from a GeoPackage and displayed in the map view.
+
+## How it works
 
 1. Create and load a `GeoPackage`, specifying the path to the local .gpkg file.
-2. When it is done loading, get the `GeoPackageRaster`s inside with `geoPackage.getGeoPackageRasters()`.
-3. Construct a `RasterLayer` with the `GeoPackageRaster` in the list you want to use.
-4. Add the raster layer to the map as an operational layer `map.getOperationalLayers().add(rasterLayer)`.
+2. When it is done loading, get the `GeoPackageRaster`s from the geopackage.
+3. Construct a `RasterLayer` with the geopackage raster you want to use.
```
JonLavi: Shuld these not say 'GeoPackage' since it's a named term? We're also writing 'GeoPackage' in all the sections above

```suggestion
2. When it is done loading, get the `GeoPackageRaster`s from the GeoPackage.
3. Construct a `RasterLayer` with the GeoPackage raster you want to use.
```

raster/raster-layer-url/README.md
```diff
@@ -1,8 +1,16 @@
-# Raster Layer URL
+# Raster layer URL
```
JonLavi: Just to highlight, we're not using the parenteses here around 'URL', so whichever way we choose, perhaps we should keep it consistent.

raster/rgb-renderer/README.md
```diff
@@ -1,31 +1,37 @@
...
 
-![](RgbRenderer.png)
+## Use case
+
+An RGB renderer is used to adjust the color bands of a multispectral image. Remote sensing images acquired from satellites often contain values representing the reflection of multiple spectrums of light. Changing the RGB renderer of such rasters can be used to differentiate and highlight particular features that reflect light differently, such as different vegetation types, or turbidity in water.
 
 ## How to use the sample
 
-Choose one of the stretch parameter types. The other options will adjust based on the chosen type. Add your inputs and press the Update button to update the renderer.
+Choose one of the stretch parameter types. The other options will adjust based on the chosen type. Add your inputs and select the 'Render' button to update the renderer.
```
JonLavi: In our sample the button is called 'Update'. I also think it's more correct to say 'click' rather than 'select' when talking about buttons:

```suggestion
Choose one of the stretch parameter types. The other options will adjust based on the chosen type. Add your inputs and click the 'Update' button to update the renderer.
```

raster/rgb-renderer/README.md
```diff
@@ -1,31 +1,37 @@
...
-1. Create a `Raster` from a multispectral raster file
-2. Create a `RasterLayer` from the raster
-3. Create a `Basemap` from the raster layer and set it to the map
-4. Create a `RgbRenderer`, specifying the stretch parameters and other properties
-5. Set the renderer on the raster layer with `rasterLayer.setRenderer(renderer)`
+1. Create a `Raster` from a multispectral raster file.
+2. Create a `RasterLayer` from the raster.
+3. Create a `Basemap` from the raster layer and instantiate an ArcGISMap with it.
+4. Create a `RGBRenderer`, specifying the `StretchParameters` and other properties.
+5. Set the `Renderer` on the raster layer with `rasterLayer.setRasterRenderer(renderer)`.
```
JonLavi: We've put `RGBRenderer` in backticks in step 4, so it's probably fine to just write renderer regularly in this step. We do the same in Stretch Renderer

```suggestion
5. Set the renderer on the raster layer with `rasterLayer.setRasterRenderer(renderer)`.
```

raster/stretch-renderer/README.md
```diff
@@ -1,31 +1,49 @@
...
+* PercentClipStretchParameters
 * Raster
 * RasterLayer
+* StandardDeviationStretchParameters
 * StretchParameters
 * StretchRenderer
+
+## About the data
+
+This sample uses a raster imagery tile of an area of forested mountainous terrain and rivers.
```
JonLavi: This sample uses the same data as RGB renderer so we can just steal the section from there:

```suggestion
The raster used in this sample shows an area in the south of the Shasta-Trinity National Forest, California.
```

JonLavi (commented):
Nice work @Rachael-E . Just a few minor suggestions

raster/raster-layer-geopackage/README.md
Rachael-E: yes: parentheses are ok to add here since they won't affect the samples names in code

Rachael-E:
thanks @JonLavi ! could you take another look please?

JonLavi (approved):
Looks good @Rachael-E . Good catch on the var naming 👍

PR #495, opened 2020-04-27
small readme updates

Rachael-E:
Added in the missing About the Data section to service feature manual cache (matching the data we use in the sample) and removing the reference to the "free" ArcGIS Online account which the team agreed caused confusion.


JonLavi (approved):
Thanks @Rachael-E for going back and fixing these!

PR #496, opened 2020-04-28
Update service URLs to use HTTPS

JonLavi:
I've gone ahead and updated the URLs to services to use HTTPS. This is particualrly important with samples that use authentication, since they might not be able to get past the auth challenges when the URL is specified as `http`.

URLs that I haven't updated:
`http://www.apache.org/licenses/LICENSE-2.0` (from the license comment at the top of each file
`url 'http://olympus.esri.com/artifactory/arcgisruntime-repo'` from the build.gradle files
`<fx:root type="javafx.scene.control.Dialog" xmlns:fx="http://javafx.com/fxml">` in FXML files

These would probably work as HTTPS, but it isn't critical, I think.

### DIFF
--- a/map/apply-scheduled-updates-to-preplanned-map-area/gradlew
+++ b/map/apply-scheduled-updates-to-preplanned-map-area/gradlew
@@ -7,7 +7,7 @@
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
 #
-#      https://www.apache.org/licenses/LICENSE-2.0
+#      http://www.apache.org/licenses/LICENSE-2.0
 #
 # Unless required by applicable law or agreed to in writing, software
 # distributed under the License is distributed on an "AS IS" BASIS,
--- a/map/apply-scheduled-updates-to-preplanned-map-area/gradlew.bat
+++ b/map/apply-scheduled-updates-to-preplanned-map-area/gradlew.bat
@@ -5,7 +5,7 @@
 @rem you may not use this file except in compliance with the License.
 @rem You may obtain a copy of the License at
 @rem
-@rem      https://www.apache.org/licenses/LICENSE-2.0
+@rem      http://www.apache.org/licenses/LICENSE-2.0
 @rem
 @rem Unless required by applicable law or agreed to in writing, software
 @rem distributed under the License is distributed on an "AS IS" BASIS,
--- a/map/apply-scheduled-updates-to-preplanned-map-area/gradlew.bat
+++ b/map/apply-scheduled-updates-to-preplanned-map-area/gradlew.bat
@@ -5,7 +5,7 @@
 @rem you may not use this file except in compliance with the License.
 @rem You may obtain a copy of the License at
 @rem
-@rem      http://www.apache.org/licenses/LICENSE-2.0
+@rem      https://www.apache.org/licenses/LICENSE-2.0
 @rem
 @rem Unless required by applicable law or agreed to in writing, software
 @rem distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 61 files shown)


map/apply-scheduled-updates-to-preplanned-map-area/gradlew
```diff
@@ -7,7 +7,7 @@
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
 #
-#      https://www.apache.org/licenses/LICENSE-2.0
+#      http://www.apache.org/licenses/LICENSE-2.0
```
Rachael-E, line 10: just curious about this change: you mentioned that this PR doesn't address the Apache licenses, so is this to bring this sample in line with the way the other license references are written?

map/apply-scheduled-updates-to-preplanned-map-area/gradlew.bat
```diff
@@ -5,7 +5,7 @@
 @rem you may not use this file except in compliance with the License.
 @rem You may obtain a copy of the License at
 @rem
-@rem      https://www.apache.org/licenses/LICENSE-2.0
+@rem      http://www.apache.org/licenses/LICENSE-2.0
```
Rachael-E: not sure if we should be changing gradlew.bat files: but is this just automatic from having made the above change?

scene/change-atmosphere-effect/src/main/java/com/esri/samples/change_atmosphere_effect/ChangeAtmosphereEffectSample.java
```diff
@@ -67,8 +67,8 @@ public void start(Stage stage) {
 
       // add base surface for elevation data
       Surface surface = new Surface();
-      ArcGISTiledElevationSource elevationSource = new ArcGISTiledElevationSource( "http://elevation3d.arcgis" +
-          ".com/arcgis/rest/services/WorldElevation3D/Terrain3D/ImageServer");
+      ArcGISTiledElevationSource elevationSource = new ArcGISTiledElevationSource(
+              "https://elevation3d.arcgis.com/arcgis/rest/services/WorldElevation3D/Terrain3D/ImageServer");
```
Rachael-E, line 71: looks like some excess spaces have creeped in here

Rachael-E (approved):
cool thanks @JonLavi just a query on the Apache & gradlew.bat change and the removal of some spaces needed.

map/apply-scheduled-updates-to-preplanned-map-area/gradlew
JonLavi: This wasn't intentional. I've done a find&replace to make these `https` originally, and then F&R to put them all back to `http`. So looks like this one was https in the first place. We can change it or leave it, it doesn't really matter I think.

map/apply-scheduled-updates-to-preplanned-map-area/gradlew.bat
JonLavi: Yeah this will have crept in with a find & replace. We can undo this one just to be safe

```suggestion
@rem      https://www.apache.org/licenses/LICENSE-2.0
```

scene/change-atmosphere-effect/src/main/java/com/esri/samples/change_atmosphere_effect/ChangeAtmosphereEffectSample.java
JonLavi: The indentation happens automatically when you break a line after a `(`, so it should be ok.

JonLavi:
Thanks @Rachael-E . I've reverted the instances that were changed accidentally through find&replace.

tschie (approved):
🤞 hopefully doesn't break everything

JonLavi:
Thanks both!
@tschie I forgot to mention in the PR that I did test the samples where the URL was modified, and they all worked. So, hopefully nothing should break 🤞

PR #498, opened 2020-04-28
Service Area Task - update UI

JonLavi:
I think we could lay out the UI of this sample a bit neater. Currently the buttons are all in a long row without any space. 

This PR puts the buttons in a control area, and re-arranges the buttons slightly so that it's clearer that the toggle group buttons (create facility, create barrier) are related. 

It's a bit of extra FXML code but I do think it makes the sample easier to use and it looks much smoother


network_analysis/service-area-task/src/main/resources/service_area_task.fxml
```diff
@@ -15,31 +15,55 @@
...
+        </background>
+
+        <HBox>
+            <ToggleButton fx:id="btnAddFacility" text="Add Facilities" minWidth="90">
+                <toggleGroup>
+                    <ToggleGroup fx:id="toggleGroup"/>
+                </toggleGroup>
+            </ToggleButton>
+            <ToggleButton fx:id="btnAddBarrier" toggleGroup="$toggleGroup" text="Add Barrier" onAction="#createBarrier"
+                          minWidth="90"/>
```
tschie:
```suggestion
            <fx:define>
                <ToggleGroup fx:id="toggleGroup"/>
            </fx:define>
            <ToggleButton fx:id="btnAddFacility" text="Add Facilities" minWidth="90" toggleGroup="$toggleGroup"/>
            <ToggleButton fx:id="btnAddBarrier" text="Add Barrier"  minWidth="90"  toggleGroup="$toggleGroup" onAction="#createBarrier"/>
```

tschie: I think using define is a bit cleaner

network_analysis/service-area-task/src/main/resources/service_area_task.fxml
```diff
@@ -15,31 +15,55 @@
...
+        </HBox>
+        <Button text="Show Service Areas" onAction="#showServiceAreas" minWidth="180"/>
+        <Button text="Reset" onAction="#clearRouteAndGraphics" minWidth="180"/>
```
tschie: Is there a way we can clean up all of these min/max widths? Not a big deal if not.

tschie (approved):
This is a welcome change. Just one suggestion.

network_analysis/service-area-task/src/main/resources/service_area_task.fxml
JonLavi: Definitely, thanks for pointing that out!

network_analysis/service-area-task/src/main/resources/service_area_task.fxml
JonLavi: Good point, I've gone ahead and put all the styling in the CSS file. Perhaps we can do this with samples going forward? Makes the FXML look nice and clean.

Rachael-E (approved):
cool, looks good!

JonLavi:
Thanks both!

PR #500, opened 2020-04-28
Viewshed Geoanalysis - clear previous results

JonLavi:
At the moment, clicking the map adds an additional source for a viewshed, but doesn't clear the old ones.
I've added a couple of lines to clear the previous result before a new result is added.
Sampleserver6 seems to be acting up at the moment so I can't run the sample to update with a new screenshot, therefore still a 'draft'


Rachael-E (approved):
nice find! this is a great improvement.

Rachael-E:
SS6 works ok for me today: a few ones didn't return but after a few goes it worked fine.

tschie (approved):

PR #501, opened 2020-04-28
Display KML - add loading indicator

JonLavi:
I think it would make sense to add a spinner to show when URL/Portal items are being loaded? At the moment this server is quite slow so it's confusing when selecting something from the drop-down and nothing happens for 10-20 seconds.


Rachael-E (approved):
cool, yup good UX improvement: we did the same in Android when we converted to Kotlin!

PR #502, opened 2020-04-28
Terrain exaggeration - make slider smooth and add label

JonLavi:
The exaggeration slider snaps to whole (int) values, but the code casts the value to (float). I vaguely remember there being a reason for this. But perhaps we can make the slider go smoothly throughout a range, as well as have the exaggeration update 'live', and not only when releasing the slider drag?


Rachael-E (approved):
lovely! really nice UI improvement @JonLavi

JonLavi:
Hi @tschie could you do a second review please?

tschie (approved):
Much better!

PR #503, opened 2020-04-28
Export (Vector) Tiles - UI and readability improvements

JonLavi:
With these two samples, at the moment you can start an export simultaneously as many times as you want since the button doesn't disable. What this PR does is disable the button so that only one job can run at the same time.
I've also added some whitespace and extra comments to make the code a bit more readable.


Rachael-E (approved):
cool, nice improvement: much better than the one before. Still takes an age to export the thing! Might just be my internet though.

JonLavi:
Hi @tschie , would you mind reviewing this when you have a moment? The diff makes it look a lot bigger than the change actually is..

tiled_layers/export-tiles/src/main/java/com/esri/samples/export_tiles/ExportTilesSample.java
```diff
@@ -114,62 +115,87 @@ public void start(Stage stage) {
...
-                  alert.show();
-                }
-                Platform.runLater(() -> progressBar.setVisible(false));
-              });
-            } catch (InterruptedException | ExecutionException ex) {
-              Alert alert = new Alert(Alert.AlertType.ERROR, ex.getMessage());
-              alert.show();
-              progressBar.setVisible(false);
-              progressBar.setProgress(0);
+          try {
```
tschie: looks like an extra indent here

tschie (approved):

tiled_layers/export-tiles/src/main/java/com/esri/samples/export_tiles/ExportTilesSample.java
JonLavi: Thanks, good spot!

PR #504, opened 2020-04-29
Show callout - update UI interaction

JonLavi:
Currently with this sample, clicking shows a callout. Clicking on the map again makes the callout disappear. Another click is then required to make the callout appear at a new location. I think one click should be enough to make a callout appear at a new location, so this PR is to make that small change


Rachael-E (approved):
nice!

JonLavi:
Hi @tschie could you let me know if you're happy with this small change?

display_information/show-callout/src/main/java/com/esri/samples/show_callout/ShowCalloutSample.java
```diff
@@ -59,31 +59,30 @@ public void start(Stage stage) {
...
       // click event to display the callout
       mapView.setOnMouseClicked(e -> {
-        // check that the primary mouse button was clicked and user is not
-        // panning
+
+        // check that the primary mouse button was clicked and user is not panning
         if (e.isStillSincePress() && e.getButton() == MouseButton.PRIMARY) {
+
+          // hide any previous callout
+          callout.dismiss();
```
tschie: I think this can be omitted since `showCalloutAt` will "dismiss" the callout for you. Dismissing with a right-click would be helpful to show though.

tschie (approved):
Good stuff. Just one suggestion.

display_information/show-callout/src/main/java/com/esri/samples/show_callout/ShowCalloutSample.java
JonLavi: Thanks, that's a good shout. Added a dismiss on secondary click.

PR #505, opened 2020-04-29
Display Grid - remove update button and apply changes immediately

JonLavi:
This sample has quite a few options to change, and I found it's cumbersome to have to press 'update' every time. This PR simply makes the `updateGrid()` method trigger every time a UI element is interacted with, so we don't have to go and press 'Update' manually. I've also revised the readme and the screenshot to match the new UI


Rachael-E:
hey @JonLavi, thanks for posting up a suggestion. I think we should park this one until we get feedback across this board. The reason being that this feels like quite a big design change from a UX perspective and should be proposed to the sample team first. Other SDKs still have the update button. From my perspective, it helps to be in control of all the options and then hit update rather than them happening on the fly since they might be subtle. It would be great to get the rest of the teams feedback on this though.

JonLavi:
> 
> 
> hey @JonLavi, thanks for posting up a suggestion. I think we should park this one until we get feedback across this board. The reason being that this feels like quite a big design change from a UX perspective and should be proposed to the sample team first. Other SDKs still have the update button. From my perspective, it helps to be in control of all the options and then hit update rather than them happening on the fly since they might be subtle. It would be great to get the rest of the teams feedback on this though.

Thanks @Rachael-E . There's no common issue for a 'redesign', since the change is considered  too trivial for a complete overhaul. I find that if changes are subtle, you would want to change only one parameter at a time, and then see it as soon as possible, without having to click an extra button in between. However if you don't think this change is worth it, just close the PR. Thanks!

JonLavi:
Closing as it's not viable to change the common design itself for such a trivial modification

PR #506, opened 2020-04-29
Convex Hull - add background to VBOX

JonLavi:
At the moment it's possible to click through the buttons when they are disabled, which is quite confusing. I've added a background to the VBox that holds the buttons so that it's not possible to click through them.

Screenshot doesn't need updating as it doesn't show the UI


Rachael-E (approved):
this works well, good find

JonLavi:
Hi @tschie could you do a second review on this change please?

tschie (approved):

PR #507, opened 2020-05-01
Update Graphic - Update code and improve UI

JonLavi:
When using this sample in regression testing I found that the UI seems to be enabled before any graphics were selected, and didn't disable properly when no graphic was selected. I've gone and fixed this, and rearranged the code a little so that it matches the more recent samples' structure.

I've also removed the `doneLoadingListener` from this sample since there's no direct interation with the basemap.


Rachael-E (approved):
this makes this sample run a little smoother from the UX perspective, it was too easy to click around initially before understanding the point.

JonLavi:
@tschie would you mind doing a second review when you have a moment please?

tschie (commented):
I think this could still be improved. Could we click and drag a graphic to show it moving with the mouse? We could change the scene's Cursor to `Cursor.MOVE` when a drag starts (no selection needed). Bonus points if you could press Escape to cancel the move. The Update Description button and Update Symbol comboBox could become active if there is a selection (meaning there was a click but no drag). That seems like a more realistic editing workflow.

JonLavi:
Those are some good suggestions @tschie!
I've had a quick look at implementing hold+drag, and it looks like it's going to require a bit more extra effort.
Considering I have a bit more limited scope for samples just now, how about we'll merge these changes first and then create another issue for updating with your suggestions at a later point?

PR #509, opened 2020-05-04
New sample: Animate images image overlay

Rachael-E:
I'm assuming by the time this gets through review etc that this will be going into the new release master, so that's why I chose master branch rather than dev. @JonLavi would you be able to take a first review please when you have a moment? this is my first Java sample in a while...hope I haven't missed anything too obvious off!


scene/animate-images-with-image-overlay/README.md
```diff
@@ -0,0 +1,37 @@
+# Animate images with image overlay
```
JonLavi, line 1: I think we recently discussed naming this "Animate images with image overlay**s**" so that it reads a bit smoother. Let's refer back to the common repo and see whether that change was made official.

scene/animate-images-with-image-overlay/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+![Image of animate images with image overlay](AnimateImagesWithImageOverlay.gif)
+
+## Use case
+
+An image overlay is useful for displaying fast and dynamic images; for example, rendering real-time sensor data captured from a drone. Each frame from the drone becomes a static image which is updated on the fly as the data is made available.
+
+## How to use the sample
+
+The application loads a map of the Southwestern United States. Click the "Start/Stop" button to start or stop the radar animation. Use the drop down menu to select how quickly the animation plays. Move the slider to change the opacity of the image overlay.
```
JonLavi, line 13: Perhaps we can just remove this first sentence? It's more general info / about the data rather than how to use the sample.

scene/animate-images-with-image-overlay/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+## How it works
+
+1. Create an `ImageOverlay` and add it to the `SceneView`.
+2. Set up a timer with an initial interval time of 67ms, equivalent to ~15fps.
```
JonLavi: Perhaps we can rephrase this sentence so that it's not so perscriptive, so that we don't accidentally imply that 15fps is the only way to do it? e.g.
`Set up a timer with your desired interval time. For example, an interval of 67ms is roughly equivalent to  15 fps.`

scene/animate-images-with-image-overlay/README.md
```diff
@@ -0,0 +1,37 @@
...
+* ImageOverlay
+* SceneView
+
+## About the data
+
+These radar images were captured by the US National Weather Service (NWS). They highlight the Pacific Southwest sector which is made up of part the western United States and Mexico. For more information visit the [National Weather Service](https://www.weather.gov/jetstream/gis) website.
+
+## Additional information
+
+The supported image formats are GeoTIFF, TIFF, JPEG, and PNG. `ImageOverlay` does not support the rich processing and rendering capabilities of a `RasterLayer`. Use `Raster` and `RasterLayer` for static image rendering, analysis, and persistence.
```
JonLavi, line 33: Perhaps we should add some information here on there being two ways of creating an ImageOverlay:
From a URL and from a JavaFX Image. For the latter, TIFF is not supported

scene/animate-images-with-image-overlay/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Tags
+
+3D, animation, drone, dynamic, image frame, image overlay, real time, rendering
```
JonLavi, line 37: Perhaps we should take the `drone` keyword out. It's a very specific use case, but there's no direct relevance to the API

scene/animate-images-with-image-overlay/README.metadata.json
```diff
@@ -0,0 +1,30 @@
...
+        "3D",
+        "animation",
+        "drone",
+        "dynamic",
+        "image frame",
+        "image overlay",
+        "real time",
+        "rendering"
+    ],
+    "redirect_from": "/java/latest/sample-code/animate-images-with-image-overlay.htm",
```
JonLavi: I think we're supposed to leave this entry out of the README.metadata.json, but I'm not sure any more. @mikewilburn, could you refresh our memories please?

scene/animate-images-with-image-overlay/README.metadata.json
```diff
@@ -0,0 +1,30 @@
...
+        "ImageFrame",
+        "ImageOverlay",
+        "SceneView"
+    ],
+    "snippets": [
+        "src/main/java/com/esri/samples/animate_images_with_image_overlay/AnimateImagesWithImageOverlaySample.java",
+      "src/main/java/com/esri/samples/animate_images_with_image_overlay/AnimateImagesWithImageOverlayController.java",
+      "src/main/resources/animate_images_with_image)overlay.fxml"
+    ],
+    "title": "Animate Images With Image Overlay Sample"
```
JonLavi: I don't think we need to add the suffix 'sample' here
```suggestion
    "title": "Animate Images With Image Overlay"
```

scene/animate-images-with-image-overlay/src/main/java/com/esri/samples/animate_images_with_image_overlay/AnimateImagesWithImageOverlayController.java
```diff
@@ -0,0 +1,208 @@
...
+import com.esri.arcgisruntime.mapping.ArcGISScene;
+import com.esri.arcgisruntime.mapping.ArcGISTiledElevationSource;
+import com.esri.arcgisruntime.mapping.Basemap;
+import com.esri.arcgisruntime.mapping.Surface;
+import com.esri.arcgisruntime.mapping.Viewpoint;
+import com.esri.arcgisruntime.mapping.view.Camera;
+import com.esri.arcgisruntime.mapping.view.ImageFrame;
+import com.esri.arcgisruntime.mapping.view.ImageOverlay;
+import com.esri.arcgisruntime.mapping.view.SceneView;
+import javafx.fxml.FXML;
```
JonLavi: I'm not 100% sure any more, but I think the standard order for the imports is:
```
java.*
<newline>
javafx.*
<newline>
com.esri.*
```

scene/animate-images-with-image-overlay/src/main/java/com/esri/samples/animate_images_with_image_overlay/AnimateImagesWithImageOverlayController.java
```diff
@@ -0,0 +1,208 @@
...
+public class AnimateImagesWithImageOverlayController {
+
+  @FXML
+  private SceneView sceneView;
+  @FXML
+  private Button controlAnimationButton;
+  @FXML
+  private Slider opacitySlider;
+  @FXML
+  private ComboBox<String> framesComboBox;
```
JonLavi: I think in most samples we didn't do the newline after the `@FXML` annotation. It's a bit annoying since the IntelliJ auto-foratting will muck this up again at some point.

```suggestion
  @FXML private SceneView sceneView;
  @FXML private Button controlAnimationButton;
  @FXML private Slider opacitySlider;
  @FXML private ComboBox<String> framesComboBox;
```

scene/animate-images-with-image-overlay/src/main/java/com/esri/samples/animate_images_with_image_overlay/AnimateImagesWithImageOverlayController.java
```diff
@@ -0,0 +1,208 @@
...
+
+  private List<ImageFrame> imageFrames;
+
+  private Integer imageIndex = 0;
```
JonLavi: We might not need this as a member variable, could just make it a var inside `addNextImageFrameToImageOverlay`?

scene/animate-images-with-image-overlay/src/main/java/com/esri/samples/animate_images_with_image_overlay/AnimateImagesWithImageOverlayController.java
```diff
@@ -0,0 +1,208 @@
...
+  private Integer period = 67;
+
+  private Timer timer = null;
+  private boolean isTimerRunning = true;
```
JonLavi: I think the samples convention was not to instantiate class members when we declare them, but do either in the `main` method, with a comment about what they are for, or ideally  just before you use them. I don't actually mind them being instantiated here, but best check what the common pattern is.

scene/animate-images-with-image-overlay/src/main/java/com/esri/samples/animate_images_with_image_overlay/AnimateImagesWithImageOverlayController.java
```diff
@@ -0,0 +1,208 @@
...
+      sceneView.setArcGISScene(scene);
+
+      // create a camera, looking at the pacific southwest sector
+      Point observationPoint = new Point(-116.621, 24.7773, 856977.0);
+      Camera camera = new Camera(observationPoint, 353.994, 48.5495, 0.0);
+
+      // create an envelope of the pacific southwest sector for displaying the image frame
+      Point pointForImageFrame = new Point(-120.0724273439448, 35.131016955536694, SpatialReferences.getWgs84());
+      Envelope imageFrameEnvelope = new Envelope(pointForImageFrame, 15.09589635986124, -14.3770441522488);
+      scene.setInitialViewpoint(new Viewpoint(imageFrameEnvelope, camera));
```
JonLavi, line 81: Can we just define an envelope straight away with four corner points, instead of `Evelope(center, width, height)`?

scene/animate-images-with-image-overlay/src/main/java/com/esri/samples/animate_images_with_image_overlay/AnimateImagesWithImageOverlayController.java
```diff
@@ -0,0 +1,208 @@
...
+    }
+  }
+
+  /**
+   * Create a new image frame from the image at the current index and add it to the image overlay.
+   */
+  private void addNextImageFrameToImageOverlay() {
+
+    // set image frame to image overlay
+    sceneView.getImageOverlays().get(0).setImageFrame(imageFrames.get(imageIndex));
```
JonLavi: There's only one ImageOverlay being used in this sample, so we can make it a class member instead of repeating `sceneView.getImageOverlays().get(0)` here and below?

scene/animate-images-with-image-overlay/src/main/java/com/esri/samples/animate_images_with_image_overlay/AnimateImagesWithImageOverlayController.java
```diff
@@ -0,0 +1,208 @@
...
+      File[] imageFiles = new File(System.getProperty("data.dir"), "./samples-data/PacificSouthWest").listFiles();
+      // sort the list of image files
+      if (imageFiles != null) {
+        Arrays.sort(imageFiles);
+        // create an image with the given path and use it to create an image frame
+        for (File file : imageFiles) {
+          ImageFrame imageFrame = new ImageFrame(file.getAbsolutePath(), imageFrameEnvelope);
+          imageFrames.add(imageFrame);
+        }
+      }
```
JonLavi: Some suggestions for the comments in this section. We should specify how we're sorting the file paths and ideall why (so that they display in order). Did you try without sorting? I would have thought the files are read in alphabetically when you do `.listFiles()`.
In the last section we're not actually creating an image in the intermediate step, we're using the `ImageFrame(URI, Envelope)` constructor, so we should specify that there. Perhaps it's worth adding a comment somewhere that we can do `ImageFrame(Image, Envelope)`

```suggestion
      // instantiate a new array to hold image frames
      imageFrames = new ArrayList<>();
      
      // get the image files from local storage
      File[] imageFiles = new File(System.getProperty("data.dir"), "./samples-data/PacificSouthWest").listFiles();
      
      // sort the list of image files by file name in ascending order
      if (imageFiles != null) {
        Arrays.sort(imageFiles);
        
        // create an image frame with the file path URI and add it to the array of image frames
        for (File file : imageFiles) {
          ImageFrame imageFrame = new ImageFrame(file.getAbsolutePath(), imageFrameEnvelope);
          imageFrames.add(imageFrame);
        }
      }
```

scene/animate-images-with-image-overlay/src/main/java/com/esri/samples/animate_images_with_image_overlay/AnimateImagesWithImageOverlayController.java
```diff
@@ -0,0 +1,208 @@
...
+
+      // populate the frames combo box with values
+      framesComboBox.getItems().addAll("60 frames per second", "30 frames per second", "15 frames per second");
+      // open the sample at 15fps
+      framesComboBox.getSelectionModel().select(2);
```
JonLavi: This is UI stuff so perhaps it's better to do it before we get to setting up the Scene/SceneView. It might be even possible to define all these things in the FXML file, might be worth having a look?
I also think it's counterintuitive to have the high values at the top rather than the bottom.

scene/animate-images-with-image-overlay/src/main/java/com/esri/samples/animate_images_with_image_overlay/AnimateImagesWithImageOverlayController.java
```diff
@@ -0,0 +1,208 @@
...
+   */
+  @FXML
+  private void changeImageOverlayOpacity() {
+    sceneView.getImageOverlays().get(0).setOpacity((float) opacitySlider.getValue());
+  }
+
+  /**
+   * Set up a timer to display the images at the specified frame rate from the combobox.
+   */
+  private void startNewAnimationTimer() {
```
JonLavi: This method is being called at the end of `initialize` so maybe we can put it directly after `initialize`, that way we're not making readers scroll past the two other methods first.

scene/animate-images-with-image-overlay/src/main/java/com/esri/samples/animate_images_with_image_overlay/AnimateImagesWithImageOverlayController.java
JonLavi: It might also make more sense to call it 'frameIndex', since we use it to pick an entry from `List<ImageFrame> imageFrames`, not from `imageFiles[]`

scene/animate-images-with-image-overlay/src/main/java/com/esri/samples/animate_images_with_image_overlay/AnimateImagesWithImageOverlayController.java
```diff
@@ -0,0 +1,208 @@
...
+      startNewAnimationTimer();
+
+    } catch (Exception e) {
+      // on any exception, print the stack trace
+      e.printStackTrace();
+    }
+  }
+
+  /**
+   * Create a new image frame from the image at the current index and add it to the image overlay.
```
JonLavi:
```suggestion
   * Sets the next image frame from the array of image frames to the image overlay.
```

scene/animate-images-with-image-overlay/src/main/java/com/esri/samples/animate_images_with_image_overlay/AnimateImagesWithImageOverlayController.java
```diff
@@ -0,0 +1,208 @@
...
+   */
+  private void addNextImageFrameToImageOverlay() {
```
JonLavi: It's not technically adding the next image frame, but repalcing the previous one, so maybe we should call it:
```suggestion
  private void setNextImageFrameToImageOverlay() {
```

scene/animate-images-with-image-overlay/src/main/java/com/esri/samples/animate_images_with_image_overlay/AnimateImagesWithImageOverlayController.java
```diff
@@ -0,0 +1,208 @@
...
+    if (imageIndex == imageFrames.size())
+      imageIndex = 0;
+  }
+
+  /**
+   * Controls the opacity of the image overlay using the slider.
+   */
+  @FXML
+  private void changeImageOverlayOpacity() {
+    sceneView.getImageOverlays().get(0).setOpacity((float) opacitySlider.getValue());
```
JonLavi: Same here, if you make a class member `ImageOverlay imageOverlay`, you can avoid interrogating the sceneView:
```suggestion
    imageOverlay.setOpacity((float) opacitySlider.getValue());
```

scene/animate-images-with-image-overlay/src/main/java/com/esri/samples/animate_images_with_image_overlay/AnimateImagesWithImageOverlayController.java
```diff
@@ -0,0 +1,208 @@
...
+    timer.scheduleAtFixedRate(timerTask, 1L, period);
+  }
+
+  /**
+   * Handles the rate at which the image frames are displayed using the combo box.
+   */
+  @FXML
+  private void handleframesComboBoxInteraction() {
+    // set the period for the chosen fps 
+    switch (framesComboBox.getSelectionModel().getSelectedItem()) {
```
JonLavi, line 147: If you wanted, you could add Key-Value pairs of String-Int to the ComboBox, and use a cell factory to display the key (15/30/60 fps), and then on selection just put in the corresponding value (17/33/67) straight into the `startNewAnimationTimer`. Let me know if you want me to walk you through it.

scene/animate-images-with-image-overlay/src/main/resources/animate_images_with_image_overlay.fxml
```diff
@@ -0,0 +1,58 @@
...
+
+            <Button fx:id="controlAnimationButton" text="Stop" maxWidth="500" GridPane.rowIndex="0" GridPane.columnIndex="0" onAction="#handlecontrolAnimationButtonClicked"/>
+            <ComboBox fx:id="framesComboBox" GridPane.rowIndex="0" GridPane.columnIndex="1" maxWidth="Infinity" onAction="#handleframesComboBoxInteraction"/>
+            <Label text="Opacity:" GridPane.rowIndex="1" GridPane.columnIndex="0"/>
+            <Slider fx:id="opacitySlider" GridPane.rowIndex="1" showTickLabels="true" majorTickUnit="50" GridPane.columnIndex="1" maxWidth="Infinity" max="1" value="1" onMouseDragged="#changeImageOverlayOpacity" onMouseReleased="#changeImageOverlayOpacity"/>
+            
+        </GridPane>
+    </VBox>
+    
+</StackPane>
```
JonLavi: GitHub is conveniently letting us know there's a newline missing here 😄

scene/animate-images-with-image-overlay/src/main/resources/animate_images_with_image_overlay.fxml
```diff
@@ -0,0 +1,58 @@
...
+            <Background>
+                <fills>
+                    <BackgroundFill>
+                        <fill>
+                            <Color opacity="0.3"/>
+                        </fill>
+                    </BackgroundFill>
+                </fills>
+            </Background>
+        </background>
```
JonLavi: Every time I see this in FXML files I cringe a bit 😞 
I've recently used CSS to get rid of some of this stuff and hide it in styles.css:
https://github.com/Esri/arcgis-runtime-samples-java/pull/498/files

Perhaps we can make that a regular thing with new samples? 


The trick is to add `styleClass="stack-pane"` to the StackPane, and then you can target individual FXML elements with rules, e.g. for a VBox:
```
.stack-pane VBox {
  -fx-text-fill: white;
    -fx-background-color: rgba(0, 0, 0, 0.3);
    -fx-padding: 10;
    -fx-max-width: 200;
    -fx-max-height: 120;
    -fx-spacing: 10;
}
``` 

It'll also allow you to hide away a lot of the `max width` etc. definitions.

JonLavi (commented):
Nice work @Rachael-E , fancy sample!
Just a few comments on JavaDoc, and a bit of code arrangement, but nothing major.
I think we could streamline the ComboBox functionality but I'll leave that up to you to implement or not. I also think we should start using the CSS files in a more focused way for samples have a lot of UI.

Lastly, could you record a gif without the mouse pointer present?

scene/animate-images-with-image-overlay/README.md
Rachael-E: I think it's ok to keep this in, it adds some context. We have other samples where we say the map is loading over such and such a place.

scene/animate-images-with-image-overlay/README.md
Rachael-E: I think this is probably more API ref than sample description territory. Whilst it's always nice to know the different options, we often use one of many constructors in a sample without laying out all the other options since it isn't directly used in the sample. There's also the possibility more constructors will be added in the future and we'd have an overhead of remembering to update sample doc.

scene/animate-images-with-image-overlay/README.md
Rachael-E: there is no direct relevance to the API, however I remember Jared saying it's one of the main use cases for this functionality, so we should leave this in.

scene/animate-images-with-image-overlay/src/main/java/com/esri/samples/animate_images_with_image_overlay/AnimateImagesWithImageOverlayController.java
Rachael-E: yes that's it, good spot!

scene/animate-images-with-image-overlay/src/main/java/com/esri/samples/animate_images_with_image_overlay/AnimateImagesWithImageOverlayController.java
Rachael-E: this has to be a member variable so that the count is tracked correctly: instantiating it as 0 inside the variable just makes the index set to 0 everytime. If I'm misunderstanding, could you let me know what you were thinking please?

scene/animate-images-with-image-overlay/src/main/java/com/esri/samples/animate_images_with_image_overlay/AnimateImagesWithImageOverlayController.java
Rachael-E: so the current implementation ensures the image frame and viewpoint sits exactly in the correct geographic location for the data provided. I could change it to an envelope with four corners if you thought there was a significant benefit in doing so over the existing implementation, though it ends up in the same lines of code plus a change in setting up the viewpoint logic. As always there's more than one way to skin a cat! I'm just not sure of the benefit of the 4 point envelope over the existing implementation.

scene/animate-images-with-image-overlay/README.md
JonLavi: I've never actually noticed that kind of info being added into 'how to use the sample'. I thought we were generally trying to put it in 'about the data'

scene/animate-images-with-image-overlay/src/main/java/com/esri/samples/animate_images_with_image_overlay/AnimateImagesWithImageOverlayController.java
JonLavi: I just find the constructur `Envelope(center, width, height)` incredibly unintuitive. We aren't extracting the geographic location from the images or any other data source, so there's not really much of a point in doing it that way.The way we are creating the envelope up there is essentially the same as adding/subtracting 15.09589635986124, -14.3770441522488 from the x/y of the centre-point, respectively. We could just write these values out so it's clearer where in the world these corner points are. But yeah, a matter of preference!

scene/animate-images-with-image-overlay/src/main/java/com/esri/samples/animate_images_with_image_overlay/AnimateImagesWithImageOverlayController.java
JonLavi: I got mixed up, I meant perhaps instantiate it in the `initialize` method, but really it makes sense where it is just now, so lets leave it as is.

scene/animate-images-with-image-overlay/src/main/java/com/esri/samples/animate_images_with_image_overlay/AnimateImagesWithImageOverlayController.java
Rachael-E: Great thanks for the comment improvement suggestions. And yep, so we do need to sort to be on the safe side - here's the doc from the `.listFiles()`:

> There is no guarantee that the name strings in the resulting array will appear in any specific order; they are not, in particular,guaranteed to appear in alphabetical order.

scene/animate-images-with-image-overlay/src/main/java/com/esri/samples/animate_images_with_image_overlay/AnimateImagesWithImageOverlayController.java
Rachael-E: I spent a while researching about how to define the above in the FXML but had no success :( if you know of a way could you let me know please?

scene/animate-images-with-image-overlay/README.md
Rachael-E: Mike just confirmed it's to stay as singular

scene/animate-images-with-image-overlay/src/main/java/com/esri/samples/animate_images_with_image_overlay/AnimateImagesWithImageOverlayController.java
Rachael-E: Thanks, I'd considered the use of a cell factory but then decided to go for the switch since it's a fairly basic set up and we can comment to show the conversion "maths". We have other Java samples that use the switch for similar "simplicity" so it seemed fitting. Though we also have cell factory for simple conversions too! Unless you have strong feelings about it we can leave as is.

JonLavi: I think the whole conversion stuff distracts from the actual sample's intentions, but if there's other samples that do this too then you can just leave it.

I think having this at the beginning of the sample and then using a cell factory is just as informative:
```
framesComboBox.getItems().addAll(
new ImmutablePair("60 frames per second", "17"), // 1000ms/17 = 60 fps
new ImmutablePair( "30 frames per second", "33"), // 1000ms/30 = 30 fps
new ImmutablePair( "15 frames per second", "67") // 1000ms/15 = 15 fps
);
```

scene/animate-images-with-image-overlay/src/main/java/com/esri/samples/animate_images_with_image_overlay/AnimateImagesWithImageOverlayController.java
JonLavi: I don't know whether this is possible off the top of my head, it was just a passing thought! If you couldn't find any way then I doubt I'll find anything either, so no worries leaving it like this

scene/animate-images-with-image-overlay/src/main/java/com/esri/samples/animate_images_with_image_overlay/AnimateImagesWithImageOverlayController.java
JonLavi: Good find!

Rachael-E:
@JonLavi thanks so much for the review! interesting about moving styling to css. Can you take another look at my changes please, for the CSS I'm not sure I've missed other bits that could move in there too. Thanks!

scene/animate-images-with-image-overlay/src/main/resources/animate_images_with_image_overlay.fxml
```diff
@@ -25,34 +25,17 @@
...
-                <fills>
-                    <BackgroundFill>
-                        <fill>
-                            <Color opacity="0.3"/>
-                        </fill>
-                    </BackgroundFill>
-                </fills>
-            </Background>
-        </background>
         <GridPane hgap="30" vgap="10" maxWidth="500">
```
JonLavi: Just found a couple of resource for CSS in fxml:
https://docs.oracle.com/javafx/2/layout/style_css.htm
https://docs.oracle.com/javafx/2/api/javafx/scene/doc-files/cssref.html#gridpane

You could get rid of the rules here by adding to the css:
```
Grid Pane {
	-fx-max-width: 500;
    -fx-hgap: 30;
    -fx-vgap: 10;
}
```

It looks like we could get away with removing the `styleClass="stack-pane"` and `styleClass="panel-region"` (there's no other stack-pane that we need to differentiate that one from in the fxml file). Might be worth trying to remove them and see how the CSS reacts.

I think moving the 'min/maxWidth' parameters to the CSS is definitely ok, it's more cosmetic rather than any direct relevance to the code. I think it'd be good to highlight the `fx:id` parts and the method calls (`onAction="#handleframesComboBoxInteraction"`), so let's bring those to the front of the node declarations.

Looks like you can also define the alignment within the StackPane in the CSS file:
https://docs.oracle.com/javafx/2/api/javafx/scene/doc-files/cssref.html#stackpane

You could even declare most of the params for the slider:
https://docs.oracle.com/javafx/2/api/javafx/scene/doc-files/cssref.html#slider

only thing I can't see in there is whether we can control the element's position inside the grid pane (i.e. replace `GridPane.rowIndex="0"` etc.)

I'll let you decide when too much is too much 😄

JonLavi (approved):
Looks good @Rachael-E. There's still some potential for hiding away some of the cosmetic stuff of the FXML in the CSS, but I'll leave that up to you!

scene/animate-images-with-image-overlay/src/main/resources/animate_images_with_image_overlay.fxml
Rachael-E: ok, I've done what I can with the css! thanks for the helpful links.

Rachael-E:
thanks @JonLavi I've had a go at the CSS. Have another look if you like! In the meanwhile I'll ask @tschie to do a second review please.

scene/animate-images-with-image-overlay/README.metadata.json
```diff
@@ -0,0 +1,30 @@
...
+        "ImageFrame",
+        "ImageOverlay",
+        "SceneView"
+    ],
+    "snippets": [
+        "src/main/java/com/esri/samples/animate_images_with_image_overlay/AnimateImagesWithImageOverlaySample.java",
+      "src/main/java/com/esri/samples/animate_images_with_image_overlay/AnimateImagesWithImageOverlayController.java",
+      "src/main/resources/animate_images_with_image)overlay.fxml"
+    ],
+    "title": "Animate Images With Image Overlay"
```
tschie:
```suggestion
    "title": "Animate Images with Image Overlay"
```

scene/animate-images-with-image-overlay/README.metadata.json
```diff
@@ -0,0 +1,30 @@
...
+        "rendering"
+    ],
+    "redirect_from": "",
+    "relevant_apis": [
+        "ImageFrame",
+        "ImageOverlay",
+        "SceneView"
+    ],
+    "snippets": [
+        "src/main/java/com/esri/samples/animate_images_with_image_overlay/AnimateImagesWithImageOverlaySample.java",
```
tschie:
```suggestion
      "src/main/java/com/esri/samples/animate_images_with_image_overlay/AnimateImagesWithImageOverlaySample.java",
```

scene/animate-images-with-image-overlay/README.metadata.json
```diff
@@ -0,0 +1,30 @@
...
+      "src/main/java/com/esri/samples/animate_images_with_image_overlay/AnimateImagesWithImageOverlayController.java",
+      "src/main/resources/animate_images_with_image)overlay.fxml"
```
tschie:
```suggestion
      "src/main/resources/animate_images_with_image_overlay.fxml"
```

scene/animate-images-with-image-overlay/src/main/java/com/esri/samples/animate_images_with_image_overlay/AnimateImagesWithImageOverlayController.java
```diff
@@ -0,0 +1,209 @@
...
+      File[] imageFiles = new File(System.getProperty("data.dir"), "./samples-data/PacificSouthWest").listFiles();
+      // sort the list of image files by file name in ascending order
+      if (imageFiles != null) {
+        Arrays.sort(imageFiles);
+        // create an image frame from the file path and add it to the list of image frames
+        for (File file : imageFiles) {
+          ImageFrame imageFrame = new ImageFrame(file.getAbsolutePath(), imageFrameEnvelope);
+          imageFrames.add(imageFrame);
+        }
+      }
```
tschie:
```suggestion
      
      // get the image files from local storage as an unordered list
      File[] imageFiles = new File(System.getProperty("data.dir"), "./samples-data/PacificSouthWest").listFiles();
      // sort the list of image files by file name in ascending order
      if (imageFiles != null) {
        imageFrames = Arrays.stream(imageFiles)
                .sorted()
                .map(f -> new ImageFrame(f.getAbsolutePath(), imageFrameEnvelope))
                .collect(Collectors.toList());
      }
```

tschie: Should also handle the case of missing data with an alert.

scene/animate-images-with-image-overlay/src/main/java/com/esri/samples/animate_images_with_image_overlay/AnimateImagesWithImageOverlayController.java
```diff
@@ -0,0 +1,209 @@
...
+      e.printStackTrace();
+    }
+  }
+
+  /**
+   * Set up a timer to display the images at the specified frame rate from the combobox.
+   */
+  private void startNewAnimationTimer() {
+
+    timer = new Timer(true);
```
tschie: I think the `Timeline` API we use in other samples would work well here:
```java
animation = new Timeline();
animation.setCycleCount(-1); // loop animation
animation.getKeyFrames().add(new KeyFrame(Duration.millis(Math.round(1000 / fps)), e -> {
  imageOverlay.setImageFrame(imageFrames.get(frameIndex));
  // update to the next frame
  frameIndex = (frameIndex + 1) % imageFrames.size();
}));
````

You only need to create it once when you create the frames and your start/stop button could then just do `animation.play()` and `animation.pause()`.

File: scene/animate-images-with-image-overlay/src/main/java/com/esri/samples/animate_images_with_image_overlay/AnimateImagesWithImageOverlayController.java [hunk 29]
```diff
@@ -0,0 +1,209 @@
...
+  @FXML
+  private void changeImageOverlayOpacity() {
+    imageOverlay.setOpacity((float) opacitySlider.getValue());
+  }
+
+  /**
+   * Handles the rate at which the image frames are displayed using the combo box.
+   */
+  @FXML
+  private void handleframesComboBoxInteraction() {
```
tschie:
```suggestion
  private void handleFramesComboBoxInteraction() {
```

scene/animate-images-with-image-overlay/src/main/java/com/esri/samples/animate_images_with_image_overlay/AnimateImagesWithImageOverlayController.java
```diff
@@ -0,0 +1,209 @@
...
+      timer.cancel();
+      startNewAnimationTimer();
+    }
+  }
+
+  /**
+   * Stops/starts the animation of the image frames on the image overlay.
+   */
+  @FXML
+  private void handlecontrolAnimationButtonClicked() {
```
tschie:
```suggestion
  private void handleControlAnimationButtonClicked() {
```

scene/animate-images-with-image-overlay/src/main/java/com/esri/samples/animate_images_with_image_overlay/AnimateImagesWithImageOverlayController.java
```diff
@@ -0,0 +1,209 @@
...
+      imageFrames = new ArrayList<>();
+      // get the image files from local storage as an unordered list
+      File[] imageFiles = new File(System.getProperty("data.dir"), "./samples-data/PacificSouthWest").listFiles();
+      // sort the list of image files by file name in ascending order
+      if (imageFiles != null) {
+        Arrays.sort(imageFiles);
+        // create an image frame from the file path and add it to the list of image frames
+        for (File file : imageFiles) {
+          ImageFrame imageFrame = new ImageFrame(file.getAbsolutePath(), imageFrameEnvelope);
+          imageFrames.add(imageFrame);
```
JonLavi: @Rachael-E regarding the performance aspect of this, have you tried holding on to an `ArrayList<Image>` and creating the `ImageFrame` ad-hoc, only when it needs to be displayed? There seems to be quite a difference in memory consumption between a collection of `Image` and `ImageFrame`, as other platforms are reporting

Rachael-E: thanks @JonLavi no haven't tried this: is this the agreed way forward from the samples team perspective? I've been out the loop :D just waiting on @tschie 2nd review and I can make changes then: I know we're already going to change the drop down text from "xx fps" to "slow" etc. If there's anything else to go in I'd be grateful to hear it!

JonLavi: There's no concensus as of yet for which way is to be chosen across all platforms. It looks like for some, an `ArrayList<Image>` provies a less smooth animation. I think it would be a matter of testing it out and seeing whether it runs smoothly still and consumes less memory. Hopefully the team can agree to add some info to the 'Additional Information' section of the Readme - it's still in PR.

tschie (changes requested):

Rachael-E:
@JonLavi I've gone through Tyler's comments, and also chatted with Colin about `List<Image>` vs `List<ImageFrame>`. Seeing as the difference isn't huge and both run smoothly, we agreed to stick with ImageFrame to preload them all first. Could you take another look over please before I merge? Thanks!

scene/animate-images-with-image-overlay/src/main/java/com/esri/samples/animate_images_with_image_overlay/AnimateImagesWithImageOverlayController.java
```diff
@@ -120,34 +115,22 @@ public void initialize() {
...
-      }
-    };
-    timer.scheduleAtFixedRate(timerTask, 1L, period);
+    animation = new Timeline();
+    animation.setCycleCount(-1); // loop animation
+    animation.getKeyFrames().add(new KeyFrame(Duration.millis(period), e -> {
+      // set image frame to image overlay
+      imageOverlay.setImageFrame(imageFrames.get(frameIndex));
+      // update to the next frame
+      frameIndex = (frameIndex + 1) % imageFrames.size();
```
JonLavi, line 128: I'm surprised this can't be done with just `frameIndex++`. The modulo operation looks a bit funky 😄

scene/animate-images-with-image-overlay/README.md
```diff
@@ -0,0 +1,37 @@
...
+An image overlay is useful for displaying fast and dynamic images; for example, rendering real-time sensor data captured from a drone. Each frame from the drone becomes a static image which is updated on the fly as the data is made available.
+
+## How to use the sample
+
+The application loads a map of the Southwestern United States. Click the "Start/Stop" button to start or stop the radar animation. Use the drop down menu to select how quickly the animation plays. Move the slider to change the opacity of the image overlay.
+
+## How it works
+
+1. Create an `ImageOverlay` and add it to the `SceneView`.
+2. Set up a timer with an interval period.
```
JonLavi: We're technically now using the `Timeline` class so may need to find better wording for this?

scene/animate-images-with-image-overlay/README.metadata.json
```diff
@@ -0,0 +1,33 @@
+{
+    "category": "Scene",
+    "description": "Animate a series of images with an image overlay.",
+    "ignore": false,
+    "images": [
+        "AnimateImageWithImageOverlay.gif"
```
JonLavi:
```suggestion
        "AnimateImagesWithImageOverlay.gif"
```

scene/animate-images-with-image-overlay/src/main/java/com/esri/samples/animate_images_with_image_overlay/AnimateImagesWithImageOverlayController.java
```diff
@@ -0,0 +1,189 @@
...
+    switch (framesComboBox.getSelectionModel().getSelectedItem()) {
+      case "Fast":
+        period = 17;
+        break;
+      case "Medium":
+        period = 33;
+        break;
+      case "Slow":
+        period = 67;
+        break;
```
JonLavi, line 156: we should really have a default case in this `switch`, you could just put it on 'slow'?

scene/animate-images-with-image-overlay/src/main/java/com/esri/samples/animate_images_with_image_overlay/AnimateImagesWithImageOverlayController.java
```diff
@@ -0,0 +1,189 @@
...
+import com.esri.arcgisruntime.mapping.view.SceneView;
+
+public class AnimateImagesWithImageOverlayController {
+
+  @FXML private SceneView sceneView;
+  @FXML private Button controlAnimationButton;
+  @FXML private Slider opacitySlider;
+  @FXML private ComboBox<String> framesComboBox;
+
+  private List<ImageFrame> imageFrames;
```
JonLavi, line 54: In theory this could be a local var. Are we worried about GC being a problem here?

scene/animate-images-with-image-overlay/src/main/resources/animate_images_with_image_overlay/style.css
```diff
@@ -0,0 +1,37 @@
...
+    -fx-show-tick-labels: true;
+    -fx-major-tick-unit: 50;
+} .axis {
+    -fx-tick-label-fill: white;}
+
+Label {
+    -fx-text-fill: white;
+}
+
+
```
JonLavi: lots of blank space here that can go

scene/animate-images-with-image-overlay/src/main/resources/animate_images_with_image_overlay/style.css
```diff
@@ -0,0 +1,37 @@
...
+Button {
+    -fx-width: 500;
+}
+
+Slider {
+    -fx-show-tick-labels: true;
+    -fx-major-tick-unit: 50;
+} .axis {
+    -fx-tick-label-fill: white;}
+
```
JonLavi:
```suggestion
}

.axis {
    -fx-tick-label-fill: white;
}

```

scene/animate-images-with-image-overlay/src/main/resources/animate_images_with_image_overlay/main.fxml
```diff
@@ -0,0 +1,40 @@
...
+<?import javafx.scene.control.Slider?>
+<?import javafx.scene.control.Button?>
+<?import javafx.scene.control.ComboBox?>
+<?import javafx.scene.control.Label?>
+<StackPane fx:controller="com.esri.samples.animate_images_with_image_overlay.AnimateImagesWithImageOverlayController"
+           xmlns:fx="http://javafx.com/fxml" stylesheets="/animate_images_with_image_overlay/style.css">
+    <SceneView fx:id="sceneView"/>
+    <VBox>
+    <StackPane.margin>
+        <Insets bottom="20"/>
```
JonLavi, line 31: I think you can squeeze this parameter into the CSS as well

JonLavi (approved):
Nice work @Rachael-E . There's just a few things to tidy and a couple of questions, but nothing big to add

scene/animate-images-with-image-overlay/src/main/java/com/esri/samples/animate_images_with_image_overlay/AnimateImagesWithImageOverlayController.java
Rachael-E: this won't work because it would just count up infinitely: going beyond the total size of the array which is `240`. The modulo here keeps the logic concise and neat, otherwise we'd have to do as I had before which adds more codelines: 

```
    frameIndex++;	
    // reset index once all files have been loaded	
    if (frameIndex == imageFrames.size())	
      frameIndex = 0;
```

scene/animate-images-with-image-overlay/src/main/java/com/esri/samples/animate_images_with_image_overlay/AnimateImagesWithImageOverlayController.java
Rachael-E: I think because there's no other way any other value can be chosen in the combobox, we don't need a default here: it can only be one of "fast", "medium" or "slow".  In the code at the beginning we select index 2 from the combo box, and don't set that to anything else in the code (it would throw an error if we tried to set it to something beyond that index).

scene/animate-images-with-image-overlay/src/main/java/com/esri/samples/animate_images_with_image_overlay/AnimateImagesWithImageOverlayController.java
JonLavi: True, but at the cost of 2 extra lines of code and a comment, that is much more easy to comprehend

PR #511, opened 2020-05-05
Garbage collection adjustments

Rachael-E:
As per @mbcoder's feedback from a customer, this is a PR to remove a garbage collection issue with `Loadable`s. 

This PR moves `Loadable` local variable instances to be class variables so as to avoid garbage collection.


Rachael-E:
@colinanderson, please would you be able to take a look at the not so simple conversions here: these are for cases in: 

1. GroupLayerSample (where I've made `GroupLayer` a member variable rather than the childLayer from that in the `forEach` loop, line 97)
2. PictureMarkerSymbolSample (where I've made each PictureMarkerSymbol which is passed into a method a member variable: wasn't sure if there was a neater way to handle this within the method `placePictureMarkerSymbol` itself, which is where the addDoneLoadingListener occurs line 137)
3. DownloadPreplannedMapController (where I've made `List<PreplannedMapArea> preplannedMapAreas` the member variable rather than each preplanned map area within that list - each preplanned map area has an add done loading listener on line 111)
4. OpenMapURLSample (similar to above, only with a list of webmaps: each webmap has an add done loading listener on line 77)

Please could you also sanity check that it's ok I've made `ArcGISMap` a class member in samples where it has an addDoneLoadingListener: I assume this is needed in the same way a Loadable is unless something under the hood is handled differently.

colinanderson (commented):
@Rachael-E I've pretty much just looked over the samples you indicated so I won't approve the full PR but those ones look ok. A couple of comments...

PictureMarkerSymbolSample - I think you have an unnecessary `Platform.runLater`here
https://github.com/Esri/arcgis-runtime-samples-java/pull/511/files#diff-bb6c7349fa30c092c7d962ec363d2f34R109 as you have added the listener on the UI thread so it should run on the UI thread.

Random one - https://github.com/Esri/arcgis-runtime-samples-java/pull/511/files#diff-bb3c53e60f4708181eda81dc21c60a67R127. I think `facilitiesFeatureLayer` and `incidentsFeatureLayer` avoid garbage collection by being added to the operational layers so don't need to be class members. I think that might apply in other samples where you create a layer then add it to the operational layers.

Rachael-E:
thanks @colinanderson! Good point about the loadables that are in operational layers, and therefore attached to the map view which shouldn't lose scope anyway. I've reverted those changes and also removed the `Platform.runLater` you pointed out, it seems to work just as well without that.

@JonLavi could you take a look through these updated changes please?

analysis/viewshed-geoprocessing/src/main/java/com/esri/samples/viewshed_geoprocessing/ViewshedGeoprocessingSample.java
```diff
@@ -103,7 +106,7 @@ public void start(Stage stage) {
       progress.setMaxWidth(30);
 
       // create the geoprocessing task with the service URL and load it
-      GeoprocessingTask geoprocessingTask = new GeoprocessingTask("https://sampleserver6.arcgisonline" +
+      geoprocessingTask = new GeoprocessingTask("https://sampleserver6.arcgisonline" +
```
JonLavi, line 109: Not entirely related: do we actually have a character limit for lines? 
I think in some samples we're just keeping the URL on one long line. Would be nice to keep it consistent :)

display_information/update-graphics/src/main/java/com/esri/samples/update_graphics/UpdateGraphicsSample.java
```diff
@@ -152,7 +153,7 @@ public void start(Stage stage) {
       selectedGraphic = new Graphic();
 
       // create a ArcGISMap with basemap light gray canvas
-      ArcGISMap map = new ArcGISMap(Basemap.Type.LIGHT_GRAY_CANVAS, 56.075844, -2.681572, 13);
+      map = new ArcGISMap(Basemap.Type.LIGHT_GRAY_CANVAS, 56.075844, -2.681572, 13);
 
       // enable buttons when map view is done loading
       map.addDoneLoadingListener(() -> {
```
JonLavi, line 159: Just for my own understanding: we need `map` as a member variable even if we're not using that object within it's own `DoneLoadingListener`? 
I'm asking because in other cases we usually use the object itself inside the listener, but we don't do that here. The only actions are on the UI elements.

In other words: if we add a `DoneLoadingListener` to an Object, that Object needs to be a member var?

editing/update-geometries/src/main/java/com/esri/samples/update_geometries/UpdateGeometriesSample.java
```diff
@@ -112,7 +113,7 @@ public void start(Stage stage) {
                     Iterator<Feature> features = selectedQueryResult.iterator();
                     if (features.hasNext()) {
                       // move selected feature to clicked location
-                      ArcGISFeature selected = (ArcGISFeature) features.next();
+                      selected = (ArcGISFeature) features.next();
```
JonLavi: Could we use this opportunity to rename the variable to `selectedFeature`? We've generally been doing that in samples, and I find it's clearer for users (although a lot more verbose).

image_layers/change-sublayer-renderer/src/main/java/com/esri/samples/change_sublayer_renderer/ChangeSublayerRendererSample.java
```diff
@@ -42,9 +42,10 @@
 import com.esri.arcgisruntime.symbology.SimpleLineSymbol;
 
 public class ChangeSublayerRendererSample extends Application {
-
-  private MapView mapView;
+
```
JonLavi: Two sneaky spaces here are not needed
```suggestion

```

kml/display-kml/src/main/java/com/esri/samples/display_kml/DisplayKMLSample.java
```diff
@@ -42,7 +42,7 @@
 public class DisplayKMLSample extends Application {
 
   private MapView mapView;
-
+
```
JonLavi: More sneaky spaces
```suggestion

```

map_view/identify-layers/src/main/java/com/esri/samples/identify_layers/IdentifyLayersSample.java
```diff
@@ -42,8 +42,11 @@
 import com.esri.arcgisruntime.mapping.view.MapView;
 
 public class IdentifyLayersSample extends Application {
-
+
```
JonLavi:
```suggestion

```

symbology/picture-marker-symbol/src/main/java/com/esri/samples/picture_marker_symbol/PictureMarkerSymbolSample.java
```diff
@@ -46,6 +45,12 @@
   private File orangeSymbolPath;
   private GraphicsOverlay graphicsOverlay;
 
+  // keeps loadables in scope to avoid garbage collection
+  private ArcGISMap map;
+  private PictureMarkerSymbol campsiteSymbol;
```
JonLavi, line 50: I've just learnt that PictureMarkerSymbol are Loadable! Every day is a school day!
Are they affected by this problem as well, even though we don't add a `DoneLoadingListener`?

JonLavi (approved):
Nice work @Rachael-E, the changes look pretty straightforward.
Only suggestion would be perhaps to phrase the comment by the member variables in active voice?
`  // keep loadables in scope to avoid garbage collection`
as opposed to `keeps`?
You're the documentation pro so I'll leave that to you to choose 😄 

Other than that a couple of sneaky formatting suggestions and some questions for my understanding

analysis/viewshed-geoprocessing/src/main/java/com/esri/samples/viewshed_geoprocessing/ViewshedGeoprocessingSample.java
Rachael-E: the character limit should be roughly 120 (if you haven't got it already there's a setting in IntelliJ you can use to show you the limit (File -> Settings -> Editor -> Code Style -> Hard wrap at: 120). This would be a good thing for us to doc in a Java sample contribution guide!  For this particular example,bit's way off the 120 line even when moving "new GeoprocessingTask" to a new line. So for examples like this it's fine to split for readability.

JonLavi: Yep, I have this set up already. However the Java API repo and the samples repo don't necessarily have the same code style convention. Was just trying to point out the inconsistency I've been seeing between samples

display_information/update-graphics/src/main/java/com/esri/samples/update_graphics/UpdateGraphicsSample.java
Rachael-E: @colinanderson could you comment on this please? I think I was under the impression we would still need it so that there isn't potentially a reference lost for the addDoneLoadingListener, but perhaps it is just for cases when the loadable is called again within the done loading listener.

editing/update-geometries/src/main/java/com/esri/samples/update_geometries/UpdateGeometriesSample.java
Rachael-E: sure!

symbology/picture-marker-symbol/src/main/java/com/esri/samples/picture_marker_symbol/PictureMarkerSymbolSample.java
Rachael-E: yes, an `addDoneLoadingListener` is used later in the method `placePictureMarkerSymbol` where a picture marker symbol is a parameter :)

JonLavi: Woops, I obviously completely missed that! Trying to do too many things at once, and failing a bit.

display_information/update-graphics/src/main/java/com/esri/samples/update_graphics/UpdateGraphicsSample.java
colinanderson: >Just for my own understanding: we need map as a member variable even if we're not using that object within it's own DoneLoadingListener?

@Rachael-E @JonLavi 
Yes, the garbage collection issue is that the object  can get garbage collected because it is unreferenced and the done listener never runs. Referencing the loadable inside the done listener doesn't stop it being garbage collected - I assume it becomes a circular reference to a local variable and the variable isn't referenced anymore in the method so can be garbage collected. I'd have to do some revision to get the exact facts though!

Rachael-E: great, thanks Colin!

Rachael-E:
thanks @JonLavi! I've made those changes. Could you take another look please?

JonLavi (approved):
Looks great @Rachael-E !

Rachael-E:
@mbcoder could you do a second review please?

map/download-preplanned-map/src/main/java/com/esri/samples/download_preplanned_map/DownloadPreplannedMapController.java
```diff
@@ -101,7 +102,7 @@ private void initialize() {
       preplannedMapAreasFuture.addDoneListener(() -> {
         try {
           // get the preplanned areas and add them to the list view
-          List<PreplannedMapArea> preplannedMapAreas = preplannedMapAreasFuture.get();
+          preplannedMapAreas = preplannedMapAreasFuture.get();
           preplannedAreasListView.getItems().addAll(preplannedMapAreas);
```
mbcoder, line 106: @Rachael-E this looks like a listenable future which won't get garbage collected.

mbcoder (changes requested):

mbcoder (approved):

PR #512, opened 2020-05-11
New Sample: Identify raster cell

JonLavi:
Another fun and interactive sample.
I think the logic for freezing/unlocking the callout worked out ok, but let me know if you have any other ideas.


raster/identify-raster-cell/README.md
```diff
@@ -0,0 +1,37 @@
+# Identify raster cell
+
+Get the cell value of a local raster at the clicked location and display the result in a callout.
```
Rachael-E: was this what was finally agreed upon in the design? we don't usually put UI interactions in here such as "at the clicked location", since this is more a description of how to use the sample (and the main way to use the sample is to hover with the mouse anyway). How about just 
```suggestion
Get the cell value of a local raster and display the result in a callout.
```

raster/identify-raster-cell/src/main/java/com/esri/samples/identify_raster_cell/IdentifyRasterCellSample.java
```diff
@@ -0,0 +1,220 @@
...
+            stage.setWidth(800);
+            stage.setHeight(700);
+            stage.setScene(scene);
+            stage.show();
+
+            // create a map view
+            mapView = new MapView();
+
+            // create a Map with an oceans basemap
+            ArcGISMap map = new ArcGISMap(Basemap.Type.OCEANS, -33.9, 18.6, 9);
```
Rachael-E: since the raster layer extents are being used to set the map viewpoint once it's loaded, they are overriding the settings here. I think we should have the level of detail a little closer, but with the raster layer changing the map viewpoint, we should handle that change there rather than here (since this is overwritten).  For here we can just do:
```suggestion
            ArcGISMap map = new ArcGISMap(Basemap.createOceans());
```

raster/identify-raster-cell/src/main/java/com/esri/samples/identify_raster_cell/IdentifyRasterCellSample.java
```diff
@@ -0,0 +1,220 @@
...
+            // create a raster layer
+            rasterLayer = new RasterLayer(raster);
+
+            // add the raster as an operational layer
+            map.getOperationalLayers().add(rasterLayer);
+
+            // set viewpoint on the raster
+            rasterLayer.addDoneLoadingListener(() -> {
+                if (map.getLoadStatus() == LoadStatus.LOADED) {
+                    mapView.setViewpointGeometryAsync(rasterLayer.getFullExtent(), 150);
```
Rachael-E: I found the map was too far zoomed out: the focus and majority of the map should be on the raster layer. I suggest a little closer using a padding of 50 instead of 150.
```suggestion
                    mapView.setViewpointGeometryAsync(rasterLayer.getFullExtent(), 50);
```

raster/identify-raster-cell/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+![Image of identify raster cell](IdentifyRasterCell.png)
+
+## Use case
+
+You may want to identify a raster layer to get its exact cell value in the case the approximate value conveyed by its symbology is not sufficient. The information available for the raster cell depends on the type of raster layer being identified. For example, a 3-band satellite or aerial image might provide 8-bit RGB values, whereas a digital elevation model (DEM) would provide floating point z values. By identifying a raster cell of a DEM, you can retrieve the precise elevation of a location.
+
+## How to use the sample
+
+Move the mouse pointer over an area of the raster to identify the raster cell at that location, it and see the cell attributes information displayed in a callout. You can click the prmary mouse button to lock the callout in place, and click again to release the callout and resume identifying on-the-fly.
```
Rachael-E: some missing words here, I've put it back together along with fixing a wee typo. 
```suggestion
Move the mouse pointer over an area of the raster to identify the raster cell at that location. The raster cell attribute information will display in a callout. You can click the primary mouse button to lock the callout in place, and click again to release the callout and resume identifying on-the-fly.
```

raster/identify-raster-cell/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+## How it works
+
+1. Add a listener to the `MapView` to capture mouse clicks.
+2. On click:
+  * Dismiss the `Callout`, if one is showing.
+  * Call `identifyLayerAsync(...)` passing in the layer, screen point, tolerance, whether to return popups only, and maximum number of results.
```
Rachael-E: kinda think we should specify it's a raster layer here
```suggestion
  * Call `identifyLayerAsync(...)` passing in the raster layer, screen point, tolerance, whether to return popups only, and maximum number of results.
```

raster/identify-raster-cell/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Relevant API
+
+* IdentifyLayerResult
+* RasterCell
+* RasterLayer
+
+## About the data
+
+The data shown is an NDVI classification derived from MODIS imagery between 27 Apr 2020 and 4 May 2020. It comes from the [NASA Worldview application](https://worldview.earthdata.nasa.gov/). In a normalized difference vegetation index, or [NDVI](https://en.wikipedia.org/wiki/Normalized_difference_vegetation_index), values range between -1 and +1 with the positive end of the spectrum showing green vegetation.
```
Rachael-E, line 33: I've noticed that none of the data goes below 0.0 in the sample, and with the description as is, I'd be expecting the data to show values between -1 and +1. It seems the description is a generalisation, but I think it would be good to tweak it slightly to more accurately reflect the values in the sample. However I noticed the imagery/data we're using is described as EVI (Enhanced Vegetation Index) in its file name (not as NDVI). I can't remember who authored the data here but it would be worth to double check with them as we may need to change the description to EVI. 

I think a better description could be the following since it matches the data value ranges shown and isn't a general description (taken from https://www.sentinel-hub.com/eoproducts/ndvi-normalized-difference-vegetation-index), though this is assumes it is NDVI rather than EVI:

"Low, positive values represent shrub and grassland (approximately 0.2 to 0.4), while high values indicate temperate and tropical rainforests (values approaching 1)."

raster/identify-raster-cell/README.metadata.json
```diff
@@ -0,0 +1,29 @@
+{
+    "category": "Raster",
+    "description": "\nGet the cell value of a local raster at the clicked location and display the result in a callout.",
```
Rachael-E: I don't think we usually show the new paragraph notation here?
```suggestion
    "description": "Get the cell value of a local raster at the clicked location and display the result in a callout.",
```

raster/identify-raster-cell/src/main/java/com/esri/samples/identify_raster_cell/IdentifyRasterCellSample.java
```diff
@@ -0,0 +1,220 @@
...
+                } else {
+                    Alert alert = new Alert(Alert.AlertType.ERROR, "Raster Layer Failed to Load!");
+                    alert.show();
+                }
+            });
+
+            // get a handle on the callout
+            callout = mapView.getCallout();
+            callout.setMouseTransparent(true);
```
Rachael-E, line 98: I'm not too sure what this line of code does? I was expecting the mouse arrow icon to become transparent or for something to become to transparent but when I comment out the code I see no change?

raster/identify-raster-cell/src/main/java/com/esri/samples/identify_raster_cell/IdentifyRasterCellSample.java
```diff
@@ -0,0 +1,220 @@
...
+        // identify the layers at the clicked location
+        ListenableFuture<IdentifyLayerResult> identifyLayerResultFuture
+                = mapView.identifyLayerAsync(rasterLayer, point, 10, false, 1);
+
+        identifyLayerResultFuture.addDoneListener(() -> {
+            try {
+                // get the result of the query
+                IdentifyLayerResult identifyLayerResult = identifyLayerResultFuture.get();
+
+                // get the read only list of geo-elements (they contain RasterCell's)
```
Rachael-E:
```suggestion
                // get the read only list of geo-elements (they contain RasterCells)
```

Rachael-E (commented):
cool @JonLavi really nice sample! Fun to see the logic the mouse interacting with the mapview. Nothing major, a few typos to fix up, some tweaks to the map constructor and a query about the "About the data" section which would be good to clarify.

raster/identify-raster-cell/src/main/java/com/esri/samples/identify_raster_cell/IdentifyRasterCellSample.java
JonLavi: It basically means that hovering over the callout makes the app act as if we're hovering over the map. In other words, the callout is transparent to the mouse interaction. 
What it does is basically allow the on-the-fly identification to keep working if we move the mouse over the callout. Without it the app would stop identifying because we're technically hovering on the callout, and not on the map.
I'll add a better comment!

Rachael-E: ahhh ok I get it now, thanks for explaining!

raster/identify-raster-cell/README.md
JonLavi: Thanks for catching this!

raster/identify-raster-cell/src/main/java/com/esri/samples/identify_raster_cell/IdentifyRasterCellSample.java
JonLavi: I've added a comment that hopefully explains this now. Let me know what you think?

raster/identify-raster-cell/README.md
JonLavi: @TADraeseke I think you added this data? Any thoughts on Rachaels suggestions?

TADraeseke: Hi @Rachael-E @JonLavi - Yes, I authored the data and struggled to get it to the normalized -1 to 1. My plan is to update the data at some point soon and push up the new data to the existing item on AGOL.

So I'd recommend proceeding as is for now, as the data will be corrected soon entirely on the back end :)

JonLavi: Thanks @TADraeske! We'll leave it as is for now then.

JonLavi:
@Rachael-E would you mind having a second look after the changes?

raster/identify-raster-cell/README.metadata.json
```diff
@@ -1,6 +1,6 @@
 {
     "category": "Raster",
-    "description": "\nGet the cell value of a local raster at the clicked location and display the result in a callout.",
+    "description": "Get the cell value of a local raster at the clicked location and display the result in a callout.",
```
Rachael-E: this needs to be updated to match the new description 
```suggestion
    "description": "Get the cell value of a local raster and display the result in a callout.",
```

Rachael-E (approved):
thanks @JonLavi , looks good, just one small change to be made and looks good to me.

raster/identify-raster-cell/README.metadata.json
JonLavi: Good memory, totally forgot about updating this!

JonLavi:
Thanks for the review @Rachael-E. 
@tschie should be ready for your review now. Thanks!

JonLavi:
Hi @mbcoder, would you mind reviewing when you have some time?

raster/identify-raster-cell/build.gradle
```diff
@@ -0,0 +1,77 @@
...
+
+group = 'com.esri.samples'
+
+ext {
+    arcgisVersion = '100.8.0'
+}
+
+javafx {
+    version = "11.0.1"
+    modules = [ 'javafx.controls', 'javafx.fxml' ]
```
mbcoder: to be sure you should add all modules.  If anyone adds functionality to the app it will break.  We should add all modules to all samples by default.

```
    modules = [ 'javafx.controls',
                'javafx.base',
                'javafx.graphics',
                'javafx.fxml',
                'javafx.web',
                'javafx.media']

```

raster/identify-raster-cell/build.gradle
```diff
@@ -0,0 +1,77 @@
...
+}
+
+group = 'com.esri.samples'
+
+ext {
+    arcgisVersion = '100.8.0'
+}
+
+javafx {
+    version = "11.0.1"
```
mbcoder: we should be using 11.0.2 which is the latest public version

mbcoder (changes requested):
Looks okay, just a couple of points with the build script

JonLavi:
@mbcoder @Rachael-E since the discussion is ongoing, I've reverted the build.gradle to only include the relevant javafx modules.
@mbcoder , would you please do another review so we could merge this sample?

mbcoder (approved):
Approved for merge - We can address gradle issues in another pull request

JonLavi:
Thanks both!

PR #515, opened 2020-06-01
Feature collection layer from portal - Title to sentence case

JonLavi:
Just had a look at this sample and noticed the title in the readme is not sentence case.
A quick PR to fix that

### DIFF
--- a/feature_layers/feature-collection-layer-from-portal/README.md
+++ b/feature_layers/feature-collection-layer-from-portal/README.md
@@ -1,4 +1,4 @@
-# Feature Collection Layer From Portal
+# Feature collection layer from portal
 
 Create a feature collection layer from a portal item.
 


feature_layers/feature-collection-layer-from-portal/README.md
```diff
@@ -1,4 +1,4 @@
-# Feature Collection Layer From Portal
+# Feature collection layer from portal
```
Rachael-E, line 1: I've noticed there's a mixture of cases throughout this readme when referring to Portal and it made me wonder what the correct case should be. Since Portal is an ArcGIS product, it would make sense to me that it remains capitalized but would definitely be worth checking with others, since the case also switches on the ArcGIS Portal help pages!

JonLavi: I would tend to agree but I think the standard is to only capitalize when directly referring to an API class in the programming context, but not when talking about the concept of a portal

Rachael-E: yup I agree above, though as I understand it in this instance the reference to Portal isn't an API class but an actual ArcGIS product https://enterprise.arcgis.com/en/portal/

it seems to refer to itself as lower case mostly though in the linked pages.

Rachael-E (approved):

JonLavi:
Thanks @Rachael-E! 
Merging as this is quite a trivial change.

PR #516, opened 2020-06-04
Fix up image references for displaying on webpages

Rachael-E:
This PR contains a number of changes to ensure our sample images display correctly on our website. 

A number of samples contained differing references to the file type of an image between its README and metadata.json file. The metadata image type needs to be identical to the README image type so that the build copies over the image referenced from the README. This wasn't happening when e.g. the README referenced a .gif but the metadata referenced a .png. The website couldn't access a copied .png since that's not referenced in the README. This PR fixes up these discrepancies (updating the metadata to use the file type used in the README for consistency with the sample viewer), along with fixing incorrectly named image paths. 

@JonLavi please could do a first review?

### DIFF
--- a/analysis/line-of-sight-geoelement/README.metadata.json
+++ b/analysis/line-of-sight-geoelement/README.metadata.json
@@ -3,7 +3,7 @@
   "description": "Show a line of sight between two moving objects.",
   "ignore": false,
   "images": [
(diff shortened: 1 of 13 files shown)


JonLavi (approved):
Good catch on all the mis-named screenshot files 👍

PR #518, opened 2020-06-05
Move resources

tschie:
Moves every sample's resources into a namespaced folder. For example: src/main/resources/**viewshed_location**/style.css.

- renamed all root FXML files to main.fxml since they no longer need to be unique
- fixed some styles
- added missing fxml files from snippet list

I checked every sample in the sample viewer to confirm that everything compiles, has no obvious style issues, and lists all fxml files in the code section.


analysis/viewshed-location/README.metadata.json
```diff
@@ -23,7 +23,7 @@
   "snippets": [
     "src/main/java/com/esri/samples/viewshed_location/ViewshedLocationSample.java",
     "src/main/java/com/esri/samples/viewshed_location/ViewshedLocationController.java",
-    "src/main/resources/viewshed_location.fxml"
+    "src/main/resources/viewshed_location/main.fxml"
   ],
   "title": "Viewshed Location"
 }
```
JonLavi, line 29: Newline is missing here, just fyi

JonLavi (approved):
Nice work @tschie . Only spotted one missing newline, but looks like that was there previously.

PR #519, opened 2020-06-09
Nearest Vertex / Convex Hull - normalise point for wrapping around date line

JonLavi:
A quick modification to nearest vertex / convex hull, where panning across the dateline results in unexpected behaviour. This change introduces normalizing the clicked map points.
There's a similar workflow in `AddFeatures`.


Rachael-E (approved):

JonLavi:
Thanks @Rachael-E !
@tschie would you mind doing a review as well?

PR #521, opened 2020-07-07
Fix missing redirect URL

jf990:
`redirect_from` should not be an array of empty strings. if using array, all strings must be valid URLs.


mikewilburn:
This is a duplicate of #522 which has just been merged. @jf990, I'll close this one.

PR #523, opened 2020-07-08
Correct redirect_from arrays

jf990:
I missed a few. we need this merged asap because this breaks the entire developer site build.


jf990:
@JonLavi @mikewilburn this is critical, please review ASAP as its holding up the site build.

mikewilburn (approved):
Changes LGTM. Thanks for finding and correcting these mistakes.

Since @JonLavi and @Rachael-E are in a different timezone, I'll go ahead and merge this, but circle back with them separately to follow up on this.

PR #524, opened 2020-07-08
Another metadata array fix

jf990:
I missed another one from the prior PR.


jf990:
@mikewilburn

mikewilburn (approved):
LGTM. We'll additionally look out for this pattern - empty strings within arrays broken out over several lines - going forward.

PR #525, opened 2020-07-10
Fix BufferList sample bug

Rachael-E:
Moves the `setInitialViewpoint` logic before setting the `mapView` to the `ArcGISMap`: if the map initial viewpoint logic is ordered after setting the map view to the map, it won't execute. Also updated the comments to be clearer. @JonLavi please could you take a look and test to make sure this works ok? Thanks!


JonLavi (approved):
Good find @Rachael-E .
This works on the SV 100.8.0, but when I build the SV 100.9.0-2840, there's no map / service loaded. We should investigate in a different issue

PR #528, opened 2020-07-15
New Sample: Realistic Lighting and Shadows

sclaridge:
Hello, 

I was able to take the existing [Realistic Lighting and Shadows](https://github.com/Esri/arcgis-runtime-samples-java/blob/realistic_environment_effect_sample/src/main/java/com/esri/samples/scene/realistic_lighting_and_shadows/RealisticLightingAndShadowsSample.java) Java sample and separate the view and the controller. Since it's my first sample I'm sure there is a lot I can improve. 

* the Portland buildings scene layer and camera, based on Hudson's Android sample
* `changeTimeOfDay()` to handle the changing slider value, inspired by the existing `setSliderLabels()`
* `setTimeLabel()`, inspired by `setSliderLabels()`
* in main.fxml, I want to make sure I replicated the padding and margins with FXML
* using SimpleDateFormat to set the time zones and tidy up the time label
* the button event handlers inside `initialize()` - should these be in a separate method?

Thanks


scene/realistic-light-and-shadows/README.md
```diff
@@ -0,0 +1,32 @@
+# Realistic Lighting and Shadows
```
JonLavi: The sample titles should be sentence case. 
We should double check the design to see if it's correct in there as well, just in case.

```suggestion
# Realistic lighting and shadows
```

scene/realistic-light-and-shadows/README.metadata.json
```diff
@@ -0,0 +1,35 @@
...
+    "LightingMode",
+    "SceneView.setSunLighting",
+    "SceneView.setSunTime"
+  ],
+  "snippets": [
+    "src/main/java/com.esri.samples.realistic_lighting_and_shadows/RealisticLightAndShadowsSample.java",
+    "src/main/java/com.esri.samples.realistic_lighting_and_shadows/RealisticLightAndShadowsController.java",
+    "src/main/resources/realistic_light_and_shadows/main.fxml"
+  ],
+  "title": "Realistic Lighting and Shadows"
```
JonLavi:
```suggestion
  "title": "Realistic lighting and shadows"
```

scene/realistic-light-and-shadows/build.gradle
```diff
@@ -0,0 +1,80 @@
...
+}
+
+javafx {
+    version = "11.0.2"
+    modules = ['javafx.controls',
+               'javafx.base',
+               'javafx.graphics',
+               'javafx.fxml',
+               'javafx.web',
+               'javafx.media']
```
JonLavi: We've recently had a discussion about which modules to add here.
There's a good argument for not adding all modules, but only the ones necessary for this particular sample, to reduce the amount of data that will be downloaded the first time you run it.

I think you will only need `controls` and `fxml`?

scene/realistic-light-and-shadows/build.gradle
```diff
@@ -0,0 +1,80 @@
...
+task productionZip(type: Zip) {
+    group = 'distribution'
+    from copyNatives
+    from jar.destinationDir
+    into(project.name)
+    baseName = project.name
+}
+
+wrapper {
+    gradleVersion = '5.0'
```
JonLavi: We've also recently noticed we can update this to a newer wrapper version:

```suggestion
    gradleVersion = '6.5.1'
```

scene/realistic-light-and-shadows/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -0,0 +1,5 @@
+distributionBase=GRADLE_USER_HOME
+distributionPath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-5.0-bin.zip
```
JonLavi: Another spot we should update the gradle version:

```suggestion
distributionUrl=https\://services.gradle.org/distributions/gradle-6.5.1-bin.zip
```

scene/realistic-light-and-shadows/settings.gradle
```diff
@@ -0,0 +1 @@
+rootProject.name = 'Realistic Lighting and Shadows'
```
JonLavi: We usually have this parameter as the name of the sample's folder. At the same time, we should probably rename the folder this sample sits in to reflect the full sample name `realistic-lighting-and-shadows`

```suggestion
rootProject.name = 'realistic-lighting-and-shadows'
```

scene/realistic-light-and-shadows/src/main/java/com/esri/samples/realistic_lighting_and_shadows/RealisticLightingAndShadowsController.java
```diff
@@ -0,0 +1,203 @@
...
+import java.util.GregorianCalendar;
+import java.util.TimeZone;
+
+import com.esri.arcgisruntime.mapping.ArcGISScene;
+import com.esri.arcgisruntime.mapping.ArcGISTiledElevationSource;
+import com.esri.arcgisruntime.mapping.Basemap;
+import com.esri.arcgisruntime.mapping.Surface;
+import com.esri.arcgisruntime.mapping.view.AtmosphereEffect;
+import com.esri.arcgisruntime.mapping.view.SceneView;
+import com.esri.arcgisruntime.mapping.view.Camera;
```
JonLavi: There's a particular order we want the imports to be in, and you can set up IntelliJ to do this for you automatically (see the screenshot below, the UI is a bit terrible since the window is not tall enough to show everything. Give me a shout if you get stuck).
The order would be:
```
import java.*
<newline>
import javafx.*
<newline>
import com.esri.*
<newline>
import static all other imports
```

image

scene/realistic-light-and-shadows/src/main/java/com/esri/samples/realistic_lighting_and_shadows/RealisticLightingAndShadowsController.java
```diff
@@ -0,0 +1,203 @@
...
+
+public class RealisticLightingAndShadowsController {
+
+  @FXML
+  private SceneView sceneView;
```
JonLavi: In some of our samples we have the `@FXML` tag on the same line we define the class members. I know the IDE tends to want to put these on different lines, so in some of our samples it's just like you have it here.
I kind of thing it's neater if they are on one line, but if the IDE formats it this way then it ought to be the 'best practice', so let's see if @Rachael-E has any strong feelings.

scene/realistic-light-and-shadows/src/main/java/com/esri/samples/realistic_lighting_and_shadows/RealisticLightingAndShadowsController.java
```diff
@@ -0,0 +1,203 @@
...
+      ArcGISScene scene = new ArcGISScene();
+      scene.setBasemap(Basemap.createTopographic());
+
+      // add the SceneView to the stack pane
+      sceneView.setArcGISScene(scene);
+
+      // add base surface for elevation data
+      surface = new Surface();
+      ArcGISTiledElevationSource elevationSource = new ArcGISTiledElevationSource("http://elevation3d.arcgis" +
+        ".com/arcgis/rest/services/WorldElevation3D/Terrain3D/ImageServer");
```
JonLavi: We don't strictly enforce a line length in samples, so I think we can just write this on one line:

```suggestion
      ArcGISTiledElevationSource elevationSource = new ArcGISTiledElevationSource("http://elevation3d.arcgis.com/arcgis/rest/services/WorldElevation3D/Terrain3D/ImageServer");
```

scene/realistic-light-and-shadows/src/main/java/com/esri/samples/realistic_lighting_and_shadows/RealisticLightingAndShadowsController.java
```diff
@@ -0,0 +1,203 @@
...
+  private Calendar calendar;
+
+  public void initialize() {
+    try {
+
+      // create a scene and add a basemap to it
+      ArcGISScene scene = new ArcGISScene();
+      scene.setBasemap(Basemap.createTopographic());
+
+      // add the SceneView to the stack pane
```
JonLavi: What's actually happening in the line below is that we're adding the ArcGISScene to our scene view, not the scene view to the stack pane, so let's update this comment:
```suggestion
      // add the scene to the scene view
```

scene/realistic-light-and-shadows/src/main/java/com/esri/samples/realistic_lighting_and_shadows/RealisticLightingAndShadowsController.java
```diff
@@ -0,0 +1,203 @@
...
+      sceneView.setArcGISScene(scene);
+
+      // add base surface for elevation data
```
JonLavi: We can be a tad more specific about where we're adding the base surface to
```suggestion
      // add a base surface for elevation data to the scene
```

scene/realistic-light-and-shadows/src/main/java/com/esri/samples/realistic_lighting_and_shadows/RealisticLightingAndShadowsController.java
```diff
@@ -0,0 +1,203 @@
...
+      surface = new Surface();
+      ArcGISTiledElevationSource elevationSource = new ArcGISTiledElevationSource("http://elevation3d.arcgis" +
+        ".com/arcgis/rest/services/WorldElevation3D/Terrain3D/ImageServer");
+      surface.getElevationSources().add(elevationSource);
+      scene.setBaseSurface(surface);
+
+      // add a scene layer
+      final String buildings = "http://tiles.arcgis.com/tiles/P3ePLMYs2RVChkJx/arcgis/rest/services" +
+        "/DevA_BuildingShells" +
+        "/SceneServer/layers/0";
```
JonLavi: Same here, I think we could do this in one line. If the auto-formatting is chopping up lines for you, you can increase the character limit in the settings.

```suggestion
      final String buildings = "http://tiles.arcgis.com/tiles/P3ePLMYs2RVChkJx/arcgis/rest/services/DevA_BuildingShells/SceneServer/layers/0";
```

scene/realistic-light-and-shadows/src/main/java/com/esri/samples/realistic_lighting_and_shadows/RealisticLightingAndShadowsController.java
```diff
@@ -0,0 +1,203 @@
...
+      ArcGISSceneLayer sceneLayer = new ArcGISSceneLayer(buildings);
+      scene.getOperationalLayers().add(sceneLayer);
+
+      // add a camera and initial camera position
+      Camera camera = new Camera(45.54605153789073, -122.69033380511073, 941.0002111233771, 162.58544227544266, 60.0,
+        0.0);
```
JonLavi:
```suggestion
      Camera camera = new Camera(45.54605153789073, -122.69033380511073, 941.0002111233771, 162.58544227544266, 60.0, 0.0);
```

scene/realistic-light-and-shadows/src/main/java/com/esri/samples/realistic_lighting_and_shadows/RealisticLightingAndShadowsController.java
```diff
@@ -0,0 +1,203 @@
...
+      sceneView.setViewpointCamera(camera);
+
+      // set atmosphere effect to realistic
+      sceneView.setAtmosphereEffect(AtmosphereEffect.REALISTIC);
+
+      //set the time label on the control panel
```
JonLavi: Just a wee missing space here:

```suggestion
      // set the time label on the control panel
```

scene/realistic-light-and-shadows/src/main/java/com/esri/samples/realistic_lighting_and_shadows/RealisticLightingAndShadowsController.java
```diff
@@ -0,0 +1,203 @@
...
+      setTimeLabel();
+
+      // set the slider to display tick labels as time strings
+      setSliderLabels();
+
+      //update the atmosphere effect based on the button clicked
```
JonLavi: Another missing space:
```suggestion
      // update the atmosphere effect based on the button clicked
```

scene/realistic-light-and-shadows/src/main/java/com/esri/samples/realistic_lighting_and_shadows/RealisticLightingAndShadowsController.java
```diff
@@ -0,0 +1,203 @@
...
+      noSunButton.setOnAction(event -> sceneView.setSunLighting(LightingMode.NO_LIGHT));
```
JonLavi: Instead of three distinct buttons, we might be better off using a toggle group with radio buttons (or a drop-down menu). The implication of either of these control elements is that one option always has to be chosen/active at any given time. Buttons on the other hand imply you perform a 'one off' action the moment you click them.
Check out how to create a toggle group here:
https://github.com/Esri/arcgis-runtime-samples-java/blob/master/network_analysis/service-area-task/src/main/resources/service_area_task/main.fxml#L38-L43

I think you can then interrogate the toggle group as to which radio button is selected, and pass that value into your method `setSunLighting`

scene/realistic-light-and-shadows/src/main/java/com/esri/samples/realistic_lighting_and_shadows/RealisticLightingAndShadowsLauncher.java
```diff
@@ -0,0 +1,11 @@
...
+
+/**
+ * Wrapper required for launching a JavaFX 11 app through Gradle or from a jar.
+ */
+public class RealisticLightingAndShadowsLauncher {
+
+  public static void main(String[] args) {
+    RealisticLightingAndShadowsSample.main(args);
+  }
+}
```
JonLavi: We're missing a new line in the end of this file. Github is kindly pointing that out 😄

scene/realistic-light-and-shadows/src/main/java/com/esri/samples/realistic_lighting_and_shadows/RealisticLightingAndShadowsSample.java
```diff
@@ -0,0 +1,68 @@
...
+  public void start(Stage stage) throws IOException {
+
+    // set up the scene
+    FXMLLoader loader = new FXMLLoader(getClass().getResource("/realistic_lighting_and_shadows/main.fxml"));
+    Parent root = loader.load();
+    controller = loader.getController();
+    Scene scene = new Scene(root);
+
+    // set title, size, and add JavaFX scene to stage
+    stage.setTitle("Realistic Environmental Atmosphere Sample");
```
JonLavi: Don't forget to update this to match the new sample's title:
```suggestion
    stage.setTitle("Realistic Lighting and Shadows Sample");
```

scene/realistic-light-and-shadows/src/main/java/com/esri/samples/realistic_lighting_and_shadows/RealisticLightingAndShadowsSample.java
```diff
@@ -0,0 +1,68 @@
...
+    stage.setWidth(800);
+    stage.setHeight(700);
+    stage.setScene(scene);
+    stage.show();
+  }
+
+  /**
+   * /**
+   * Stops and releases all resources used in application.
+   */
```
JonLavi: Something slightly odd happened with this JavaDoc:
```suggestion
  /**
   * Stops and releases all resources used in application.
   */
```

scene/realistic-light-and-shadows/src/main/java/com/esri/samples/realistic_lighting_and_shadows/RealisticLightingAndShadowsSample.java
```diff
@@ -0,0 +1,68 @@
...
+  /**
+   * Opens and runs application.
+   *
+   * @param args arguments passed to this application
+   */
+  public static void main(String[] args) {
+
+    Application.launch(args);
+  }
+
```
JonLavi: We can remove this extra line:
```suggestion
```

scene/realistic-light-and-shadows/src/main/resources/realistic_lighting_and_shadows/main.fxml
```diff
@@ -0,0 +1,65 @@
...
+            <Background>
+                <fills>
+                    <BackgroundFill>
+                        <fill>
+                            <Color opacity="0.3"/>
+                        </fill>
+                    </BackgroundFill>
+                </fills>
+            </Background>
+        </background>
```
JonLavi: We can refactor a lot of this styling stuff into the CSS file, so that it's easier for the users to focus just on the control elements. Check out these files from `animate images with image overlay`:

https://github.com/Esri/arcgis-runtime-samples-java/blob/master/scene/animate-images-with-image-overlay/src/main/resources/animate_images_with_image_overlay/main.fxml

You could also extract the styling from the VBoxes, Slider, and future RadioButtons.

scene/realistic-light-and-shadows/src/main/java/com/esri/samples/realistic_lighting_and_shadows/RealisticLightingAndShadowsController.java
```diff
@@ -0,0 +1,203 @@
...
+    SimpleDateFormat dateFormat = new SimpleDateFormat("EEE MMM dd HH:mm");
+    dateFormat.setTimeZone(TimeZone.getTimeZone("PST"));
+    String dateAndTimeTidied = dateFormat.format(calendar.getTime());
+
+    // set a label to display the tidied date and time
+    time.setText(dateAndTimeTidied);
+  }
+
+  /**
+   * Disposes application resources.
```
JonLavi: I think most newer samples try and use plain english here, so let's add `Disposes of...`
```suggestion
   * Disposes of application resources.
```

scene/realistic-light-and-shadows/src/main/java/com/esri/samples/realistic_lighting_and_shadows/RealisticLightingAndShadowsController.java
```diff
@@ -0,0 +1,203 @@
...
+    // when the slider changes, update the hour of the day based on the value of the slider
+    timeSlider.valueProperty().addListener((observable, oldValue, newValue) -> {
+
+        // get value from the slider
+        Double timeFromSlider = timeSlider.getValue();
+        // to get minutes from timer value, split double to the two values after decimal place
+        String timeAsString = timeFromSlider.toString();
+
+        // get the hour value from the slider
+        int hourFromSlider = timeFromSlider.intValue();
```
JonLavi: I've noticed the Kotlin sample has the slider represent minutes in `int` instead of a double:
https://github.com/Esri/arcgis-runtime-samples-android/pull/836/files?file-filters%5B%5D=.gradle&file-filters%5B%5D=.kt&file-filters%5B%5D=.png&file-filters%5B%5D=.pro&file-filters%5B%5D=.xml&file-filters%5B%5D=dotfile#diff-70792a6978609f728c4f1fb8210f44a3R87-R98

It looks like it may make the code a little more succinct, with no conversion between strings/substrings and changing the values from Double over String to Int.
Once you have the `int minutes` you can easily convert to hours and update your calendar?

scene/realistic-light-and-shadows/src/main/java/com/esri/samples/realistic_lighting_and_shadows/RealisticLightingAndShadowsController.java
```diff
@@ -0,0 +1,203 @@
...
+      // add a camera and initial camera position
+      Camera camera = new Camera(45.54605153789073, -122.69033380511073, 941.0002111233771, 162.58544227544266, 60.0,
+        0.0);
+      sceneView.setViewpointCamera(camera);
+
+      // set atmosphere effect to realistic
+      sceneView.setAtmosphereEffect(AtmosphereEffect.REALISTIC);
+
+      //set the time label on the control panel
+      setTimeLabel();
```
JonLavi: We probably don't want to extract this code into a method, but just put it in-line. The reason would be that there's no refactoring value just now. It might be counter-intuitive to write out a long `initialize()` method, instead of having many helper methods, but it does allow us to more or less read the file top to bottom and have a good idea of what's happening (instead of jumping back and forth between method definitions). Definitely not 'best coding practice', but helpful from a code sample perspective

scene/realistic-light-and-shadows/src/main/java/com/esri/samples/realistic_lighting_and_shadows/RealisticLightingAndShadowsController.java
```diff
@@ -0,0 +1,203 @@
...
+          // set the calendar for given hour and minute from slider value
+          calendar.set(2018, 7, 10, hourFromSlider, minuteFromSlider);
+
+        } else {
+          calendar.set(2018, 7, 10, hourFromSlider, 00);
+        }
+        // tidy string to just return date and time (hours and minutes)
+        SimpleDateFormat dateFormat = new SimpleDateFormat("EEE MMM dd HH:mm");
+        dateFormat.setTimeZone(TimeZone.getTimeZone("PST"));
+        String dynamicDateAndTimeTidied = dateFormat.format(calendar.getTime());
```
JonLavi: Once you remove the `setTimeLabel()` method, you could refactor this block of code by re-using the `dateFormat` object that will now be in the `initialize()` method (by making `dateFormat` a class-member)

scene/realistic-light-and-shadows/src/main/java/com/esri/samples/realistic_lighting_and_shadows/RealisticLightingAndShadowsController.java
```diff
@@ -0,0 +1,203 @@
...
+          String subString = timeAsString.substring(3, 5);
+
+          int minutes = Integer.valueOf(subString);
+          // convert figures into minutes
+          float actualMinutes = ((float) minutes / (float) 100) * (float) 60;
+          // round into an integer
+          int minuteFromSlider = Math.round(actualMinutes);
+
+          // set the calendar for given hour and minute from slider value
+          calendar.set(2018, 7, 10, hourFromSlider, minuteFromSlider);
```
JonLavi: I _think_ you could simplify setting the hour/minute by doing:
```
            calendar.set(Calendar.HOUR_OF_DAY, hourFromSlider);
            calendar.set(Calendar.MINUTE, minuteFromSlider);
```
That way, you shouldn't have to specify the year/month/day.
Let me know if that works!

scene/realistic-light-and-shadows/src/main/java/com/esri/samples/realistic_lighting_and_shadows/RealisticLightingAndShadowsController.java
```diff
@@ -0,0 +1,203 @@
...
+
+        } else {
+          calendar.set(2018, 7, 10, hourFromSlider, 00);
+        }
```
JonLavi: I think by grabbing the slider value as `int`s representing minutes, you'd be able to avoid the if/else statement based on the length of the string. Curious to see if it's possible to borrow Kotlin's design!

scene/realistic-light-and-shadows/src/main/java/com/esri/samples/realistic_lighting_and_shadows/RealisticLightingAndShadowsController.java
```diff
@@ -0,0 +1,203 @@
...
+        // set the sun time to calendar
+        sceneView.setSunTime(calendar);
+      }
+    );
+  }
+
+  /**
+   * Set labels to display on the slider.
+   */
+  private void setSliderLabels() {
```
JonLavi: Even though I suggested getting rid of `setTimeLabel()` and adding the code within that into `initialize()`, I'd be happy for this to stay here, as it's mainly UI setup (so less relevant to the sample's logic.

I've never had to deal with setting custom labels, but a quick search online shows the same pattern. It looks like this is what may be tying the slider to return `Double` values? Either way, you should be able to get it to return `int`s that represent minutes quite easily.

JonLavi (commented):
Nice work @sclaridge , it's looking very good! And well done on dealing with the date/time stuff, it's always a bit tricky.

A few comments:
For the sake of tidyness let's give the folder name for this sample the full title name:
`realistic-light-and-shadows` -> `realistic-lighting-and-shadows`.

The main comment I have would be using a RadioButtons instead of dedicated buttons for each lighting mode, because it has the implication of a chosen and 'active' setting, rather than a one-off operation of a button.

Also looking at Kotlin's implementation, it may be easier to grab the slider value as an `int` representing minutes, and to calculate the hours:minutes from that to simplify the code a little.

JonLavi:
Hi @sclaridge , just to get back to the specific points you asked about, and I didn't manage to address within the code:

>    the Portland buildings scene layer and camera, based on Hudson's Android sample

This is just fine. We generally keep the data and viewpoints consistent between all the SDKs so it's nice and neat 
 
>    the button event handlers inside initialize() - should these be in a separate method?

These are good where they are, but once we move to using RadioButtons, it may be possible to trigger a method using a reference in the FXML definition of each RadioButton, e.g.
`<RadioButton onAction="#changeLightingMode"/>`

scene/realistic-light-and-shadows/src/main/java/com/esri/samples/realistic_lighting_and_shadows/RealisticLightingAndShadowsController.java
sclaridge: Would this be the best way to check which radio button has been selected?
```
// update the atmosphere effect based on the button clicked
      toggleGroup.selectedToggleProperty().addListener((observable, oldValue, newValue) -> {
        if (noSunButton.isSelected()){
          sceneView.setSunLighting(LightingMode.NO_LIGHT);
        }
        else if (sunOnlyButton.isSelected()){
          sceneView.setSunLighting(LightingMode.LIGHT);
        }
        else if (sunAndShadowsButton.isSelected()){
          sceneView.setSunLighting(LightingMode.LIGHT_AND_SHADOWS);
        }
      });
```

JonLavi: That would certainly work! A perhaps sleeker way would be to set the 'user data' (it's like a readable property) of the toggle button to be a value of the `LightingMode` enum, and read that out in the listener. The `surface placement` placement sample does this, so you could borrow from there:
https://github.com/Esri/arcgis-runtime-samples-java/blob/master/scene/surface-placement/src/main/java/com/esri/samples/surface_placement/SurfacePlacementSample.java#L142-L159

@jenmerritt will be turning `surface placement` into a sample using fxml, so you could both check if there's a way to set the `user data` within the fxml or not (I don't actually know!)

scene/realistic-light-and-shadows/src/main/java/com/esri/samples/realistic_lighting_and_shadows/RealisticLightingAndShadowsController.java
sclaridge: This does work!

scene/realistic-light-and-shadows/src/main/java/com/esri/samples/realistic_lighting_and_shadows/RealisticLightingAndShadowsController.java
sclaridge: I couldn't find a way to set the user data to the LightMode in FXML. However, does this seem correct?
```
// set the lighting mode of the buttons
noSunButton.setUserData(LightingMode.NO_LIGHT);
sunOnlyButton.setUserData(LightingMode.LIGHT);
sunAndShadowsButton.setUserData(LightingMode.LIGHT_AND_SHADOWS);

// update the atmosphere effect based on the button clicked
toggleGroup.selectedToggleProperty().addListener((observable, oldToggle, newToggle) -> {
	if (toggleGroup.getSelectedToggle() != null)
	  sceneView.setSunLighting((LightingMode) toggleGroup.getSelectedToggle().getUserData());
  });
```

JonLavi: Looks great!

scene/realistic-lighting-and-shadows/src/main/java/com/esri/samples/realistic_lighting_and_shadows/RealisticLightingAndShadowsController.java
```diff
@@ -61,14 +60,12 @@ public void initialize() {
...
-        ArcGISTiledElevationSource("http://elevation3d.arcgis.com/arcgis/rest/services/WorldElevation3D/Terrain3D/ImageServer");
+      ArcGISTiledElevationSource elevationSource = new ArcGISTiledElevationSource("http://elevation3d.arcgis.com/arcgis/rest/services/WorldElevation3D/Terrain3D/ImageServer");
       surface.getElevationSources().add(elevationSource);
       scene.setBaseSurface(surface);
 
       // add a scene layer
-      final String buildings =
-        "http://tiles.arcgis.com/tiles/P3ePLMYs2RVChkJx/arcgis/rest/services/DevA_BuildingShells/SceneServer/layers/0";
+      final String buildings = "http://tiles.arcgis.com/tiles/P3ePLMYs2RVChkJx/arcgis/rest/services/DevA_BuildingShells/SceneServer/layers/0";
       ArcGISSceneLayer sceneLayer = new ArcGISSceneLayer(buildings);
```
JonLavi: I don't think we're re-using this URL either, so we can just use it in place.
```suggestion
      ArcGISSceneLayer sceneLayer = new ArcGISSceneLayer("http://tiles.arcgis.com/tiles/P3ePLMYs2RVChkJx/arcgis/rest/services/DevA_BuildingShells/SceneServer/layers/0");
```

scene/realistic-lighting-and-shadows/src/main/java/com/esri/samples/realistic_lighting_and_shadows/RealisticLightingAndShadowsController.java
```diff
@@ -76,8 +77,20 @@ public void initialize() {
       // set atmosphere effect to realistic
       sceneView.setAtmosphereEffect(AtmosphereEffect.REALISTIC);
 
+      // set a new calendar and add a date and time
+      calendar = new GregorianCalendar(2018, 7, 10, 12, 00, 0);
+      calendar.setTimeZone(TimeZone.getTimeZone("PST"));
+
       // set the time label on the control panel
```
JonLavi: Looks like this comment might be out of date now?

scene/realistic-lighting-and-shadows/src/main/java/com/esri/samples/realistic_lighting_and_shadows/RealisticLightingAndShadowsController.java
```diff
@@ -76,8 +77,20 @@ public void initialize() {
...
-      setTimeLabel();
+      sceneView.setSunTime(calendar);
+
+      // tidy string to just return date and time (hours and minutes)
+      dateFormat = new SimpleDateFormat("EEE MMM dd HH:mm");
+      dateFormat.setTimeZone(TimeZone.getTimeZone("PST"));
+      String dateAndTimeTidied = dateFormat.format(calendar.getTime());
+
+      // set a label to display the tidied date and time
+      time.setText(dateAndTimeTidied);
```
JonLavi: Perhaps we can call this 'format the string'? We should also name the Label object `timeLabel`, since it helps to identify what kind of class it is.

```suggestion
      // format the string to just return date and time (hours and minutes)
      dateFormat = new SimpleDateFormat("EEE MMM dd HH:mm");
      dateFormat.setTimeZone(TimeZone.getTimeZone("PST"));
      String formattedDateAndTime = dateFormat.format(calendar.getTime());

      // set a label to display the tidied date and time
      timeLabel.setText(formattedDateAndTime);
```

scene/realistic-lighting-and-shadows/src/main/java/com/esri/samples/realistic_lighting_and_shadows/RealisticLightingAndShadowsController.java
```diff
@@ -119,14 +132,13 @@ public void changeTimeOfDay() {
...
+          calendar.set(Calendar.HOUR_OF_DAY, hourFromSlider);
+          calendar.set(Calendar.MINUTE, minuteFromSlider);
 
         } else {
           calendar.set(2018, 7, 10, hourFromSlider, 00);
         }
         // tidy string to just return date and time (hours and minutes)
-        SimpleDateFormat dateFormat = new SimpleDateFormat("EEE MMM dd HH:mm");
-        dateFormat.setTimeZone(TimeZone.getTimeZone("PST"));
         String dynamicDateAndTimeTidied = dateFormat.format(calendar.getTime());
```
JonLavi: Should we call this `formatted` as well?

scene/realistic-lighting-and-shadows/src/main/java/com/esri/samples/realistic_lighting_and_shadows/RealisticLightingAndShadowsController.java
```diff
@@ -176,6 +172,43 @@ public Double fromString(String string) {
...
+        else return "Light only";
+      }
+
+      @Override
+      public LightingMode fromString(String string) {
+        return null;
+      }
+    });
+
+    comboBox.setCellFactory(comboBox -> new ListCell<LightingMode>() {
```
JonLavi: Nice work on using the cell factory here!
We often put these new definitions of the ListCell class in their own file. Check out an example here:
https://github.com/Esri/arcgis-runtime-samples-java/blob/master/utility_network/trace-a-utility-network/src/main/java/com/esri/samples/trace_a_utility_network/UtilityTerminalListCell.java

With the if/else block you have below, we might be able to get away without the `StringConverter` above on lines 180- 195 (but I might be wrong). This is because you can shove the values from the `LightningMode` enum into the combobox, and use the ListCell to customise what is actually displayed in the UI.

scene/realistic-lighting-and-shadows/src/main/resources/realistic_lighting_and_shadows/main.fxml
```diff
@@ -33,9 +36,15 @@
     <!--Slider box-->
     <VBox styleClass="panel-region">
         <Label fx:id="time"/>
-        <Button fx:id="noSunButton" text="No sun light effect" maxWidth="Infinity"/>
-        <Button fx:id="sunOnlyButton" text="Sun light only" maxWidth="Infinity"/>
-        <Button fx:id="sunAndShadowsButton" text="Sun light with shadows" maxWidth="Infinity"/>
+        <ComboBox fx:id="comboBox" maxWidth="Infinity">
+            <items>
+                <FXCollections fx:factory="observableArrayList">
+                    <LightingMode fx:value="NO_LIGHT"/>
```
JonLavi: Great find on this! I didn't know it's possible to populate runtime API values into this using FXML 👍

scene/realistic-lighting-and-shadows/src/main/java/com/esri/samples/realistic_lighting_and_shadows/RealisticLightingAndShadowsController.java
```diff
@@ -95,50 +96,44 @@ public void initialize() {
       // set the slider to display tick labels as time strings
       setSliderLabels();
 
-      
+      // set the combo box to display the lighting modes as strings
+      setComboBoxLabels();
```
JonLavi: I think you might not need this any more if you're using the ListCell method to modify the ComboBox display of each entry

scene/realistic-lighting-and-shadows/src/main/java/com/esri/samples/realistic_lighting_and_shadows/RealisticLightingAndShadowsController.java
```diff
@@ -0,0 +1,217 @@
...
+import com.esri.arcgisruntime.mapping.view.Camera;
+import com.esri.arcgisruntime.mapping.view.LightingMode;
+import com.esri.arcgisruntime.mapping.view.SceneView;
+
+import static java.lang.StrictMath.floor;
+
+public class RealisticLightingAndShadowsController {
+
+  @FXML private SceneView sceneView;
+  @FXML private Label time;
```
JonLavi: We should name this `timeLabel` so that it's easier to tell what Type this object is from the var name

scene/realistic-lighting-and-shadows/src/main/java/com/esri/samples/realistic_lighting_and_shadows/RealisticLightingAndShadowsController.java
```diff
@@ -0,0 +1,217 @@
...
+
+      // add a base surface for elevation data to the scene
+      surface = new Surface();
+      ArcGISTiledElevationSource elevationSource = new ArcGISTiledElevationSource("http://elevation3d.arcgis.com/arcgis/rest/services/WorldElevation3D/Terrain3D/ImageServer");
+      surface.getElevationSources().add(elevationSource);
+      scene.setBaseSurface(surface);
+
+      // add a scene layer
+      final String buildings = "http://tiles.arcgis.com/tiles/P3ePLMYs2RVChkJx/arcgis/rest/services/DevA_BuildingShells/SceneServer/layers/0";
+      ArcGISSceneLayer sceneLayer = new ArcGISSceneLayer(buildings);
```
JonLavi: I don't think the string variable is being re-used so we can just write the URL in line
```suggestion
      ArcGISSceneLayer sceneLayer = new ArcGISSceneLayer("http://tiles.arcgis.com/tiles/P3ePLMYs2RVChkJx/arcgis/rest/services/DevA_BuildingShells/SceneServer/layers/0");
```

scene/realistic-lighting-and-shadows/src/main/java/com/esri/samples/realistic_lighting_and_shadows/RealisticLightingAndShadowsController.java
```diff
@@ -0,0 +1,217 @@
...
+      // on any error, display the stack trace.
+      e.printStackTrace();
+    }
+  }
+
+  /**
+   * Set the sun time based on the time from the slider.
+   */
+  @FXML
+  public void changeTimeOfDay() {
```
JonLavi: I personally thing 'update' is a bit more fitting for a method name

```suggestion
  public void updateTimeOfDay() {
```

scene/realistic-lighting-and-shadows/src/main/java/com/esri/samples/realistic_lighting_and_shadows/RealisticLightingAndShadowsController.java
```diff
@@ -0,0 +1,217 @@
...
+        // get the minutes from the slider
+        double afterDecimal = newValue.doubleValue() - floor(newValue.doubleValue());
+        int minutes = (int) (afterDecimal * 60);
+
+        // set the calendar for given hour and minute from slider value
+        calendar.set(Calendar.HOUR_OF_DAY, hours);
+        calendar.set(Calendar.MINUTE, minutes);
+
+        // tidy string to just return date and time (hours and minutes)
+        String dateAndTimeTidied = dateFormat.format(calendar.getTime());
```
JonLavi: We can also say 'format' the string here

scene/realistic-lighting-and-shadows/src/main/resources/realistic_lighting_and_shadows/main.fxml
```diff
@@ -0,0 +1,50 @@
...
+                    <LightingMode fx:value="LIGHT"/>
+                    <LightingMode fx:value="LIGHT_AND_SHADOWS"/>
+                </FXCollections>
+            </items>
+        </ComboBox>
+        <Slider fx:id="timeSlider" value="12" max="24" majorTickUnit="4" minorTickCount="60" showTickMarks="true"
```
JonLavi: If we multiply the min/max values here by 60, we can get the values from the slider represented in minutes. It might make the calculation in `changeTimeOfDay` a tad simpler by simply doing
```
int hours = sliderValue / 60;
int minutes = sliderValue % 60;
```

I accept that this is a matter of preference though, so up to you :)

scene/realistic-lighting-and-shadows/src/main/java/com/esri/samples/realistic_lighting_and_shadows/RealisticLightingAndShadowsController.java
```diff
@@ -0,0 +1,217 @@
...
+        if (hour == 20) return "8pm";
+
+        return "Midnight";
+      }
+
+      @Override
+      public Double fromString(String string) {
+        return null;
+      }
+    });
```
JonLavi: On second thought, we can also extract this new class definition for StringConverter into a separate file, then just do
`timeSlider.setLabelFormatter(new CustomStringConverter());`
That way we can leave this method out and have a single line in the main `initialize()` method.

My thought behind this is that it's more JavaFX/UI setup, and our main goal with the Java samples is to teach people the main concepts of the Runtime API. In this case the 'meat' of this sample is your `changeTimeOfDay()` method so we want to bring that front and centre

scene/realistic-lighting-and-shadows/src/main/java/com/esri/samples/realistic_lighting_and_shadows/RealisticLightingAndShadowsSample.java
```diff
@@ -0,0 +1,65 @@
...
+  /**
+   * Opens and runs application.
+   *
+   * @param args arguments passed to this application
+   */
+  public static void main(String[] args) {
+
+    Application.launch(args);
+  }
+}
```
JonLavi: GitHub is telling us there's a missing newline here.

JonLavi (commented):
Nice work @sclaridge , the ComboBox is looking great. I have a few more suggestions for refactoring some of the UI setup into their own files.
Check out this file for an example:
https://github.com/Esri/arcgis-runtime-samples-java/blob/master/utility_network/trace-a-utility-network/src/main/java/com/esri/samples/trace_a_utility_network/UtilityTerminalListCell.java

Also a couple of suggestions for wording/variable naming, but feel free to keep yours if you have a preference.

JonLavi:
Also, I think this sample would look great in the SampleViewer if we recorded a gif!
I use 'snagit' (which should be available in your SoftwareCenter) to record gifs. It's quite intuitive, but let me know if you get stuck. The dimensions for the gif should be the same as the screenshot, 800x400 px

scene/realistic-lighting-and-shadows/README.md
```diff
@@ -2,7 +2,7 @@
 
 Show realistic lighting and shadows for the time of day.
 
-![Image of realistic lighting and shadows](RealisticLightingAndShadows.png)
+![Image of realistic lighting and shadows](RealisticLightingAndShadows.gif)
```
JonLavi: Sorry about the confusion @sclaridge ! We generally keep both the PNG and the GIF if there is one. The readme then links the PNG, and the SampleViewer will automatically show the GIF if there is one

scene/realistic-lighting-and-shadows/src/main/java/com/esri/samples/realistic_lighting_and_shadows/ComboBoxStringConverter.java
```diff
@@ -0,0 +1,38 @@
...
+ * limitations under the License.
+ */
+
+package com.esri.samples.realistic_lighting_and_shadows;
+
+import javafx.util.StringConverter;
+
+import com.esri.arcgisruntime.mapping.view.LightingMode;
+
+public class ComboBoxStringConverter extends StringConverter<LightingMode> {
```
JonLavi: This is great. We just need to add a JavaDoc style comment, to explain a little bit what this class does:

```suggestion
/**
 * Converts the LightningMode values to strings to display the selected value in the ComboBox.
 */
public class ComboBoxStringConverter extends StringConverter<LightingMode> {
```

scene/realistic-lighting-and-shadows/src/main/java/com/esri/samples/realistic_lighting_and_shadows/LightingModeListCell.java
```diff
@@ -0,0 +1,35 @@
...
+ * limitations under the License.
+ */
+
+package com.esri.samples.realistic_lighting_and_shadows;
+
+import javafx.scene.control.ListCell;
+
+import com.esri.arcgisruntime.mapping.view.LightingMode;
+
+public class LightingModeListCell extends ListCell<LightingMode> {
```
JonLavi: Same here, we just need to give this class a little bit of documentation

```suggestion
/**
 * Converts the LightningMode values to strings to display in the open ComboBox.
 */
public class LightingModeListCell extends ListCell<LightingMode> {
```

scene/realistic-lighting-and-shadows/src/main/java/com/esri/samples/realistic_lighting_and_shadows/RealisticLightingAndShadowsController.java
```diff
@@ -38,17 +36,15 @@
...
 
   @FXML private SceneView sceneView;
-  @FXML private Label time;
+  @FXML private Label timeLabel;
   @FXML private Slider timeSlider;
   @FXML private ComboBox<LightingMode> comboBox;
   private Surface surface;
   private Calendar calendar;
-  private SimpleDateFormat dateFormat;
+  private SimpleDateFormat formatted;
```
JonLavi: Sorry if I caused any confusion here: I think `dateFormat` is a good variable name for this object. The comments previously were just to suggest to use the phrasing `format the string` instead of `tidy the string`.

scene/realistic-lighting-and-shadows/src/main/java/com/esri/samples/realistic_lighting_and_shadows/RealisticLightingAndShadowsController.java
```diff
@@ -78,26 +73,32 @@ public void initialize() {
...
       sceneView.setSunTime(calendar);
 
-      // tidy string to just return date and time (hours and minutes)
-      dateFormat = new SimpleDateFormat("EEE MMM dd HH:mm");
-      dateFormat.setTimeZone(TimeZone.getTimeZone("PST"));
-      String dateAndTimeTidied = dateFormat.format(calendar.getTime());
+      // format the string to just return the date and time (hours and minutes)
+      formatted = new SimpleDateFormat("EEE MMM dd HH:mm");
+      formatted.setTimeZone(TimeZone.getTimeZone("PST"));
+      String formattedDateAndTime = formatted.format(calendar.getTime());
```
JonLavi: Sorry about the confusion here! We should still give the `SimpleDateFormat` variable a representative name:

```suggestion
      // format the string to just return the date and time (hours and minutes)
      dateFormat= new SimpleDateFormat("EEE MMM dd HH:mm");
      dateFormat.setTimeZone(TimeZone.getTimeZone("PST"));
      String formattedDateAndTime = formatted.format(calendar.getTime());
```

scene/realistic-lighting-and-shadows/src/main/java/com/esri/samples/realistic_lighting_and_shadows/SliderStringConverter.java
```diff
@@ -0,0 +1,38 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.esri.samples.realistic_lighting_and_shadows;
+
+import javafx.util.StringConverter;
+
+public class SliderStringConverter extends StringConverter<Double> {
```
JonLavi: Let's give this a bit of documentation as well. Feel free to just follow the examples I gave at the top

scene/realistic-lighting-and-shadows/src/main/resources/realistic_lighting_and_shadows/main.fxml
```diff
@@ -17,34 +17,25 @@
...
-                    <LightingMode fx:value="LIGHT"/>
-                    <LightingMode fx:value="LIGHT_AND_SHADOWS"/>
-                </FXCollections>
-            </items>
         </ComboBox>
-        <Slider fx:id="timeSlider" value="12" max="24" majorTickUnit="4" minorTickCount="60" showTickMarks="true"
-                showTickLabels="true" onMouseDragged="#changeTimeOfDay">
+        <Slider fx:id="timeSlider" max="1440" majorTickUnit="240" minorTickCount="3600" value="720" showTickMarks="true"
+                showTickLabels="true" onMouseDragged="#updateTimeOfDay">
         </Slider>
```
JonLavi: Nice work on updating the Slider. Remember you don't need an opening and closing tag if there's nothing between them. You could just write `<Slider />`. Same goes for the ComboBox above now that the values are added in the Controller class
```suggestion
        <Slider fx:id="timeSlider" max="1440" majorTickUnit="240" minorTickCount="3600" value="720" showTickMarks="true"
                showTickLabels="true" onMouseDragged="#updateTimeOfDay" />
```

JonLavi: It also looks like you could move some of the styling of the slider into the CSS by adding the following to the `.slider` rule, and removing the respective params here:
```
  -fx-show-tick-labels: true;
  -fx-show-tick-marks: true;
  -fx-major-tick-unit: 240;
  -fx-major-tick-unit: 3600;
```
Better test it out though, since CSS can be a bit iffy with FXML. If it doesn't work easily happy to just leave this styling in the FXML

scene/realistic-lighting-and-shadows/README.md
JonLavi: I'm also noticing the GIF looks a little small, it's at 400x200 px I think, and should be 800x400

JonLavi (commented):
This looks great @sclaridge !
Just a couple of things to fix, which were mainly my fault because I must have been a bit confusing with my previous comments 😄

JonLavi (approved):
Nice work, I think this looks good!
It doesn't look like the minor tick marks are actually showing, so we can just get rid of them. With the label showing the exact selected time, there's plenty of information there already.

scene/realistic-lighting-and-shadows/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+![Image of realistic lighting and shadows](RealisticLightingAndShadows.png)
+
+## Use case
+
+You can use realistic lighting to evaluate the shadow impact of buildings and utility infrastructure on the surrounding community. This could be useful for civil engineers and urban planners, or for events management assessing the impact of building shadows during an outdoor event.
+
+## How to use the sample
+
+Select one of the three lighting options to show that lighting effect on the SceneView. Select a time of day from the slider (based on a 24hr clock) to show the lighting for that time of day in the SceneView.
```
Rachael-E: in the "How to use the sample" we avoid directly referencing any API classes since the focus is on how a user interacts with a sample. I'd suggest the following:

```suggestion
Select one of the three available lighting options to display that lighting effect. Adjust the slider to show the lighting effect for a particular time of day. The 3D buildings will display shadows when "Sun light with shadows" is selected. 
```

scene/realistic-lighting-and-shadows/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+## How it works
+To add realistic lighting and shadows to the SceneView:
```
Rachael-E: during earlier README formats, we kept a sentence like this in to lead the reader into the steps. However a decision was reached last year to remove any introductory sentences and jump straight into the steps. So we can remove this sentence. 
```suggestion
```

scene/realistic-lighting-and-shadows/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+1. Create an `ArcGISScene` and display it in a `SceneView`.
+2. Create a `Calendar` to define the time of day.
+3. Set the sun time to that calendar with `sceneView.setSunTime(calendar)`.
+4. Set the lighting mode of the SceneView to **no light**, **light**, or **light and shadows** with `sceneView.setSunLighting(LightingMode)`.
```
Rachael-E: since we use enums to define the lighting mode, we should refer to them as such here, along with referring to lighting mode as a class (since it's the first time we're referring to lighting mode) and SceneView as scene view (since we've already described it as a class in step 1, and can therefore refer to it in plain english after that first usage) e.g. 
```suggestion
4. Set the `LightingMode` of the scene view to `NO_LIGHT`, `LIGHT`, or `LIGHT_AND_SHADOWS` with `sceneView.setSunLighting(LightingMode)`.
```

scene/realistic-lighting-and-shadows/src/main/java/com/esri/samples/realistic_lighting_and_shadows/ComboBoxStringConverter.java
```diff
@@ -0,0 +1,40 @@
...
+ * limitations under the License.
+ */
+
+package com.esri.samples.realistic_lighting_and_shadows;
+
+import javafx.util.StringConverter;
+
+import com.esri.arcgisruntime.mapping.view.LightingMode;
+/**
+ * Converts the LightningMode values to strings to display in the open ComboBox.
```
Rachael-E: this can be simplified with the following, since it's not only displaying the string when the combobox is opened. Also corrected the typo here!
```suggestion
 * Converts the LightingMode values to strings to display in the ComboBox.
```

scene/realistic-lighting-and-shadows/src/main/java/com/esri/samples/realistic_lighting_and_shadows/LightingModeListCell.java
```diff
@@ -0,0 +1,37 @@
...
+ * limitations under the License.
+ */
+
+package com.esri.samples.realistic_lighting_and_shadows;
+
+import javafx.scene.control.ListCell;
+
+import com.esri.arcgisruntime.mapping.view.LightingMode;
+/**
+ * Converts the LightningMode values to strings to display in the open ComboBox.
```
Rachael-E: Lightning mode would be really cool though!
```suggestion
 * Converts the LightingMode values to strings to display in the open ComboBox.
```

scene/realistic-lighting-and-shadows/src/main/java/com/esri/samples/realistic_lighting_and_shadows/RealisticLightingAndShadowsController.java
```diff
@@ -0,0 +1,157 @@
...
+import com.esri.arcgisruntime.mapping.view.LightingMode;
+import com.esri.arcgisruntime.mapping.view.SceneView;
+
+public class RealisticLightingAndShadowsController {
+
+  @FXML private SceneView sceneView;
+  @FXML private Label timeLabel;
+  @FXML private Slider timeSlider;
+  @FXML private ComboBox<LightingMode> comboBox;
+  private Surface surface;
```
Rachael-E: this can be converted to a local variable on line 60, since it's only being referenced within the `initialize` method.

scene/realistic-lighting-and-shadows/src/main/java/com/esri/samples/realistic_lighting_and_shadows/RealisticLightingAndShadowsController.java
```diff
@@ -0,0 +1,157 @@
...
+    try {
+
+      // create a scene and add a basemap to it
+      ArcGISScene scene = new ArcGISScene();
+      scene.setBasemap(Basemap.createTopographic());
+
+      // add the scene to the scene view
+      sceneView.setArcGISScene(scene);
+
+      // add a base surface for elevation data to the scene
```
Rachael-E:
```suggestion
      // add a base surface with an elevation source to the scene
```

scene/realistic-lighting-and-shadows/src/main/java/com/esri/samples/realistic_lighting_and_shadows/RealisticLightingAndShadowsController.java
```diff
@@ -0,0 +1,157 @@
...
+      surface = new Surface();
+      ArcGISTiledElevationSource elevationSource = new ArcGISTiledElevationSource("http://elevation3d.arcgis.com/arcgis/rest/services/WorldElevation3D/Terrain3D/ImageServer");
+      surface.getElevationSources().add(elevationSource);
+      scene.setBaseSurface(surface);
+
+      // add a scene layer
```
Rachael-E: we could include a little more info here to help a new user understand that it is the scene layer that is showing the 3D buildings
```suggestion
      // add 3D building shells with a scene layer
```

scene/realistic-lighting-and-shadows/src/main/java/com/esri/samples/realistic_lighting_and_shadows/RealisticLightingAndShadowsController.java
```diff
@@ -0,0 +1,157 @@
...
+      ArcGISSceneLayer sceneLayer = new ArcGISSceneLayer("http://tiles.arcgis.com/tiles/P3ePLMYs2RVChkJx/arcgis/rest/services/DevA_BuildingShells/SceneServer/layers/0");
+      scene.getOperationalLayers().add(sceneLayer);
+
+      // add a camera and initial camera position
```
Rachael-E: we can be a little more concise with the comment here:
```suggestion
      // add a camera and set the scene view's viewpoint to it
```

scene/realistic-lighting-and-shadows/src/main/java/com/esri/samples/realistic_lighting_and_shadows/RealisticLightingAndShadowsController.java
```diff
@@ -0,0 +1,157 @@
...
+      Camera camera = new Camera(45.54605153789073, -122.69033380511073, 941.0002111233771, 162.58544227544266, 60.0,0.0);
+      sceneView.setViewpointCamera(camera);
+
+      // set atmosphere effect to realistic
+      sceneView.setAtmosphereEffect(AtmosphereEffect.REALISTIC);
+
+      // set a calendar with a date and time
+      calendar = new GregorianCalendar(2018, 7, 10, 12, 00, 0);
```
Rachael-E: I got two IDE tips appearing for this constructor. `minute` can be replaced with "0" and the integer reflecting the month should be replaced with an enum e.g.
```suggestion
      calendar = new GregorianCalendar(2018, Calendar.AUGUST, 10, 12, 0, 0);
```

scene/realistic-lighting-and-shadows/src/main/java/com/esri/samples/realistic_lighting_and_shadows/RealisticLightingAndShadowsController.java
```diff
@@ -0,0 +1,157 @@
...
+      // set a label to display the formatted date and time
+      timeLabel.setText(formattedDateAndTime);
+
+      // set the slider to display tick labels as time strings
+      timeSlider.setLabelFormatter(new SliderStringConverter());
+
+      // add the lighting modes to the combo box
+      comboBox.getItems().add(LightingMode.NO_LIGHT);
+      comboBox.getItems().add(LightingMode.LIGHT);
+      comboBox.getItems().add(LightingMode.LIGHT_AND_SHADOWS);
```
Rachael-E: you can reduce the number of lines here by using `addAll`: 
```suggestion
      comboBox.getItems().addAll(LightingMode.LIGHT, LightingMode.LIGHT_AND_SHADOWS, LightingMode.NO_LIGHT);

```

scene/realistic-lighting-and-shadows/src/main/java/com/esri/samples/realistic_lighting_and_shadows/RealisticLightingAndShadowsController.java
```diff
@@ -0,0 +1,157 @@
...
+
+      // show the name of the lighting modes in the combo box
+      comboBox.setConverter(new ComboBoxStringConverter());
+      comboBox.setCellFactory(comboBox -> new LightingModeListCell());
+
+      // update the atmosphere effect based on the lighting mode chosen from the combo box
```
Rachael-E: atmosphere effect is a separate part of our API (e.g. `sceneView.setAtmosphereEffect(AtmosphereEffect.REALISTIC);`), so we should be specific to the API that we are showcasing in this sample with the following: 
```suggestion
      // update the sun lighting based on the lighting mode chosen from the combo box
```

scene/realistic-lighting-and-shadows/src/main/resources/realistic_lighting_and_shadows/main.fxml
```diff
@@ -0,0 +1,39 @@
...
+
+<?import com.esri.arcgisruntime.mapping.view.SceneView?>
+<?import javafx.geometry.Insets?>
+<?import javafx.scene.control.Label?>
+<?import javafx.scene.control.Slider?>
+<?import javafx.scene.control.ComboBox?>
+<StackPane fx:controller="com.esri.samples.realistic_lighting_and_shadows.RealisticLightingAndShadowsController"
+           xmlns="http://javafx.com/javafx"
+           xmlns:fx="http://javafx.com/fxml" stylesheets="/realistic_lighting_and_shadows/style.css">
+    <!--SDK SceneView-->
```
Rachael-E: we don't need "SDK" in here, just our API class name is clear enough.
```suggestion
    <!--SceneView-->
```

scene/realistic-lighting-and-shadows/src/main/resources/realistic_lighting_and_shadows/style.css
```diff
@@ -0,0 +1,28 @@
...
+  -fx-text-fill: white;
+}
+
+.slider {
+  -fx-padding: 10;
+  -fx-show-tick-labels: true;
+  -fx-show-tick-marks: true;
+  -fx-major-tick-unit: 240;
+  -fx-minor-tick-count: 60;
+}
```
Rachael-E: our sample viewer seems to have some problems with this styling specifically. it's an issue we need to work out but for now, please can you move this styling back into main.fxml?

scene/realistic-lighting-and-shadows/src/main/java/com/esri/samples/realistic_lighting_and_shadows/ComboBoxStringConverter.java
```diff
@@ -0,0 +1,40 @@
...
+ */
+public class ComboBoxStringConverter extends StringConverter<LightingMode> {
+
+  @Override
+  public String toString(LightingMode mode) {
```
Rachael-E: from a naming perspective, always best err on the side of super clear if there is space for it: let's change "mode" for "lightingMode"
```suggestion
  public String toString(LightingMode lightingMode) {
```

scene/realistic-lighting-and-shadows/src/main/java/com/esri/samples/realistic_lighting_and_shadows/LightingModeListCell.java
```diff
@@ -0,0 +1,37 @@
...
+ */
+public class LightingModeListCell extends ListCell<LightingMode> {
+
+  @Override
+  protected void updateItem(LightingMode mode, boolean empty) {
```
Rachael-E: as above, let's make the following change:
```suggestion
  protected void updateItem(LightingMode lightingMode, boolean empty) {
```

scene/realistic-lighting-and-shadows/src/main/java/com/esri/samples/realistic_lighting_and_shadows/LightingModeListCell.java
```diff
@@ -0,0 +1,37 @@
...
+
+    super.updateItem(mode, empty);
+    if (mode == LightingMode.LIGHT) setText("Sun light only");
+    else if (mode == LightingMode.LIGHT_AND_SHADOWS) setText("Sun light with shadows");
+    else if (mode == LightingMode.NO_LIGHT) setText("No sun light effect");
+
+    else setText("Sun light only");
```
Rachael-E: A couple of things here. 
1. I remembered a sample discussion about the naming in the drop down boxes. I checked with Hudson and he said a decision had been made to go with the enum names for the drop boxes, but in plain english. See suggestion below.

2. With this code the way it is, it's wired up so that when the application is launched, it's setting the combobox to say "Sun light only" without listening in to what `LightingMode` the combobox is actually set to. For example, if for whatever reason the LightingMode came back as `null`, then the text would display "Sun light only" which wouldn't be accurate. 

To get over this, we can simply add a null check in both this class and ComboBoxStringConverter to be on the safe side, e.g.

 ```java
    super.updateItem(lightingMode, empty);
    if (lightingMode != null) {
      if (lightingMode == LightingMode.LIGHT) setText("Light");
      else if (lightingMode == LightingMode.LIGHT_AND_SHADOWS) setText("Light and shadows");
      else if (lightingMode == LightingMode.NO_LIGHT) setText("No light");
    }

```

In addition to this, in the controller, we can add in logic there to select the appropriate enum from the drop down (assuming we want the sample to launch on "no light"): see my suggestion there.

File: scene/realistic-lighting-and-shadows/src/main/java/com/esri/samples/realistic_lighting_and_shadows/RealisticLightingAndShadowsController.java [hunk 67]
```diff
@@ -0,0 +1,157 @@
...
+      comboBox.getSelectionModel().selectedItemProperty().addListener(e -> {
+        sceneView.setSunLighting(comboBox.getSelectionModel().getSelectedItem());
+      });
+
```
Rachael-E, line 106: to ensure the option chosen in the drop box is reflecting the lighting mode shown in the app, we can add the following here (assuming we want the sample to launch on no light) :

```java
      // launch the app with lighting mode set to NO_LIGHT
      comboBox.getSelectionModel().select(0);
```

scene/realistic-lighting-and-shadows/src/main/java/com/esri/samples/realistic_lighting_and_shadows/ComboBoxStringConverter.java
Rachael-E: similar to my comments in LightingModeListCell.java: we should ensure we put in a null check here, and update the strings to the enum plain-english names.

```java

  @Override
  public String toString(LightingMode lightingMode) {
    if (lightingMode != null) {
      if (lightingMode == LightingMode.LIGHT) return "Light";
      else if (lightingMode == LightingMode.LIGHT_AND_SHADOWS) return "Light and shadows";
      else if (lightingMode == LightingMode.NO_LIGHT) return "No light";
    }
    return "";
  }
```

scene/realistic-lighting-and-shadows/src/main/java/com/esri/samples/realistic_lighting_and_shadows/ComboBoxStringConverter.java
```diff
@@ -0,0 +1,40 @@
+/*
```
Rachael-E: Just a tidy up suggestion.

For both this class and the `LightingModeListCell` class: since these have a small amount of content, these can live at the bottom of the `RealisticLightingAndShadowsController` class, as we do in some of our other samples (see e.g. https://github.com/Esri/arcgis-runtime-samples-java/blob/e75be37b2eb83a672f57afc1b1cb64e5549506b3/map/create-and-save-map/src/main/java/com/esri/samples/create_and_save_map/CreateAndSaveMapController.java#L176). 

This ensures our samples remain clean (as you've done a great job doing separating out these converters), but also that there isn't a large amount of classes sitting within the sample package. It also shows our MVC pattern more clearly with the converter classes hidden within the controller class.

Rachael-E (changes requested):
great job with your first sample @sclaridge you've done really well to get your head around the MVC pattern and also building some custom converters! There are a few things to look at/fix up: a number of my comments are related to documentation in the readme/code comments and also some best code practice type things to consider which require minor refactoring. Let me know if you have any questions :)

scene/realistic-lighting-and-shadows/src/main/java/com/esri/samples/realistic_lighting_and_shadows/ComboBoxStringConverter.java
JonLavi: @Rachael-E if this is the way forward, then there should be an issue to change up the samples where these are separated to their own files. Since I've started, all the samples using ListCell have had these in their individual files. 

I personally think that's the correct way to do it, since essencially ListCell is UI code unrelated to the controller, so putting it in it's own class reduces clutter in the controller and puts the Runtime code in the focus of what's going on there. After all, if we could do this easily, we'd probably do this in the FXML file as well.

Either way, there should at least be a clear guidance in the contribution guidelines, I think

Rachael-E: thanks @JonLavi, you're right, we currently don't have a consistent way this is handled in the samples and it would be good to reach a design pattern agreement. Either approach works: it's just a matter of deciding what is the best practice pattern! We should get feedback on that during a Java catch up meeting. I'm of the opinion in the instance where it's a standard converter it is fine to "hide" at the bottom of the controller class (in super basic cases may not even need to be in Class format at all). This then focusses the user on our Controller class and our API. If however there was a more "involved" conversion (such as that of GroupLayer sample) then it merits its own abstracted class.

scene/realistic-lighting-and-shadows/README.metadata.json
```diff
@@ -1,35 +1,38 @@
...
+        "sun",
+        "time",
+        "ArcGISScene",
+        "LightingMode",
+        "SceneView.setSunLighting",
+        "SceneView.setSunTime"
+    ],
+    "redirect_from": [
+        ""
+    ],
```
Rachael-E: it looks like the script has made an array with an empty string here: @JonLavi this was one of the formats that breaks the website. Is there something in the script that needs to be updated to catch this? The correct format to use in the instance of there not being a redirect from is just an empty string:

```suggestion
    "redirect_from": "",
```

scene/realistic-lighting-and-shadows/src/main/java/com/esri/samples/realistic_lighting_and_shadows/RealisticLightingAndShadowsController.java
```diff
@@ -146,6 +148,66 @@ public void updateTimeOfDay() {
...
+    @Override
+    public LightingMode fromString(String string) {
+      return null;
+    }
+  }
+
+  /**
+   * Converts time values to strings to display on the slider.
+   */
+  public class SliderStringConverter extends StringConverter<Double> {
```
Rachael-E: mark as private

scene/realistic-lighting-and-shadows/src/main/java/com/esri/samples/realistic_lighting_and_shadows/RealisticLightingAndShadowsController.java
```diff
@@ -146,6 +148,66 @@ public void updateTimeOfDay() {
...
+        else if (lightingMode == LightingMode.LIGHT_AND_SHADOWS) setText("Light and shadows");
+        else if (lightingMode == LightingMode.NO_LIGHT) setText("No light");
+      }
+    }
+  }
+
+  /**
+   * Converts the LightingMode values to strings to display in the ComboBox.
+   */
+  public class ComboBoxStringConverter extends StringConverter<LightingMode> {
```
Rachael-E: these should be marked as `private` classes so that they are only accessible within the Controller.

scene/realistic-lighting-and-shadows/src/main/java/com/esri/samples/realistic_lighting_and_shadows/RealisticLightingAndShadowsController.java
```diff
@@ -146,6 +148,66 @@ public void updateTimeOfDay() {
     );
   }
 
+  /**
+   * Converts the LightingMode values to strings to display in the open ComboBox.
+   */
+  public class LightingModeListCell extends ListCell<LightingMode> {
```
Rachael-E: mark as private

scene/realistic-lighting-and-shadows/src/main/resources/realistic_lighting_and_shadows/style.css
```diff
@@ -17,10 +17,6 @@ VBox {
 
 .slider {
   -fx-padding: 10;
```
Rachael-E: we're still working out the sample viewer details: I'm not sure if leaving the padding in here will work ok in there. To be on the safe side, let's also move that into FXML (@JonLavi unless you have experience of the padding here having worked before?). Sorry for the uncertainty around this for now: still a work in progress for us to figure out!

Rachael-E (changes requested):
thanks for those edits @sclaridge ! Just a few small changes, the most important one being the redirect string. Looks like we might have to alter the script so it doesn't put out an array with an empty string (we've gotten ourselves into trouble before in the past with that: it breaks the website build!)

scene/realistic-lighting-and-shadows/README.metadata.json
JonLavi: Yep the `"redirect from": [""]` shouldn't be happening and I am not entirely sure why it did.
Possibly because the previous file had `""` for this value, and the script decided to shove it in an array.
I'll investigate when I get a chance.

scene/realistic-lighting-and-shadows/src/main/resources/realistic_lighting_and_shadows/style.css
JonLavi: I haven't noticed this in the SV before, but to be fair it's all new territory, as you said. The easy way would be to leave it in the FXML, yes

Rachael-E (approved):
@sclaridge thanks. this looks good to merge now.

PR #531, opened 2020-07-16
Feature enhancement to Surface Placement

jenmerritt:
Added in the relative to scene surface placement enhancement in the Surface Placement sample. Requirements also included adding a scene to the existing sample, adding a slider for the z-value and updating the readme.

Couple of points:
- I update the z-value in the slider implementation and implemented it consistently with the layout of creating the initial graphics rather than using a loop, be interested to know if that was a reasonable decision in terms of ease of understanding what's going on.

- I made the label text colour blue as per the design, there was some discussion about different colours but I think it's overall readable in the blue. I increased the font and symbol size a bit which helped.

- Line 182 - Wasn't sure if the word 'controls' is standard in the naming of JavaFX boxes or if it had just been used previously. Could rename the radio button box to actually mention it since there's now 2 bits of functionality.

Thanks!


scene/surface-placement/src/main/java/com/esri/samples/surface_placement/SurfacePlacementSample.java
```diff
@@ -18,12 +18,15 @@
 
 import java.util.Arrays;
 
+import com.esri.arcgisruntime.layers.ArcGISSceneLayer;
```
JonLavi: We generally want to have the imports in a specific order, with the Runtime imports coming last.
Luckily we can set up the IDE to take care of that for us!
Theres some details in Simone's recent PR showing how to set it up:
https://github.com/Esri/arcgis-runtime-samples-java/pull/528#discussion_r455145020

scene/surface-placement/src/main/java/com/esri/samples/surface_placement/SurfacePlacementSample.java
```diff
@@ -87,6 +90,11 @@ public void start(Stage stage) {
       surface.getElevationSources().add(elevationSource);
       scene.setBaseSurface(surface);
 
+      // add a scene layer
+      final String buildings = "http://tiles.arcgis.com/tiles/P3ePLMYs2RVChkJx/arcgis/rest/services/Buildings_Brest/SceneServer/layers/0";
+      ArcGISSceneLayer sceneLayer = new ArcGISSceneLayer(buildings);
```
JonLavi: In samples, we generally don't define a string variable for URLs unless they are being re-used, so here you could just write this URL directly inside the method call:
```suggestion
      ArcGISSceneLayer sceneLayer = new ArcGISSceneLayer("http://tiles.arcgis.com/tiles/P3ePLMYs2RVChkJx/arcgis/rest/services/Buildings_Brest/SceneServer/layers/0");
```

scene/surface-placement/src/main/java/com/esri/samples/surface_placement/SurfacePlacementSample.java
```diff
@@ -100,43 +108,55 @@ public void start(Stage stage) {
...
       absoluteOverlay.getSceneProperties().setSurfacePlacement(SurfacePlacement.ABSOLUTE);
 
-      // create point for graphic location
-      Point point = new Point(-4.04, 53.06, 1000, SpatialReferences.getWgs84());
+      GraphicsOverlay relativeToSceneOverlay = new GraphicsOverlay();
+      relativeToSceneOverlay.getSceneProperties().setSurfacePlacement(SurfacePlacement.RELATIVE_TO_SCENE);
+
+      // create points for graphic locations
+      Point surfaceRelatedPoint = new Point(-4.4609257, 48.3903965, 70, SpatialReferences.getWgs84());
+      Point sceneRelatedPoint = new Point(-4.4610562, 48.3902727, 70, SpatialReferences.getWgs84());
```
JonLavi: Although these var names might be more accurate (I don't actuall know!), the rest of the vars in this sample have a slightly different naming convention. So perhaps we should update these to match the rest (e.g.`relativeToSceneOverlay`). I'll leave the final word here to @Rachael-E , she's the nomenclature expert.

```suggestion
      Point relativeToSurfacePoint = new Point(-4.4609257, 48.3903965, 70, SpatialReferences.getWgs84());
      Point relativeToScenePoint = new Point(-4.4610562, 48.3902727, 70, SpatialReferences.getWgs84());
```

scene/surface-placement/src/main/java/com/esri/samples/surface_placement/SurfacePlacementSample.java
```diff
@@ -100,43 +108,55 @@ public void start(Stage stage) {
...
 
       // create a red triangle symbol
-      SimpleMarkerSymbol triangleSymbol = new SimpleMarkerSymbol(SimpleMarkerSymbol.Style.TRIANGLE, 0xFFFF0000, 10);
+      SimpleMarkerSymbol triangleSymbol = new SimpleMarkerSymbol(SimpleMarkerSymbol.Style.TRIANGLE, 0xFFFF0000, 12);
 
-      // create a text symbol for each elevation mode
+      // create a text symbol for each surface placement type
```
JonLavi: Good catch on this - the sample was previously called `elevation mode` in Java, but it was renamed `surface placement` to be consistent with the other sdks!

scene/surface-placement/src/main/java/com/esri/samples/surface_placement/SurfacePlacementSample.java
```diff
@@ -168,10 +188,57 @@ public void start(Stage stage) {
...
 
-      // add the scene view and control box to the stack pane
-      stackPane.getChildren().addAll(sceneView, controlsVBox);
+      // create slider to adjust the z-value
+      Slider zValueSlider = new Slider(0, 150, 70);
+      Label zValueSliderLabel = new Label("Z-Value:");
+      zValueSlider.setShowTickMarks(true);
+      zValueSlider.setShowTickLabels(true);
+      zValueSlider.setMajorTickUnit(50);
+      zValueSlider.setOrientation(Orientation.VERTICAL);
```
JonLavi: Since this is getting to be quite a lot of UI setup, you would want to make this sample into an FXML one, and define this slider (as well as the toggle buttons) in the main.fxml file.

scene/surface-placement/src/main/java/com/esri/samples/surface_placement/SurfacePlacementSample.java
```diff
@@ -168,10 +188,57 @@ public void start(Stage stage) {
...
+        drapedFlatOverlay.getGraphics().get(1).setGeometry(newSurfaceRelatedPoint);
+
+        relativeOverlay.getGraphics().get(0).setGeometry(newSurfaceRelatedPoint);
+        relativeOverlay.getGraphics().get(1).setGeometry(newSurfaceRelatedPoint);
+
+        absoluteOverlay.getGraphics().get(0).setGeometry(newSurfaceRelatedPoint);
+        absoluteOverlay.getGraphics().get(1).setGeometry(newSurfaceRelatedPoint);
+
+        relativeToSceneOverlay.getGraphics().get(0).setGeometry(newSceneRelatedPoint);
+        relativeToSceneOverlay.getGraphics().get(1).setGeometry(newSceneRelatedPoint);
```
JonLavi: I think a loop would be fine here, since it's quite clear that we're updating all the graphics in all the graphics overlays with one specific `Point`

scene/surface-placement/src/main/java/com/esri/samples/surface_placement/SurfacePlacementSample.java
```diff
@@ -168,10 +188,57 @@ public void start(Stage stage) {
...
+      zValueSlider.setOrientation(Orientation.VERTICAL);
+
+      zValueSlider.valueProperty().addListener((observableValue, oldValue, newValue) -> {
+
+        // get z-value from slider
+        Double zValueFromSlider = zValueSlider.getValue();
+
+        // create new points with updated z-value
+        Point newSurfaceRelatedPoint = new Point(surfaceRelatedPoint.getX(), surfaceRelatedPoint.getY(), zValueFromSlider, surfaceRelatedPoint.getSpatialReference());
+        Point newSceneRelatedPoint = new Point(sceneRelatedPoint.getX(), sceneRelatedPoint.getY(), zValueFromSlider, sceneRelatedPoint.getSpatialReference());
```
JonLavi: Thinking from a performance point of view (which might not entirely be in the spirit of samples):
Every time one of these slider event fires, these two lines are repeating the `.getX()`, `.getY()` and `.getSpatialReference()` calls. We know these aren't likely to change. The only thing we really change is the Z-value. So to save ourselves doing that, we could just save these values into variables outside of this event listener, and refer to them.

What you definitely can do is save the spatial reference as a varialbe outside of this lambda, and re-use it, since it'll be the same for both points

I'll leave it again to @Rachael-E to decide whether this should be updated, or left as is for user clarity. Either way, might be a helpful observation regarding not repeating method calls?

scene/surface-placement/src/main/java/com/esri/samples/surface_placement/SurfacePlacementSample.java
```diff
@@ -168,10 +188,57 @@ public void start(Stage stage) {
...
+
+      // create a controls area for the z-value slider
+      VBox sliderVBox = new VBox();
+      sliderVBox.setBackground(
+        new Background(new BackgroundFill(Paint.valueOf("rgba(0,0,0,0.5)"), CornerRadii.EMPTY, Insets.EMPTY)));
+      sliderVBox.setPadding(new Insets(10.0));
+      sliderVBox.setMaxSize(70, 200);
+      sliderVBox.getStyleClass().add("panel-region");
+      sliderVBox.getChildren()
+        .addAll(zValueSliderLabel, zValueSlider);
```
JonLavi: There's generally no need to 'chop' the method chaining if the line doesn't get too long 😄

scene/surface-placement/src/main/java/com/esri/samples/surface_placement/SurfacePlacementSample.java
```diff
@@ -168,10 +188,57 @@ public void start(Stage stage) {
...
+
+      });
+
+      // create a controls area for the z-value slider
+      VBox sliderVBox = new VBox();
+      sliderVBox.setBackground(
+        new Background(new BackgroundFill(Paint.valueOf("rgba(0,0,0,0.5)"), CornerRadii.EMPTY, Insets.EMPTY)));
+      sliderVBox.setPadding(new Insets(10.0));
+      sliderVBox.setMaxSize(70, 200);
+      sliderVBox.getStyleClass().add("panel-region");
```
JonLavi: You can define a lot of this stuff in the CSS file itself, so that it's easier for the users to focus just on the control elements.
Once you do `sliderVBox.getStyleClass().add("panel-region");`, all the CSS rules targeting this class would apply, and then you could just apply the style class to both the slider and the radio-button vboxes.

Check out these files from animate images with image overlay for some inspiration:

https://github.com/Esri/arcgis-runtime-samples-java/blob/master/scene/animate-images-with-image-overlay/src/main/resources/animate_images_with_image_overlay/main.fxml

You could also extract the styling from the VBoxes, Slider, and future RadioButtons.

scene/surface-placement/build.gradle
```diff
@@ -6,7 +6,7 @@ plugins {
 group = 'com.esri.samples'
 
 ext {
-    arcgisVersion = '100.8.0'
+    arcgisVersion = '100.9.0-2840'
 }
 
 javafx {
```
JonLavi, line 12: While we have this sample open, let's update the JavaFX version to `11.0.2`.
You'll also need to include the module `'javafx.fxml'` when you make this into an fxml sample.
Lastly, on line 77, lets update:
`gradleVersion = '6.5.1'`
You'll also have to change this version number in a line inside `\gradle\wrapper\gradle-wrapper.properties`

scene/surface-placement/src/main/java/com/esri/samples/surface_placement/SurfacePlacementSample.java
```diff
@@ -168,10 +188,57 @@ public void start(Stage stage) {
...
+      Label zValueSliderLabel = new Label("Z-Value:");
+      zValueSlider.setShowTickMarks(true);
+      zValueSlider.setShowTickLabels(true);
+      zValueSlider.setMajorTickUnit(50);
+      zValueSlider.setOrientation(Orientation.VERTICAL);
+
+      zValueSlider.valueProperty().addListener((observableValue, oldValue, newValue) -> {
+
+        // get z-value from slider
+        Double zValueFromSlider = zValueSlider.getValue();
```
JonLavi: You can get away with using `double`, the primitive type here:

```suggestion
        double zValueFromSlider = zValueSlider.getValue();
```

The IDE does a better job at explaining why than I would ever do:

> The Double class wraps a value of the primitive type double in an object. An object of type Double contains a single field whose type is double.
> In addition, this class provides several methods for converting a double to a String and a String to a double, as well as other constants and methods useful when dealing with a double.

JonLavi (commented):
Nice work Jen! The functionality is all good, and the updates for the readme as well. I can see on the screenshot that the `ArcGISSceneLayer` loads fine for you, but I didn't have any luck on my machine. Might be a VPN problem, or temporary service problem, so I'll try again in the next review round.

>    I update the z-value in the slider implementation and implemented it consistently with the layout of creating the initial graphics rather than using a loop, be interested to know if that was a reasonable decision in terms of ease of understanding what's going on.

I think a loop would be fine here, since it's quite clear that we're updating all the graphics in all the graphics overlays with one specific `Point`. The comment you have in there right now does a perfect job of outlining what would happen inside a loop.

>    I made the label text colour blue as per the design, there was some discussion about different colours but I think it's overall readable in the blue. I increased the font and symbol size a bit which helped.

That looks great, no reason not to increase the text size!

>    Line 182 - Wasn't sure if the word 'controls' is standard in the naming of JavaFX boxes or if it had just been used previously. Could rename the radio button box to actually mention it since there's now 2 bits of functionality.

Not sure if it's a JavaFX convention, but we generally name the V/Hboxes where we collect different UI elements `controls_`, so I'd agree that it would be good to update it. However, because there is currently so much UI setup in this sample, it's probably even wiser to turn it into an FXML sample.

scene/surface-placement/src/main/java/com/esri/samples/surface_placement/SurfacePlacementSample.java
jenmerritt: Have left this for Rachael to comment when she reviews.

scene/surface-placement/src/main/java/com/esri/samples/surface_placement/SurfacePlacementSample.java
jenmerritt: This is probably resolved via the loop that's been implemented in the latest commits.

scene/surface-placement/build.gradle
jenmerritt: I updated these versions - if you could double check I've done things in the right place, thanks!

JonLavi: Yep this is perfect 👍

scene/surface-placement/src/main/java/com/esri/samples/surface_placement/SurfacePlacementController.java
```diff
@@ -0,0 +1,203 @@
...
+
+      absoluteOverlay = new GraphicsOverlay();
+      absoluteOverlay.getSceneProperties().setSurfacePlacement(LayerSceneProperties.SurfacePlacement.ABSOLUTE);
+
+      relativeToSceneOverlay = new GraphicsOverlay();
+      relativeToSceneOverlay.getSceneProperties().setSurfacePlacement(LayerSceneProperties.SurfacePlacement.RELATIVE_TO_SCENE);
+
+      // create points for graphic locations
+      surfaceRelatedPoint = new Point(-4.4609257, 48.3903965, 70, SpatialReferences.getWgs84());
+
```
JonLavi: There's no need for the empty line here I think

scene/surface-placement/src/main/java/com/esri/samples/surface_placement/SurfacePlacementController.java
```diff
@@ -0,0 +1,203 @@
...
+      relativeToSceneOverlay.getGraphics().add(new Graphic(sceneRelatedPoint, relativeToSceneText));
+
+      // add graphics overlays to the scene view
+      sceneView.getGraphicsOverlays().addAll(Arrays.asList(drapedBillboardedOverlay, relativeOverlay, absoluteOverlay, relativeToSceneOverlay));
+
+      // link up the ToggleGroup RadioButtons to the corresponding draped overlay
+      drapedBillboardedRadioButton.setUserData(drapedBillboardedOverlay);
+      drapedFlatRadioButton.setUserData(drapedFlatOverlay);
+
+      // listener for toggles between billboarded and flat draped surface placement modes on the JavaFX ToggleGroup
```
JonLavi: Maybe we can rephrase this comment a bit to describe the action that the code performs:
```suggestion
      // add a listener to the toggle group to switch between billboarded and flat draped surface placement modes
```

scene/surface-placement/src/main/java/com/esri/samples/surface_placement/SurfacePlacementController.java
```diff
@@ -0,0 +1,203 @@
...
+      toggleGroup.selectedToggleProperty().addListener((observableValue, oldToggle, newToggle) -> {
+        if (toggleGroup.getSelectedToggle() != null) {
+          sceneView.getGraphicsOverlays().remove(oldToggle.getUserData());
```
JonLavi, line 113: Maybe it'll be safer to `clear` all the graphics overlays here, so that we're essentally just zapping everything that's visible, and not relying on the `oldToggle` property.

```suggestion
          sceneView.getGraphicsOverlays().clear();
```

scene/surface-placement/src/main/java/com/esri/samples/surface_placement/SurfacePlacementController.java
```diff
@@ -0,0 +1,203 @@
...
+
+    } catch (Exception e) {
+      // on any exception, print the stack trace
+      e.printStackTrace();
+    }
+
+  }
+
+  /**
+   * Sets the Z-Value to the value selected by the JavaFX slider.
```
JonLavi: We could just say 'slider' here I think

```suggestion
   * Sets the Z-Value to the value selected by the slider.
```

scene/surface-placement/src/main/java/com/esri/samples/surface_placement/SurfacePlacementController.java
```diff
@@ -0,0 +1,203 @@
...
+   */
+
```
JonLavi: No need for this empty line
```suggestion
```

scene/surface-placement/src/main/java/com/esri/samples/surface_placement/SurfacePlacementController.java
```diff
@@ -0,0 +1,203 @@
...
+  @FXML
+  private void changeZValue(){
+        // get z-value from slider
```
JonLavi: We generally try and have the comments in plain english, so lets add a `the`:

```suggestion
        // get the z-value from slider
```

scene/surface-placement/src/main/java/com/esri/samples/surface_placement/SurfacePlacementController.java
```diff
@@ -0,0 +1,203 @@
...
+        double zValueFromSlider = zValueSlider.getValue();
+
+        // create new points with updated z-value
```
JonLavi: As above, no harm in adding an article, I think:

```suggestion
        // create new points with the updated z-value
```

scene/surface-placement/src/main/java/com/esri/samples/surface_placement/SurfacePlacementController.java
```diff
@@ -0,0 +1,199 @@
...
+
+  }
+
+  /**
+   * Sets the Z-Value to the value selected by the JavaFX slider.
+   */
+
+  @FXML
+  private void changeZValue(){
+    // get z-value from slider
```
JonLavi: Lets add some articles here so that the comments are plain english:
```suggestion
    // get the z-value from the slider
```

scene/surface-placement/src/main/java/com/esri/samples/surface_placement/SurfacePlacementController.java
```diff
@@ -0,0 +1,199 @@
...
+    double zValueFromSlider = zValueSlider.getValue();
```
JonLavi: I think just zValue would be enough after we got it (there's no other source for a zValue after all):

```suggestion
    double zValue = zValueSlider.getValue();
```

scene/surface-placement/src/main/java/com/esri/samples/surface_placement/SurfacePlacementController.java
```diff
@@ -0,0 +1,199 @@
...
+
+    // update the geometry of each of the existing graphics to include the new z-value
+    for(int i = 0; i < sceneView.getGraphicsOverlays().size(); i++){
```
JonLavi: You could use the 'smart' loop syntax for this if you know you're going to make an operation on each item in the list:
```suggestion
    for (GraphicsOverlay graphicsOverlay : sceneView.getGraphicsOverlays().size()) {
```
Also mind to have a space between the `for` keyword and the parentheses, as well as before the `{`

The same would apply to the inner loop.

You'd only really want to use the `i++` type of loop if you know you might have a unique condition for starting/ending/iterating over your array (e.g. starting with `i=2`, or going until `i < size -1`).

scene/surface-placement/src/main/resources/surface_placement/main.fxml
```diff
@@ -0,0 +1,52 @@
...
+        <RadioButton toggleGroup="$toggleGroup" fx:id="drapedBillboardedRadioButton" text="Draped Billboarded" selected="true" />
+        <RadioButton toggleGroup="$toggleGroup" fx:id="drapedFlatRadioButton" text="Draped Flat" />
+    </VBox>
+    <!--Slider controls for z-value-->
+    <VBox StackPane.alignment="TOP_RIGHT" styleClass="panel-region, slider-controls" >
+        <StackPane.margin>
+            <Insets topRightBottomLeft="10"/>
+        </StackPane.margin>
+        <Label fx:id="zValueSliderLabel" text="Z-Value: " />
+        <Slider fx:id="zValueSlider" min="0" max="150" value="70" onMouseDragged="#changeZValue" onMouseReleased="#changeZValue"  />
```
JonLavi: Nice work on this file, just a sneaky extra space here:

```suggestion
        <Slider fx:id="zValueSlider" min="0" max="150" value="70" onMouseDragged="#changeZValue" onMouseReleased="#changeZValue" />
```

scene/surface-placement/src/main/resources/surface_placement/style.css
```diff
@@ -1,7 +1,38 @@
...
+  -fx-pref-height:160;
+  -fx-show-tick-marks: true;
+  -fx-show-tick-labels: true;
+  -fx-major-tick-unit: 50;
+  -fx-minor-tick-count: 1;
+}
+
+.slider-controls .slider .axis {
+  -fx-tick-label-fill: white;
 }
```
JonLavi: GitHub is telling us there's an extra space missing here 😄

scene/surface-placement/src/main/java/com/esri/samples/surface_placement/SurfacePlacementController.java
```diff
@@ -0,0 +1,199 @@
...
+  @FXML private RadioButton drapedBillboardedRadioButton;
+  @FXML private RadioButton drapedFlatRadioButton;
+  private Surface surface;
+  private Point surfaceRelatedPoint;
+  private Point sceneRelatedPoint;
+  private GraphicsOverlay drapedBillboardedOverlay;
+  private GraphicsOverlay drapedFlatOverlay;
+  private GraphicsOverlay absoluteOverlay;
+  private GraphicsOverlay relativeOverlay;
+  private GraphicsOverlay relativeToSceneOverlay;
```
JonLavi: I think most of these variables could be local variables instead of class members, so no need to define them here.

JonLavi (commented):
Looks really good @jenmerritt !
Just a few more small suggestions for refactoring, but I think once that's fixed, it should be good to go.

scene/surface-placement/src/main/java/com/esri/samples/surface_placement/SurfacePlacementController.java
jenmerritt: I tried with clear but it then removes the other overlays too. Shall I leave with the `oldToggle`, or I could do an `addAll()` on the next line that includes the other graphics.

JonLavi: You're absolutely right, I made a mistake there! Forgot that there's a bunch of other GraphicsOverlays, not just switching between 2. It was correct the way it was before, sorry!

scene/surface-placement/src/main/java/com/esri/samples/surface_placement/SurfacePlacementController.java
jenmerritt: Nice, another good side effect of adding the loop in. Much cleaner there now, thanks!

scene/surface-placement/src/main/java/com/esri/samples/surface_placement/SurfacePlacementController.java
```diff
@@ -0,0 +1,187 @@
...
+  @FXML private SceneView sceneView;
+  @FXML private Slider zValueSlider;
+  @FXML private ToggleGroup toggleGroup;
+  @FXML private RadioButton drapedBillboardedRadioButton;
+  @FXML private RadioButton drapedFlatRadioButton;
+
+  public void initialize() {
+
+    try {
+
```
JonLavi: We can remove the extra spaces before and after the `try`

scene/surface-placement/src/main/java/com/esri/samples/surface_placement/SurfacePlacementController.java
```diff
@@ -0,0 +1,187 @@
...
+        if (toggleGroup.getSelectedToggle() != null) {
+          sceneView.getGraphicsOverlays().remove(oldToggle.getUserData());
+          sceneView.getGraphicsOverlays().add((GraphicsOverlay) toggleGroup.getSelectedToggle().getUserData());
+        }
+      });
+
+    } catch (Exception e) {
+      // on any exception, print the stack trace
+      e.printStackTrace();
+    }
```
JonLavi, line 120: We can also remove the extra spaces before and after this `catch` block`

scene/surface-placement/src/main/resources/surface_placement/style.css
```diff
@@ -1,7 +1,38 @@
...
+  -fx-show-tick-labels: true;
+  -fx-major-tick-unit: 50;
+  -fx-minor-tick-count: 1;
 }
 
-.panel-region .radio-button {
-   -fx-text-fill: white;
+.slider-controls .slider .axis {
+  -fx-tick-label-fill: white;
 }
```
JonLavi: Looks like the new line is still missing :(

JonLavi (approved):
Nice work @jenmerritt , looks good to go now!
Just a couple of blank lines that can be removed.

scene/surface-placement/README.metadata.json
```diff
@@ -33,7 +33,10 @@
         "Surface"
     ],
     "snippets": [
-        "src/main/java/com/esri/samples/surface_placement/SurfacePlacementSample.java"
+        "src/main/java/com/esri/samples/surface_placement/SurfacePlacementController.java",
+        "src/main/java/com/esri/samples/surface_placement/SurfacePlacementLauncher.java",
```
Rachael-E: These snippets feed into the code snippets which appear on our sample webpages. We don't show the launcher class on the snippets on the webpages, since it's just a reference to the 'main' method and doesn't show our API or method calls and how they link into the UI. So we can remove that class that from the list here.
```suggestion
```

scene/surface-placement/src/main/java/com/esri/samples/surface_placement/SurfacePlacementController.java
```diff
@@ -0,0 +1,183 @@
...
+  @FXML private ToggleGroup toggleGroup;
+  @FXML private RadioButton drapedBillboardedRadioButton;
+  @FXML private RadioButton drapedFlatRadioButton;
+
+  public void initialize() {
+    try {
+      // create a scene and add a basemap to it
+      ArcGISScene scene = new ArcGISScene(Basemap.Type.IMAGERY);
+
+      // add the SceneView to the stack pane
```
Rachael-E: in this step, we `// set the scene to the scene view`: the stack pane is now tucked away in the FXML file :)

scene/surface-placement/src/main/resources/surface_placement/style.css
```diff
@@ -1,7 +1,38 @@
...
+  -fx-max-height: 120;
+  -fx-spacing: 6;
+}
+
+.toggle-controls .radio-button {
+  -fx-text-fill: white;
+}
+
+.slider-controls {
+  -fx-max-width: 90;
```
Rachael-E: when viewed on a Mac, this width slightly truncates the `Z-Value:` label (even though it looks fine on Windows!). A value of 95 however corrects it. 
```suggestion
  -fx-max-width: 95;
```

scene/surface-placement/src/main/java/com/esri/samples/surface_placement/SurfacePlacementController.java
```diff
@@ -0,0 +1,183 @@
...
+        "https://elevation3d.arcgis.com/arcgis/rest/services/WorldElevation3D/Terrain3D/ImageServer");
+      surface.getElevationSources().add(elevationSource);
+      scene.setBaseSurface(surface);
+
+      // add a scene layer
+      ArcGISSceneLayer sceneLayer = new
+        ArcGISSceneLayer("http://tiles.arcgis.com/tiles/P3ePLMYs2RVChkJx/arcgis/rest/services/Buildings_Brest/SceneServer/layers/0");
+      scene.getOperationalLayers().add(sceneLayer);
+
+      // add a camera and initial camera position
```
Rachael-E: just to make this fit the code a little better, I'd suggest the following edit:
```suggestion
      // set an initial viewpoint camera position to the scene view
```

scene/surface-placement/src/main/java/com/esri/samples/surface_placement/SurfacePlacementController.java
```diff
@@ -0,0 +1,183 @@
...
+      sceneView.setArcGISScene(scene);
+
+      // add base surface for elevation data
+      Surface surface = new Surface();
+      ArcGISTiledElevationSource elevationSource = new ArcGISTiledElevationSource(
+        "https://elevation3d.arcgis.com/arcgis/rest/services/WorldElevation3D/Terrain3D/ImageServer");
+      surface.getElevationSources().add(elevationSource);
+      scene.setBaseSurface(surface);
+
+      // add a scene layer
```
Rachael-E: just to make it clear that the buildings are sourced from the scene layer, suggested edit:
```suggestion
      // add buildings to the scene with a scene layer URI
```

scene/surface-placement/src/main/java/com/esri/samples/surface_placement/SurfacePlacementController.java
```diff
@@ -0,0 +1,183 @@
...
+  /**
+   * Sets the Z-Value to the value selected by the slider.
+   */
+  @FXML
+  private void changeZValue(){
+    // get the z-value from the slider
+    double zValue = zValueSlider.getValue();
+
+    // update the geometry of each of the existing graphics to include the new z-value
+    for (GraphicsOverlay graphicsOverlay : sceneView.getGraphicsOverlays()) {
```
Rachael-E: We can replace this nested for loop with a more modern approach using Java 8's stream API (see below). Generally if you come across an instance that requires a nested for loop, you can exchange them using streams.

```java
    sceneView.getGraphicsOverlays().forEach(graphicsOverlay ->
      graphicsOverlay.getGraphics().stream().filter(
        graphic -> graphic.getGeometry().getGeometryType() == GeometryType.POINT).forEach(graphic -> {
        Geometry geometry = graphic.getGeometry();
        Point currentPoint = (Point) geometry;
        Point updatedPoint = new Point(currentPoint.getX(), currentPoint.getY(), zValue, currentPoint.getSpatialReference());
        graphic.setGeometry(updatedPoint);
      }));
```

Rachael-E (changes requested):
@jenmerritt great job on your new sample! This has really expanded out the sample from its original format and you've got your head around Graphics Overlays and the MVC pattern too, nice work! Because the sample expands on what was there before, there are instances of repeated code that we can tidy up with a few methods, whilst still highlighting the main function of the sample (surface placement mode). 

One of repeated code instances I noticed was that of creating Text Symbols: We could abstract that into a method with the following:

```java
 /**
   * Creates a new TextSymbol from a string used to identify a surface placement type.
   * @param text string to be used as text within the TextSymbol constructor
   * @return a new TextSymbol
   *
   */
  private TextSymbol createTextSymbol(String text) {
    TextSymbol textSymbol = new TextSymbol(14, text, 0xFF0000FF,
      TextSymbol.HorizontalAlignment.LEFT, TextSymbol.VerticalAlignment.MIDDLE);
    textSymbol.setOffsetX(20);

    return textSymbol;
  }
```

and then refer to this method after setting the viewpoint camera with (altering the relativeToSceneText offset and alignment manually: or we could put an if statement in the method to say if the text is equal to "RELATIVE TO SCENE" then change the offset and alignment, that might be better):

```java
      // create a text symbol for each surface placement type
      TextSymbol drapedBillboardedText = createTextSymbol("DRAPED BILLBOARDED");
      TextSymbol drapedFlatText = createTextSymbol("DRAPED FLAT");
      TextSymbol relativeText = createTextSymbol("RELATIVE");
      TextSymbol absoluteText = createTextSymbol("ABSOLUTE");
      TextSymbol relativeToSceneText = createTextSymbol("RELATIVE TO SCENE");
      relativeToSceneText.setOffsetX(-20);
      relativeToSceneText.setHorizontalAlignment(TextSymbol.HorizontalAlignment.RIGHT);
```

In addition, adding graphics to the graphics overlays was repeated multiple times with different graphics overlays and textSymbols. Since this sample isn't to demonstrate adding graphics to graphic overlays, we can hide that in a method too with the following:

```java
/**
   * Creates a graphic from a Point, SimpleMarkerSymbol, and TextSymbol and adds it to a GraphicsOverlay.
   *
   * @param graphicsOverlay the graphics overlay to add graphics to
   * @param textSymbol the text symbol to create a graphic from
   */
  private void addGraphicsToGraphicsOverlay(GraphicsOverlay graphicsOverlay, TextSymbol textSymbol) {

    // create points for graphic locations
    Point sceneRelatedPoint = new Point(-4.4610562, 48.3902727, 70, SpatialReferences.getWgs84());
    Point surfaceRelatedPoint = new Point(-4.4609257, 48.3903965, 70, SpatialReferences.getWgs84());
    // create a red triangle symbol
    SimpleMarkerSymbol triangleSymbol = new SimpleMarkerSymbol(SimpleMarkerSymbol.Style.TRIANGLE, 0xFFFF0000, 12);

    if (graphicsOverlay.getSceneProperties().getSurfacePlacement() != LayerSceneProperties.SurfacePlacement.RELATIVE_TO_SCENE) {
      graphicsOverlay.getGraphics().add(new Graphic(surfaceRelatedPoint, triangleSymbol));
      graphicsOverlay.getGraphics().add(new Graphic(surfaceRelatedPoint, textSymbol));
    } else {
      graphicsOverlay.getGraphics().add(new Graphic(sceneRelatedPoint, triangleSymbol));
      graphicsOverlay.getGraphics().add(new Graphic(sceneRelatedPoint, textSymbol));
    }
  }
```

and then after the textSymbol set up, call like the following which highlights the main purpose of this sample. This also has the benefit of focusing the user on the "journey" of setting up one graphics overlay, settings its surface placement, and adding the graphics to the graphics overlay. Then repeat those steps in order for the other graphics overlay. 

```java
      // create graphics overlays with surface placement types, and add graphics to them
      GraphicsOverlay drapedBillboardedOverlay = new GraphicsOverlay();
      drapedBillboardedOverlay.getSceneProperties().setSurfacePlacement(LayerSceneProperties.SurfacePlacement.DRAPED_BILLBOARDED);
      addGraphicsToGraphicsOverlay(drapedBillboardedOverlay, drapedBillboardedText);
```
Please just let me know if you have any questions about this refactoring: it looks like a lot but saves on number of lines of code without losing any of the good work that you did.

scene/surface-placement/src/main/java/com/esri/samples/surface_placement/SurfacePlacementSample.java
Rachael-E: sorry @jenmerritt I just saw this. Let's go with Jon's suggestion, it makes it more clear what the points are for.

scene/surface-placement/src/main/java/com/esri/samples/surface_placement/SurfacePlacementController.java
```diff
@@ -63,73 +63,41 @@ public void initialize() {
...
-
-      absoluteOverlay.getGraphics().add(new Graphic(surfaceRelatedPoint, triangleSymbol));
-      absoluteOverlay.getGraphics().add(new Graphic(surfaceRelatedPoint, absoluteText));
-
-      relativeToSceneOverlay.getGraphics().add(new Graphic(sceneRelatedPoint, triangleSymbol));
-      relativeToSceneOverlay.getGraphics().add(new Graphic(sceneRelatedPoint, relativeToSceneText));
+      addGraphicsToGraphicsOverlay(relativeToSceneOverlay, relativeToSceneText);
 
       // add graphics overlays to the scene view
       sceneView.getGraphicsOverlays().addAll(Arrays.asList(drapedBillboardedOverlay, relativeOverlay, absoluteOverlay, relativeToSceneOverlay));
```
Rachael-E: I've just noticed that this line goes over our max line length guidance (120 chars). We should split it up to avoid users having to use the horizontal scroll bar on Github (another easy to implement improve readability thing)! 

```java
      // add graphics overlays to the scene view
      sceneView.getGraphicsOverlays().addAll(Arrays.asList(
        drapedBillboardedOverlay, relativeOverlay, absoluteOverlay, relativeToSceneOverlay));
```

scene/surface-placement/src/main/java/com/esri/samples/surface_placement/SurfacePlacementController.java
```diff
@@ -63,73 +63,41 @@ public void initialize() {
       surface.getElevationSources().add(elevationSource);
       scene.setBaseSurface(surface);
 
-      // add a scene layer
+      // add buildings to the scene with a scene layer URI
       ArcGISSceneLayer sceneLayer = new
         ArcGISSceneLayer("http://tiles.arcgis.com/tiles/P3ePLMYs2RVChkJx/arcgis/rest/services/Buildings_Brest/SceneServer/layers/0");
```
Rachael-E: this is also a bit too far over our line length guidance (over by a little bit is ok though). Suggest the following new line to break it up:

```java

      // add buildings to the scene with a scene layer URI
      ArcGISSceneLayer sceneLayer = new ArcGISSceneLayer(
        "http://tiles.arcgis.com/tiles/P3ePLMYs2RVChkJx/arcgis/rest/services/Buildings_Brest/SceneServer/layers/0");
```

scene/surface-placement/src/main/java/com/esri/samples/surface_placement/SurfacePlacementController.java
```diff
@@ -160,15 +128,56 @@ private void changeZValue(){
     double zValue = zValueSlider.getValue();
```
Rachael-E, line 129: I can't comment on line 126, but I just noticed that there should be an extra space between the parenthesis and curly brackets

```java
private void changeZValue() {

```

scene/surface-placement/src/main/java/com/esri/samples/surface_placement/SurfacePlacementController.java
```diff
@@ -160,15 +128,56 @@ private void changeZValue(){
...
-        if (geometry.getGeometryType().equals(GeometryType.POINT)) {
-          Point currentPoint = (Point) geometry;
-          Point updatedPoint = new Point(currentPoint.getX(), currentPoint.getY(), zValue, currentPoint.getSpatialReference());
-          graphic.setGeometry(updatedPoint);
-        }
-      }
+        Point currentPoint = (Point) geometry;
+        Point updatedPoint = new Point(currentPoint.getX(), currentPoint.getY(), zValue, currentPoint.getSpatialReference());
+        graphic.setGeometry(updatedPoint);
+      }));
```
Rachael-E: sorry this was my fault: we should do some breaking up of these brackets here to be more java-y :) 

```java
  /**
   * Sets the Z-Value to the value selected by the slider.
   */
  @FXML
  private void changeZValue() {
    // get the z-value from the slider
    double zValue = zValueSlider.getValue();

    // update the geometry of each of the existing graphics to include the new z-value
    sceneView.getGraphicsOverlays().forEach(graphicsOverlay ->
      graphicsOverlay.getGraphics().stream().filter(
        graphic -> graphic.getGeometry().getGeometryType() == GeometryType.POINT).forEach(graphic -> {
        Geometry geometry = graphic.getGeometry();
        Point currentPoint = (Point) geometry;
        Point updatedPoint = new Point(currentPoint.getX(), currentPoint.getY(), zValue, currentPoint.getSpatialReference());
        graphic.setGeometry(updatedPoint);
      })
    );
  }

```

scene/surface-placement/src/main/java/com/esri/samples/surface_placement/SurfacePlacementController.java
```diff
@@ -160,15 +128,56 @@ private void changeZValue(){
...
+  }
+
+  /**
+   * Creates a new TextSymbol from a string used to identify a surface placement type.
+   * @param text string to be used as text within the TextSymbol constructor
+   * @return a new TextSymbol
+   *
```
Rachael-E: again probably my fault for not spotting this in the suggestion! Can remove this empty javaDoc line. 
```suggestion
```

scene/surface-placement/src/main/java/com/esri/samples/surface_placement/SurfacePlacementController.java
```diff
@@ -160,15 +128,56 @@ private void changeZValue(){
...
+   */
+  private TextSymbol createTextSymbol(String text) {
+
+    TextSymbol textSymbol = new TextSymbol(14, text, 0xFF0000FF,
+      TextSymbol.HorizontalAlignment.LEFT, TextSymbol.VerticalAlignment.MIDDLE);
+    textSymbol.setOffsetX(20);
+
+    if (text == "RELATIVE TO SCENE") {
```
Rachael-E: I'm getting an IDE suggestion here to replace the `==` operator with `.equals()`.  The latter method checks the actual contents of the string, whereas the former operator checks whether the references to the objects are equal.

Rachael-E (approved):
thanks for fixing those up @jenmerritt , just a few other minor things I missed the first time round to tidy up and we're good (including the renaming of the Points as Jon had commented on earlier).  I'll go ahead and approve so as not to hold things up, if you could ensure those changes are made before merging please thanks!

PR #533, opened 2020-07-21
Revert "Fix BufferList sample bug (#525)"

JonLavi:
This reverts commit f64b1ecff7e86967b85a911eb04e3276baffd6f9.

We've disucssed internally that the issue with this sample is related to an issue in Java, so there will be no need to make changes to the sample itself. 
So this PR is to undo the changes, in expectation of the API fix.


Rachael-E:
@JonLavi after our internal discussion alongside with Trevor's agreement to leave this as is based on improved readability, I'll close this PR. Thanks for keeping on top of these issues!

PR #536, opened 2020-07-24
Fix `display utility associations` metadata file

JonLavi:
Hi @Rachael-E,
I've cherry picked the commits we merged into `dev` yesterday to fix this metadata file.
This PR is to apply these changes to `master` as well, to make sure that the SampleViewer and dev page get a correct metadata file if they happen to be build from Master.
Again the diff shows everything is new, but this has to do with the way the script works. Some issue with encoding, I believe.
The changes are:
- fixes `"keywords"` by adding a previously missing comma
- fixes the path in `"snippets":`
- fixes the title, changing it to plural `associations`
- adds the missing `redirect_from` key

### DIFF
--- a/utility_network/display-utility-associations/README.metadata.json
+++ b/utility_network/display-utility-associations/README.metadata.json
@@ -1,30 +1,31 @@
 {
-  "category": "Utility network",
-  "description": "Create graphics for utility associations in a utility network.",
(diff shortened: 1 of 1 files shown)


Rachael-E (approved):
thanks for keeping up with these @JonLavi !

PR #537, opened 2020-07-24
Display Utility Associations - move styling from CSS into main java class

JonLavi:
Hi @Rachael-E, another update for this sample, as the CSS rules break the sample within the SampleViewer.
All I've done is move the offending rules from the CSS file into the `.java` file

### DIFF
--- a/utility_network/display-utility-associations/src/main/java/com/esri/samples/display_utility_associations/DisplayUtilityAssociationsSample.java
+++ b/utility_network/display-utility-associations/src/main/java/com/esri/samples/display_utility_associations/DisplayUtilityAssociationsSample.java
@@ -62,8 +62,7 @@ public class DisplayUtilityAssociationsSample extends Application {
       // create stack pane and application scene
       StackPane stackPane = new StackPane();
       Scene scene = new Scene(stackPane);
-      scene.getStylesheets().add(getClass().getResource("/display_utility_associations/style.css").toExternalForm());
-      
+
       // set title, size, and add scene to stage
       stage.setTitle("Display Utility Associations Sample");
       stage.setWidth(800);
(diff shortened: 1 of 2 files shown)


utility_network/display-utility-associations/src/main/java/com/esri/samples/display_utility_associations/DisplayUtilityAssociationsSample.java
```diff
@@ -142,6 +143,10 @@ else if (networkSource.getSourceType() == UtilityNetworkSource.Type.JUNCTION) {
       gridPane.add(attachmentLabel, 1, 0);
       gridPane.add(connectivityImageView, 0, 1);
       gridPane.add(connectivityLabel, 1, 1);
+      gridPane.setMaxWidth(80);
+      gridPane.setMaxHeight(100);
+      gridPane.setBackground(new Background(new BackgroundFill(Paint.valueOf("rgba(0,0,0,0.3)"), CornerRadii.EMPTY,
```
Rachael-E: these should match the original settings from the offending css: width and height need switched around and for the background to the white transparent not black. 

```suggestion
      gridPane.setMaxWidth(100);
      gridPane.setMaxHeight(80);
      gridPane.setBackground(new Background(new BackgroundFill(Paint.valueOf("rgba(255,255,255,0.3)"), CornerRadii.EMPTY,
```

Rachael-E (approved):
I'll approve since I'm away next week and don't want to hold things up, but I noticed that the settings were copied over incorrectly: I've marked them as they should be in the suggestion. Please make sure those are updated before merging, thanks! 

For my learning and future reference: do the `-fx-padding`, `-fx-vgap` and `-fx-alignment` settings work ok in the sample viewer (the ones still left in .css)?

JonLavi:
> 
> For my learning and future reference: do the `-fx-padding`, `-fx-vgap` and `-fx-alignment` settings work ok in the sample viewer (the ones still left in .css)?

`-fx-alignment`, -fx-padding`, and `-fx-vgap` definitely work in the CSS file when the sample is run in isolation, but they don't seem to have an effect in the SV. I've had to double check and have them open side by side. The difference is miniscule, but the fact that it doesn't work in the SV is a bit of a deal breaker. So I've now done all the styling in the java file, and removed the CSS file alltogether. We'll have to do a deeper investigation about how to get CSS files to catch in the SV at some point

JonLavi:
Note that the recent sample `animate images with image overlay` uses these CSS rules successfuly with the sample viewer, so there's something fishy going on particularly with this sample here

PR #541, opened 2020-07-29
Fix realistic lighting and shadows metadata

sclaridge:
Removed the duplicate main.fxml snippet in README.metadata.json.


JonLavi (approved):
Thanks for fixing that @sclaridge . The sample also works fine with the sample viewer for me now 👍

jenmerritt (approved):

PR #542, opened 2020-08-03
Add Warning to Display Layer View State Sample

jenmerritt:
Summary of additional functionality added to existing Display Layer View State Sample:
- New data
- Added implementation of new WARNING enum for two scenarios - when network fails so cannot request feature layer, or when too many features are requested. Update 9 functionality.
- Added show/hide buttons to control layer visibility to help with showing the NOT_VISIBLE status

For reference, Android implementation was original design:
https://github.com/Esri/arcgis-runtime-samples-android/tree/develop/java/display-layer-view-state


map_view/display-layer-view-state/README.md
```diff
@@ -6,33 +6,33 @@ Determine if a layer is currently being viewed.
 
 ## Use case
 
-The view state includes information on the loading state of layers and whether layers are visible at a given scale. For example, you might change how a layer is displayed in a layer list to communicate whether it is being viewed in the map. You then might show a loading spinner next to its name when the view state is LOADING, gray out the name when NOT_VISIBLE or OUT_OF_SCALE, show the name normally when ACTIVE, or with an error icon when the state is ERROR.
+The view status includes information on the loading state of layers and whether layers are visible at a given scale. You might change how a layer is displayed in a layer list to communicate whether it is being viewed in the map. For example, you could show a loading spinner next to its name when the view status is `LOADING`, grey out the name when `NOT_VISIBLE` or `OUT_OF_SCALE`, show the name normally when `ACTIVE`, or with a warning or error icon when the status is `WARNING` or `ERROR`.
```
JonLavi, line 9: I'm a bit puzzled why we're using `view status` in this description, but the sample is called `view state`. Just pointing it out, no need to change anything since this looks like what's in the design.

map_view/display-layer-view-state/README.md
```diff
@@ -6,33 +6,33 @@ Determine if a layer is currently being viewed.
...
 
 ## How to use the sample
 
-Pan and zoom around in the map. Each layer's view status is displayed. Notice that some layers configured with a min and max scale change to `OUT_OF_SCALE` at certain scales.
+Click the *Load Layer* button to create a new layer and add it to the map. As you pan and zoom around the map, note how the `LayerViewStatus` flags change; for example, `OUT_OF_SCALE` becomes true when the map is scaled outside of the layer's min and max scale range. Click the *Hide Layer* button to hide the layer and observe the view state change to `NOT_VISIBLE`.
+
+If your device supports airplane mode, you can turn this on and pan around the map to see layers display the `WARNING` status when they cannot online fetch data. Turn airplane mode back off to see the warning disappear.
```
JonLavi: I've suggested this previously in the common design:
Perhaps we should use the term `disconnect from the network` instead of `airplane mode`?
After all, airplane mode turns off bluetooth, gps, etc. as well. We don't have to imply that gps or bluetooth have any effect on this.


Either way I think there's a small mixup of words here (@TADraeseke this typo is live in android - sorry, can't seem to tag Hudson):
```suggestion
If your device supports airplane mode, you can turn this on and pan around the map to see layers display the `WARNING` status when they cannot fetch online data. Turn airplane mode back off to see the warning disappear.
```

map_view/display-layer-view-state/src/main/resources/display_layer_view_state/style.css
```diff
@@ -1,27 +1,18 @@
...
+.panel-label {
+  -fx-text-fill: white;
+  -fx-font-size: 16px;
+  -fx-text-alignment: center;
 }
 
-.color-picker .color-picker-label {
-   -fx-text-fill: black;
+.panel-button {
+  -fx-max-width: 130;
```
JonLavi: Same here, you can make a rule for any button thats inside the style class 'panel-region'. Then there's no need to do `hideLayerButton.getStyleClass().add("panel-button");` etc in the main class.

```suggestion
.panel-region .button {
  -fx-max-width: 130;
```

map_view/display-layer-view-state/src/main/resources/display_layer_view_state/style.css
```diff
@@ -1,27 +1,18 @@
...
-.panel-region .check-box {
-   -fx-text-fill: white;
-}
-
-.panel-region .radio-button {
-   -fx-text-fill: white;
+.panel-label {
+  -fx-text-fill: white;
+  -fx-font-size: 16px;
+  -fx-text-alignment: center;
```
JonLavi: It looks like this might be older code, but since we've got this file open there's a bit of stuff we might be able to improve here. There's no need to specify a unique CSS rule for a label that sits within the `panel-region` which you've defined above.
The old rule in this file will apply to any `Label` sitting within the node that get's the style class `panel-region`.
You can then get rid of `layerViewStatusLabel.getStyleClass().add("panel-label");` in the main code, since there's no need to define a specific class for labels inside the panel anymore.

So you could do:
```suggestion
.panel-region .label {
  -fx-text-fill: white;
  -fx-font-size: 16px;
  -fx-text-alignment: center;
```

map_view/display-layer-view-state/src/main/java/com/esri/samples/display_layer_view_state/DisplayLayerViewStateSample.java
```diff
@@ -72,82 +69,148 @@ public void start(Stage stage) {
...
+        }
+      });
 
-      // this layer will be OUT_OF_SCALE outside of its set min/max scales
-      final ArcGISMapImageLayer imageLayer = new ArcGISMapImageLayer(SERVICE_CENSUS);
-      imageLayer.setMinScale(MIN_SCALE);
-      imageLayer.setMaxScale(MIN_SCALE / 10);
+      // create buttons to toggle the visibility of the feature layer
+      loadLayerButton = new Button("Load Layer");
+      loadLayerButton.getStyleClass().add("panel-button");
```
JonLavi: Maybe instead of having two buttons and alternating which one is visible, we could have a single button and change it's text? When the button is pressed, you can set the visibility status to the opposite of what it currently is. Something like this:
```
// set the layer's visibility state to the opposite of the current state
featureLayer.setVisible(!featureLayer.isVisible());
```

The advantage would be that we could get rid of the smaller helper methods

map_view/display-layer-view-state/src/main/java/com/esri/samples/display_layer_view_state/DisplayLayerViewStateSample.java
```diff
@@ -72,82 +69,148 @@ public void start(Stage stage) {
...
+          featureLayer.setMaxScale(40000000 / 10);
+          // add the feature layer to the map
+          map.getOperationalLayers().add(featureLayer);
+        }
 
-      // creating a layer from a service feature table
-      final ServiceFeatureTable featureTable = new ServiceFeatureTable(SERVICE_RECREATION);
-      final FeatureLayer featureLayer = new FeatureLayer(featureTable);
+        // toggle the visibility of the buttons
+        displayHideButton();
```
JonLavi: Although utility methods are generally considered good practice, for a sample this is probably too small of a functionality to have in a separate sample. A good rule of thumb is that if your method isn't called more than once, it can probably be removed and the code written in place

map_view/display-layer-view-state/src/main/java/com/esri/samples/display_layer_view_state/DisplayLayerViewStateSample.java
```diff
@@ -72,82 +69,148 @@ public void start(Stage stage) {
...
+        if (featureLayer != null && !featureLayer.isVisible()) {
+          featureLayer.setVisible(true);
+        } else {
+          // create a feature layer from a portal item
+          final PortalItem portalItem = new PortalItem(new Portal("https://runtime.maps.arcgis.com/"),
+            "b8f4033069f141729ffb298b7418b653");
+          featureLayer = new FeatureLayer(portalItem, 0);
+          //  set a minimum and a maximum scale for the visibility of the feature layer
+          featureLayer.setMinScale(40000000);
+          featureLayer.setMaxScale(40000000 / 10);
```
JonLavi: The IDE is flagging this up, so looks like it's best to write the number without division. Should be easy enough to tell there's a factor of 10 less here, as the lines are otherwise aligned.
```suggestion
          featureLayer.setMaxScale(4000000);
```

map_view/display-layer-view-state/src/main/java/com/esri/samples/display_layer_view_state/DisplayLayerViewStateSample.java
```diff
@@ -72,82 +69,148 @@ public void start(Stage stage) {
...
+      // set the initial viewpoint for the map view
+      mapView.setViewpoint(new Viewpoint(new Point(-11e6, 45e5, SpatialReferences.getWebMercator()), 40000000));
+
       // create a control panel
-      VBox controlsVBox = new VBox(6);
-      controlsVBox.setBackground(new Background(new BackgroundFill(Paint.valueOf("rgba(0,0,0,0.3)"), CornerRadii.EMPTY,
-          Insets.EMPTY)));
-      controlsVBox.setPadding(new Insets(10.0));
-      controlsVBox.setMaxSize(220, 90);
+      controlsVBox = new VBox();
```
JonLavi: Let's move the creation of the VBox and arranging the UI down to where it's first used, where we set it up within the StackPane. We can also add all the buttons to it down there, to avoid switching context so often. So let's put all of that just before we do:

```
stackPane.getChildren().addAll(mapView, controlsVBox);
```

map_view/display-layer-view-state/src/main/java/com/esri/samples/display_layer_view_state/DisplayLayerViewStateSample.java
```diff
@@ -72,82 +69,148 @@ public void start(Stage stage) {
...
+        }
+        // get the layer's view status and display the status
+        EnumSet<LayerViewStatus> layerViewStatus = statusChangeEvent.getLayerViewStatus();
+        displayViewStateText(layerViewStatus);
+
+        // if there is an error or warning, format and display the message as an alert
+        ArcGISRuntimeException error = statusChangeEvent.getError();
+        if (error != null) {
+          Throwable cause = error.getCause();
+          String message = (cause != null) ? cause.toString() : error.toString();
```
JonLavi: We generally don't put out the whole technical error message in these alerts, but just keep it to plain english. So we could write something along the lines of 'Error getting the layer's view status'

map_view/display-layer-view-state/src/main/java/com/esri/samples/display_layer_view_state/DisplayLayerViewStateSample.java
```diff
@@ -72,82 +69,148 @@ public void start(Stage stage) {
...
-      final ArcGISTiledLayer tiledLayer = new ArcGISTiledLayer(SERVICE_TIME_ZONES);
+        // get the layer whose state has changed
+        Layer layer = statusChangeEvent.getLayer();
+        // only update the status if the layer is the feature layer being tracked
+        if (layer != featureLayer) {
+          return;
+        }
+        // get the layer's view status and display the status
+        EnumSet<LayerViewStatus> layerViewStatus = statusChangeEvent.getLayerViewStatus();
+        displayViewStateText(layerViewStatus);
```
JonLavi: Again, I'd be absolutely happy with putting the code for this method directly here in place, since it's not repeated, and it's quite a main part of the functionality (as we're showing the individual Enum values for the view state)

map_view/display-layer-view-state/src/main/java/com/esri/samples/display_layer_view_state/DisplayLayerViewStateSample.java
```diff
@@ -72,82 +69,148 @@ public void start(Stage stage) {
...
+      hideLayerButton = new Button("Hide Layer");
+      hideLayerButton.getStyleClass().add("panel-button");
+      // initially add load layer button to the control panel
+      controlsVBox.getChildren().add(loadLayerButton);
+
+      // create a listener for clicks on the load layer button
+      loadLayerButton.setOnAction(event -> {
+
+        // if the feature layer already exists and is hidden, toggle it's visibility to visible
+        if (featureLayer != null && !featureLayer.isVisible()) {
```
JonLavi: Although this is a good way to make sure the code doesn't misbehave, I don't think we can actually hit this condition the way that the code is set up? Meaning this button will be gone as soon as we load the layer... Once the layer has been loaded we only ever use the 'other' button(s). Just something to think about, since it might be possible to simplify the code (although admittedly making it a little bit less bulletproof).

JonLavi (commented):
Nice work @jenmerritt .
For the screenshot, it looks as if the UI is incredibly small. Can we resize the screenshot in a way that the UI is legible?

Regarding the code, it would be possible to simplify it a little bit by removing some of the helper methods, and probably also reducing the amount of buttons.

map_view/display-layer-view-state/src/main/java/com/esri/samples/display_layer_view_state/DisplayLayerViewStateSample.java
jenmerritt: On this one, there was a bit of discussion when this was initially implemented for the design. There's a different message depending on what the error is - requesting too many features above the max query count or due to the network issues. So the layer view status itself updates to WARNING or ERROR but there is an additional error message that accompanies it.

The 'true' method in the ternary (`cause.toString`) is the one that prints out as a Java Exception when there is a network issue, the false option (should be `error.getMessage` which I'll fix..) prints out nicely for when there's too many feature requests, according to the design discussions. 

So maybe, I could replace first message to be a manual string instead of the UnknownHostException that displays, and leave the other one as the default message "A layer has requested more features than the service maximum feature count". Thoughts?

JonLavi: That sounds like a good solution, yeah!

map_view/display-layer-view-state/src/main/java/com/esri/samples/display_layer_view_state/DisplayLayerViewStateSample.java
jenmerritt: This whole function has now been altered with the other changes and button changes, and I've altered the wording.

map_view/display-layer-view-state/src/main/java/com/esri/samples/display_layer_view_state/DisplayLayerViewStateSample.java
```diff
@@ -101,60 +93,50 @@ public void start(Stage stage) {
         EnumSet<LayerViewStatus> layerViewStatus = statusChangeEvent.getLayerViewStatus();
         displayViewStateText(layerViewStatus);
 
-        // if there is an error or warning, format and display the message as an alert
+        // if there is an error or warning, format and display a message as an Alert
         ArcGISRuntimeException error = statusChangeEvent.getError();
         if (error != null) {
           Throwable cause = error.getCause();
-          String message = (cause != null) ? cause.toString() : error.toString();
+          String message = (cause != null) ? "Network issue, unable to load layer" : error.getMessage();
```
JonLavi: Do we know certainly that this is would be a network issue? Otherwise we can just say what's wrong (`unable to load layer`)?

map_view/display-layer-view-state/src/main/java/com/esri/samples/display_layer_view_state/DisplayLayerViewStateSample.java
```diff
@@ -101,60 +93,50 @@ public void start(Stage stage) {
...
+      toggleLayerButton = new Button("Load Layer");
 
       // create a listener for clicks on the load layer button
-      loadLayerButton.setOnAction(event -> {
+      toggleLayerButton.setOnAction(event -> {
 
-        // if the feature layer already exists and is hidden, toggle it's visibility to visible
-        if (featureLayer != null && !featureLayer.isVisible()) {
-          featureLayer.setVisible(true);
+        // if the feature layer already exists toggle it's visibility to visible
```
JonLavi: Strictly speaking we're not just toggling this to visible, but toggling the visibility on/off. Also suggesting to add a comma as we have one in the `else` block:


```suggestion
        // if the feature layer already exists, toggle it's visibility
```

map_view/display-layer-view-state/src/main/java/com/esri/samples/display_layer_view_state/DisplayLayerViewStateSample.java
```diff
@@ -101,60 +93,50 @@ public void start(Stage stage) {
...
-        if (featureLayer.isVisible()) {
-          featureLayer.setVisible(false);
-          displayLoadButton();
+        // toggle the text on the button
+        if (toggleLayerButton.getText().equals("Load Layer")){
+          toggleLayerButton.setText("Hide Layer");
         } else {
-          featureLayer.setVisible(true);
-          displayHideButton();
+          toggleLayerButton.setText("Load Layer");
```
JonLavi: Show this now be `show` as the opposite of `hide`? We're only really 'loading' the layer at the very first click, and then just showing/hiding. Unless I'm misunderstanding?

```suggestion
          toggleLayerButton.setText("Show Layer");
```

map_view/display-layer-view-state/src/main/java/com/esri/samples/display_layer_view_state/DisplayLayerViewStateSample.java
```diff
@@ -16,132 +16,166 @@
...
-
-  private static final String SERVICE_TIME_ZONES =
-      "https://sampleserver6.arcgisonline.com/arcgis/rest/services/WorldTimeZones/MapServer";
-  private static final String SERVICE_CENSUS =
-      "https://sampleserver6.arcgisonline.com/arcgis/rest/services/Census/MapServer";
-  private static final String SERVICE_RECREATION =
-      "https://sampleserver6.arcgisonline.com/arcgis/rest/services/Recreation/FeatureServer/0";
+  private FeatureLayer featureLayer;
+  private Label layerViewStatusLabel;
+  private Button toggleLayerButton;
```
JonLavi: The label and button can be local variables, because they are only being used in the `start` method.

map_view/display-layer-view-state/src/main/java/com/esri/samples/display_layer_view_state/DisplayLayerViewStateSample.java
```diff
@@ -16,132 +16,166 @@
...
-      // add the map view and control panel to stack pane
+      // create a control panel and add the label and button
+      VBox controlsVBox = new VBox(6);
+      controlsVBox.setAlignment(Pos.TOP_CENTER);
+      controlsVBox.setBackground(new Background(new BackgroundFill(Paint.valueOf("rgba(0,0,0,0.5)"), CornerRadii.EMPTY,
+        Insets.EMPTY)));
+      controlsVBox.setPadding(new Insets(10.0));
+      controlsVBox.setMaxSize(400, 120);
+      controlsVBox.getStyleClass().add("panel-region");
+      controlsVBox.getChildren().addAll(layerViewStatusLabel, toggleLayerButton);
```
JonLavi: A shame none of these work properly with the sample viewer. Which is quite bizarre since they work fine with some other samples!

map_view/display-layer-view-state/README.md
JonLavi: @jenmerritt unless regression testing has destroyed my ability to comprehend text, I think there's still the small issue of the mixed up words here 😄 Just a reminder

JonLavi (approved):
Looks good to me @jenmerritt !
Just a couple of smaller suggestions regarding the comments, so I'll approve so that you can proceed with asking for a second review.
It's really perplexing to me that some of the CSS rules don't catch in the SV. Especially because it works for others. I believe we've all verified this one individually?

However don't spend too much time playing with it if it hadn't worked previously. Getting this sample merged into the SV for the U9 release would be great (if we can review it in time 👍

map_view/display-layer-view-state/src/main/java/com/esri/samples/display_layer_view_state/DisplayLayerViewStateSample.java
jenmerritt: Yes good point, now the button implementation has changed it is simple to have Load initially, then show/hide. I tweaked the if statement in the `toggleLayerButton` listener and it works well, thanks for this suggestion!

map_view/display-layer-view-state/src/main/java/com/esri/samples/display_layer_view_state/DisplayLayerViewStateSample.java
jenmerritt: Agreed on this point, it was very frustrating to have to pull them out of the css. Something for us to look at.

map_view/display-layer-view-state/README.md
jenmerritt: Thanks, fixed the word order. I've also amended to disconnect from network, that's really what I was doing when testing, and I'll mention at review.

map_view/display-layer-view-state/README.md
```diff
@@ -6,33 +6,33 @@ Determine if a layer is currently being viewed.
...
+The view status includes information on the loading state of layers and whether layers are visible at a given scale. You might change how a layer is displayed in a layer list to communicate whether it is being viewed in the map. For example, you could show a loading spinner next to its name when the view status is `LOADING`, grey out the name when `NOT_VISIBLE` or `OUT_OF_SCALE`, show the name normally when `ACTIVE`, or with a warning or error icon when the status is `WARNING` or `ERROR`.
 
 ## How to use the sample
 
-Pan and zoom around in the map. Each layer's view status is displayed. Notice that some layers configured with a min and max scale change to `OUT_OF_SCALE` at certain scales.
+Click the *Load Layer* button to create a new layer and add it to the map. As you pan and zoom around the map, note how the `LayerViewStatus` flags change; for example, `OUT_OF_SCALE` becomes true when the map is scaled outside of the layer's min and max scale range. Click the *Hide Layer* button to hide the layer and observe the view state change to `NOT_VISIBLE`.
```
Rachael-E: The readme template wiki in common samples shows that this section should be a plain english description of how to run the app, along with "a short statement on what the user will see (technically, as in a raster on a map, etc.) and add a little call to action (i.e. 'pan and zoom to explore')."

This current description (copied from the design) is API terminology heavy, and API terminology should not sit in this section in my experience. I would word smith it to be API term agnostic:

```suggestion
Click "Load Layer" to add a new layer to the map. After it loads, the current view status of the map will display at the top of the sample. Zoom in and out of the map, and note the layer disappears when the map is scaled outside of its min and max scale range. The view status will update accordingly. Turn the layer's visibility off by clicking "Hide Layer". 
```

Rachael-E (commented):
We had a good pair/trio programming session yesterday about some behaviour I noticed in the sample whilst running it. For recording/transparency purposes, documenting a summary here:

- update to "How to use the sample" readme to remove direct references to the API (see review comment)
- update the UI so that it cannot be interacted with until the feature layer has loaded
- consider changing the hide/show layer UI function from a button to a checkbox (to be more desktop GIS-y, acknowledging that different platforms do have slightly different UI e.g. Android buttons easier to handle than check boxes)
- code refactoring to avoid excess lines of codes (checking with the samples team about shortening the long if statements referring to the enums)
- I put up an alternative layout branch for including a checkbox here https://github.com/Esri/arcgis-runtime-samples-java/tree/rach_layer_view_state: if using this good to double check it still displays all the behaviour needed from the sample (e.g. loading status from the view status and too many features error).

jenmerritt:
Note: this PR is currently on hold for any merging until the release is out, and will re-target `master` branch.

Following the session and discussions we had @Rachael-E , and also running a few ideas past the Samples team on Friday, I've updated the branch with an altered UI and some changed structure. I'll summarise below what I've done that is different and some of the thinking:

- There is a button now that purely loads or reloads the layer, so that the `LOADING` view status can be seen clearly.
- The show/hide layer is controlled now by a checkbox as to whether the loaded layer is active or not.
- This show/hide functionality is only enabled when the layer is loaded, as per a done loading listener that has been added.
- The current view status is updated via the if statements, as the Samples team overall preferred this approach to make it absolutely clear what all the enums are and the message that can be set.
- Updated the readme image and also the how to use section.

Other notes:
- The combination of the load/reload button and done loading listener I think works well due to the fact that if you play around with disconnecting from the network midway through loading the layer (i.e. before the view status gets to being active), it was previously possible to get stuck at "Error", whereas now, it's possible to reconnect to the network and reload the layer, or, as before if the layer is active it will just display a warning status until the network is reconnected. So it feels more robust.

- If the network is disconnected before the layer initially loads, there are now two errors: one from the done loading listener and one from the layer view status. I think this is ok as it is correct behaviour? But open to any thoughts on this.

- I've now got the one error message which Samples team agreed is fine. It's currently "Unable to update layer view status". I'm going round in circles on this, but theoretically the layer view status does update - as even if there is a warning it will show "out of scale" rather than active if you zoom. So not sure if that is completely accurate, or whether it could be something like "Unable to update layer display"? Again any thoughts on this one appreciated.

Thanks!

map_view/display-layer-view-state/build.gradle
```diff
@@ -6,7 +6,7 @@ plugins {
 group = 'com.esri.samples'
 
 ext {
-    arcgisVersion = '100.8.0'
+    arcgisVersion = '100.9.0-2873'
```
Rachael-E: we can update this to say 100.9.0 now that the release is live!

map_view/display-layer-view-state/src/main/java/com/esri/samples/display_layer_view_state/DisplayLayerViewStateSample.java
```diff
@@ -29,116 +36,142 @@
...
+        final PortalItem portalItem = new PortalItem(new Portal("https://runtime.maps.arcgis.com/"),
+          "b8f4033069f141729ffb298b7418b653");
+        featureLayer = new FeatureLayer(portalItem, 0);
+        //  set a minimum and a maximum scale for the visibility of the feature layer
+        featureLayer.setMinScale(400000000);
+        featureLayer.setMaxScale(4000000);
+        // add the feature layer to the map
+        map.getOperationalLayers().add(featureLayer);
+
+        // add a done loading listener to the feature layer
```
Rachael-E:
```suggestion
        // enable UI interactions once the feature layer has loaded
```

map_view/display-layer-view-state/src/main/java/com/esri/samples/display_layer_view_state/DisplayLayerViewStateSample.java
```diff
@@ -29,116 +36,142 @@
...
+        featureLayer.addDoneLoadingListener(() -> {
+          if (featureLayer.getLoadStatus() == LoadStatus.LOADED){
```
Rachael-E: extra space needed here between the parenthesis and curly brackets 
```suggestion
          if (featureLayer.getLoadStatus() == LoadStatus.LOADED) {
```

map_view/display-layer-view-state/src/main/java/com/esri/samples/display_layer_view_state/DisplayLayerViewStateSample.java
```diff
@@ -29,116 +36,142 @@
...
+      CheckBox visibilityCheckBox = new CheckBox();
+      visibilityCheckBox.setText("Layer active");
+      visibilityCheckBox.setSelected(true);
+      visibilityCheckBox.setDisable(true);
+      visibilityCheckBox.setOnAction(event -> featureLayer.setVisible(visibilityCheckBox.isSelected()));
 
-      // create three layers to add to the ArcGISMap
-      final ArcGISTiledLayer tiledLayer = new ArcGISTiledLayer(SERVICE_TIME_ZONES);
+      // create a button and a listener to load a feature layer
+      Button loadLayerButton = new Button("Load Layer");
```
Rachael-E: can remove the extra white space here

map_view/display-layer-view-state/src/main/java/com/esri/samples/display_layer_view_state/DisplayLayerViewStateSample.java
```diff
@@ -29,116 +36,142 @@
...
 
-      // this layer will be OUT_OF_SCALE outside of its set min/max scales
-      final ArcGISMapImageLayer imageLayer = new ArcGISMapImageLayer(SERVICE_CENSUS);
-      imageLayer.setMinScale(MIN_SCALE);
-      imageLayer.setMaxScale(MIN_SCALE / 10);
+      loadLayerButton.setOnAction(event -> {
```
Rachael-E, line 96: let's add in a comment or two here to help a user understand why these steps are necessary:

// disable the checkbox when loading/reloading the feature layer, and, if it exists, remove it by clearing the map's operational layers

map_view/display-layer-view-state/src/main/java/com/esri/samples/display_layer_view_state/DisplayLayerViewStateSample.java
```diff
@@ -29,116 +36,142 @@
...
 
-      // create a view and set ArcGISMap to it
-      mapView = new MapView();
-      mapView.setMap(map);
+        // create a new feature layer from a portal item
+        final PortalItem portalItem = new PortalItem(new Portal("https://runtime.maps.arcgis.com/"),
+          "b8f4033069f141729ffb298b7418b653");
+        featureLayer = new FeatureLayer(portalItem, 0);
+        //  set a minimum and a maximum scale for the visibility of the feature layer
+        featureLayer.setMinScale(400000000);
```
Rachael-E: looks like a 0 needs to be removed here as discussed so we get the error message when zooming out

map_view/display-layer-view-state/src/main/java/com/esri/samples/display_layer_view_state/DisplayLayerViewStateSample.java
```diff
@@ -169,4 +202,4 @@ public static void main(String[] args) {
     Application.launch(args);
   }
 
-}
+}
```
Rachael-E: add new line at the end of the file (standard practice)

map_view/display-layer-view-state/README.md
```diff
@@ -6,33 +6,31 @@ Determine if a layer is currently being viewed.
...
 
 ## How to use the sample
 
-Pan and zoom around in the map. Each layer's view status is displayed. Notice that some layers configured with a min and max scale change to `OUT_OF_SCALE` at certain scales.
+Click "Load Layer" to add a new layer to the map. The current view status of the layer will display at the top of the sample. Zoom in and out of the map and note that the layer disappears when the map is scaled outside of its min and max scale range. If you disconnect your device from the network and pan around the map, a warning alert will display. Reconnect to the network to remove the warning. Turn the layer's visibility off by unchecking the "layer active" checkbox. The layer's current view status will update accordingly as you carry out these actions.
```
Rachael-E: just some word smithing suggestions:

```suggestion
Click "Load Layer" to add a feature layer to the map. The current view status of the layer will display at the top of the sample. Zoom in and out of the map and note the layer disappears when the map is scaled outside of its min and max scale range. Control the layer's visibility with the "Layer visibility" checkbox. If you disconnect your device from the network and pan around the map, a warning alert will display. Reconnect to the network to remove the warning. The layer's current view status will update accordingly as you carry out these actions.
```

map_view/display-layer-view-state/src/main/java/com/esri/samples/display_layer_view_state/DisplayLayerViewStateSample.java
```diff
@@ -29,116 +36,142 @@
...
-      censusLabel.getStyleClass().add("panel-label");
-      facilitiesLabel.getStyleClass().add("panel-label");
+      // create a label to display the view status
+      Label layerViewStatusLabel = new Label("Current view status: ");
 
-      // add labels to the control panel
-      controlsVBox.getChildren().addAll(worldTimeZonesLabel, censusLabel, facilitiesLabel);
+      // create a checkbox UI that toggles the visibility of the feature layer
+      CheckBox visibilityCheckBox = new CheckBox();
+      visibilityCheckBox.setText("Layer active");
```
Rachael-E: now I understand this sample better, I think we should rename this to "Layer visibility"

Rachael-E (commented):
hey @jenmerritt great job refactoring this and making the UI so robust! this flows really well now and is a great sample. 

I've added some comment/readme suggestions, and an update to the checkbox text: nothing too major. For the error messages, both are firing correctly and related to the correct code blocks: so we'll just have to settle for a user seeing two alerts, they both say different things (which I didn't initially realise). So that's fine.  

Also happy with the error message "Unable to update layer view status": makes sense to me!

One other thought: the .png shows the status of "warning" but no associated error dialog. I think for "showing off" the sample purposes we should have a screenshot that doesn't show the warning error: just the "Active" message would be great.

map_view/display-layer-view-state/src/main/java/com/esri/samples/display_layer_view_state/DisplayLayerViewStateSample.java
jenmerritt: Making "layer visible" as discussed

map_view/display-layer-view-state/README.md
jenmerritt: Implemented as above, but making "Layer visible" checkbox as discussed

jenmerritt:
@Rachael-E have addressed those comments. Also changed the base to master and merged master into my branch. Thanks!

Rachael-E (approved):
@jenmerritt great job, this looks good to merge now :)

PR #543, opened 2020-08-04
New sample: Edit features with feature-linked annotation

sclaridge:
Hello, 

- the `identifyFeature()` method identifies a feature in the topmost layer. This implementation handles the case where the user moves a point very close to a line, then wants to move either feature later. Otherwise, an exception is thrown when both features are selected, then moved. 

- I created `updateAtttributes()` to update the feature when it is moved, but the logic to update the user entered attributes is located within `EditAttributesDialog`

- in the samples review, I received feedback that the sample does not need to prompt the user if they want to move the point, as that is outlined in the README.

- is README Java specific enough?

Thanks!


editing/edit-features-with-feature-linked-annotation/src/main/java/com.esri.samples.edit_features_with_feature_linked_annotation/EditAttributesDialog.java
```diff
@@ -0,0 +1,83 @@
...
+import javafx.fxml.FXMLLoader;
+import javafx.scene.control.Alert;
+import javafx.scene.control.ButtonType;
+import javafx.scene.control.Dialog;
+import javafx.scene.control.TextField;
+
+import com.esri.arcgisruntime.data.Feature;
+import com.esri.arcgisruntime.concurrent.ListenableFuture;
+
+public class EditAttributesDialog extends Dialog<String>{
```
JonLavi: Let's add a short description to this class, just for some context. An example based on some other classes extending dialog:
```
/**
 * Custom dialog for editing feature attributes.
 */
```

Also, although it has no direct implications in the way we are using it, it would be more correct to define this as:

```suggestion
public class EditAttributesDialog extends Dialog<Feature>{
```

Since we are using this dialog to interact with a particular feature (as you have it in the constructor).

editing/edit-features-with-feature-linked-annotation/src/main/java/com.esri.samples.edit_features_with_feature_linked_annotation/EditFeaturesWithFeatureLinkedAnnotationController.java
```diff
@@ -0,0 +1,310 @@
...
+
+    try {
+      // create the map with a light gray canvas basemap centered on Loudoun, Virginia
+      ArcGISMap map = new ArcGISMap(Basemap.Type.LIGHT_GRAY_CANVAS_VECTOR, 39.0204, -77.4159, 18);
+
+      // add a map to the map view
+      mapView.setMap(map);
+
+      // create the geodatabase from the local .geodatabase file
+      Geodatabase geodatabase = new Geodatabase("loudoun_anno.geodatabase");
```
JonLavi: We don't generally add the files directly from the source code, but set up the `settings.gradle` file to download these when the gradle `run` task is started. Check out this example:
https://github.com/Esri/arcgis-runtime-samples-java/blob/master/editing/edit-and-sync-features/settings.gradle#L6-L16

There should be a link to the location of the file on ArcGIS Online (AGOL) in the sample's design.

The correct path for it would then be `/resources/edit_features_with_feature_linked_annotation/<possibly some other folder>`.
Whereas the last folder might be the result of a folder within the archive coming from AGOL (or none!).

editing/edit-features-with-feature-linked-annotation/src/main/java/com.esri.samples.edit_features_with_feature_linked_annotation/EditFeaturesWithFeatureLinkedAnnotationController.java
```diff
@@ -0,0 +1,310 @@
...
+    selectedFeature = null;
+
+    // identify across all layers
+    ListenableFuture<List<IdentifyLayerResult>> identifyLayerResultsFuture = mapView.identifyLayersAsync(screenPoint, 1, false);
+    identifyLayerResultsFuture.addDoneListener(() -> {
+      try {
+        // get the list of results from the future
+        List<IdentifyLayerResult> identifyLayerResults = identifyLayerResultsFuture.get();
+        // if one or more results have been identified
+        if (identifyLayerResults.size() > 0) {
```
JonLavi: It's generally considered better to write 
```suggestion
        if (!identifyLayerResults.size().isEmpty()) {
```

editing/edit-features-with-feature-linked-annotation/src/main/java/com.esri.samples.edit_features_with_feature_linked_annotation/EditFeaturesWithFeatureLinkedAnnotationController.java
```diff
@@ -0,0 +1,310 @@
...
+    editResult.addDoneListener(() -> {
+              try {
+                if (editResult.isDone()) {
+                  editResult.get();
+                }
+              } catch (InterruptedException | ExecutionException e) {
+                new Alert(Alert.AlertType.ERROR, "Cannot update attributes: " + e.getCause().getMessage()).show();
+              }
+            }
+    );
```
JonLavi: Nice work on this, there's only a couple of things we could improve here.
In the first line, were technically creating a Future object, which we then unpack with `.get()` to get the actual 'edit result' from inside. So Let's call this object `editResultFuture`
Since the lambda inside the `DoneListener` will only ever run if the future `isDone()`, there no need to check that again within the listener. You can just go ahead and do `.get()`.
Lastly, we don't generally display the whole error message (i.e. ` e.getCause().getMessage()` in the alerts) but just show something in plain english, describing what may have gone wrong. So let's update that message

```suggestion
    ListenableFuture<Void> editResultFuture = selectedFeature.getFeatureTable().updateFeatureAsync(selectedFeature);
    editResultFuture.addDoneListener(() -> {
              try {
                editResultFuture.get();
              } catch (InterruptedException | ExecutionException e) {
                new Alert(Alert.AlertType.ERROR, "Error updating attributes.").show();
              }
            }
    );
```

editing/edit-features-with-feature-linked-annotation/src/main/java/com.esri.samples.edit_features_with_feature_linked_annotation/EditFeaturesWithFeatureLinkedAnnotationController.java
```diff
@@ -0,0 +1,310 @@
...
+    } catch (Exception e) {
+      e.printStackTrace();
+    }
+  }
+
+  /**
+   * Create a dialog with text fields to allow editing of the given feature's 'AD_ADDRESS' and
+   * 'ST_STR_NAM' attributes.
+   */
+  private void showEditableAttributes(Feature selectedFeature) {
```
JonLavi: For your methods, it's common practice to put `@param` descriptions of the parameters in the JavaDoc above method descriptions.
Intellij Top tip: you can auto-generate a template by writing `/**` in the line above a method signature, then hitting return. All you'll have to do is then fill in the descriptions. The rest of your methods would need updating as well.
I think they should also be written in active voice, e.g. `Creates`/`Moves`, etc. But not 100% sure.

editing/edit-features-with-feature-linked-annotation/src/main/resources/edit_features_with_feature_linked_annotation/main.fxml
```diff
@@ -0,0 +1,23 @@
...
+  ~ WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+  ~ License for the specific language governing permissions and limitations under
+  ~ the License.
+  -->
+
+<?import com.esri.arcgisruntime.mapping.view.MapView?>
+<?import javafx.scene.layout.StackPane?>
+<StackPane fx:controller="com.esri.samples.edit_features_with_feature_linked_annotation.EditFeaturesWithFeatureLinkedAnnotationController"
+           xmlns:fx="http://javafx.com/fxml">
+    <MapView fx:id="mapView"/>
```
JonLavi: This FXML file doesn't really do much, and considering the main reason we use FXML is to refactor away some of the UI styling, I would personally say we could get rid of this one, and simply move all the code from the `controller` class to the `sample` class. We should keep the FXML file for the editing dialog, however, because that does do quite a bit of work - that's where all the refactoring gains are for the UI styling.

editing/edit-features-with-feature-linked-annotation/build.gradle
```diff
@@ -0,0 +1,74 @@
...
+}
+
+group = 'com.esri.samples'
+
+ext {
+    arcgisVersion = '100.9.0-2852'
+}
+
+javafx {
+    version = "11.0.1"
```
JonLavi: I think JavaFX 11.0.2 is the newest version, so we should update that here

editing/edit-features-with-feature-linked-annotation/src/main/java/com.esri.samples.edit_features_with_feature_linked_annotation/EditAttributesDialog.java
```diff
@@ -0,0 +1,83 @@
...
+
+  EditAttributesDialog(Feature selectedFeature){
+    FXMLLoader loader = new FXMLLoader(getClass().getResource("/edit_features_with_feature_linked_annotation/edit_attributes_dialog.fxml"));
+    loader.setRoot(this);
+    loader.setController(this);
+
+    setTitle("Edit Feature Attributes");
+
+    try {
+      loader.load();
```
JonLavi, line 45: I _think_ we need to put the remaining code (line 50 onwards) inside this `try` statements, since if we fail to load the fxml loader, we won't want to execute any of the other code, as we don't have access to any of these FXML defined javafx objects.

It seems like our other samples don't do this though, so let's do some research on this which is the correct pattern!

editing/edit-features-with-feature-linked-annotation/src/main/java/com.esri.samples.edit_features_with_feature_linked_annotation/EditAttributesDialog.java
```diff
@@ -0,0 +1,83 @@
...
+          editResult.addDoneListener(() -> {
+                    try {
+                      if (editResult.isDone()) {
+                        editResult.get();
+                      }
+                    } catch (InterruptedException | ExecutionException e) {
+                      new Alert(Alert.AlertType.ERROR, "Cannot update attributes: " + e.getCause().getMessage()).show();
+                    }
+                  }
+          );
```
JonLavi: This code looks like a repetition of the method 
```
private void updateAttributes(Feature selectedFeature)
```
in the controller class.

Not too sure if it's good practice, but you could modify the signature to:
```
static void updateAttributes(Feature selectedFeature)
````
By leaving out `private` and making it `static`, it would be 'protected', meaning available to this dialog class, and it being 'static' would allow you to import it into this file as a static import:
```
import static com.esri.samples.edit_features_with_feature_linked_annotation.EditFeaturesWithFeatureLinkedAnnotationController.updateAttributes;
```

And then you could replace the whole ListenableFuture block here with simply:
```
          // update feature in the feature table
          updateAttributes(selectedFeature);
```

Like I said, not sure whether this is good practice.
@colinanderson could you advise whether this is considered SOLID?
But before you go through the effort of doing this, I suppose we'd need @Rachael-E 's input whether this is user-friendly enough...

File: editing/edit-features-with-feature-linked-annotation/src/main/java/com.esri.samples.edit_features_with_feature_linked_annotation/EditFeaturesWithFeatureLinkedAnnotationController.java [hunk 10]
```diff
@@ -0,0 +1,310 @@
...
+    // set the selected feature's geometry to the new polyline
+    selectedFeature.setGeometry(polylineBuilder.toGeometry());
+
+    // update the selected feature's feature table
+    updateAttributes(selectedFeature);
+
+    // clear selection of the polyline
+    clearSelection();
+    selectedFeatureIsPolyline = false;
+    selectedFeature = null;
```
JonLavi: Looks like every time we call `clearSelection()`, we also do `selectedFeature = null`. So it would be possible to put that statement in the method as well.
We could also pop the statement `selectedFeatureIsPolyline = false;` in the method as well, although it's not necessary to set that boolean every time.

editing/edit-features-with-feature-linked-annotation/src/main/java/com.esri.samples.edit_features_with_feature_linked_annotation/EditFeaturesWithFeatureLinkedAnnotationController.java
```diff
@@ -0,0 +1,310 @@
...
+  private void movePolylineVertex(Point mapPoint) {
+
+    // get the selected feature's geometry as a polyline
+    Polyline polyline = (Polyline) selectedFeature.getGeometry();
+
+    // create a polyline builder to add and remove parts from the polyline
+    PolylineBuilder polylineBuilder = new PolylineBuilder(polyline);
+
+    // get the nearest vertex to the map point on the polyline
+    ProximityResult nearestVertex = GeometryEngine.nearestVertex(polyline, (Point) GeometryEngine.project(mapPoint, polyline.getSpatialReference()));
```
JonLavi: This is a really long line, so let's just break it down for readability:
```suggestion
    ProximityResult nearestVertex =
            GeometryEngine.nearestVertex(polyline, (Point) GeometryEngine.project(mapPoint, polyline.getSpatialReference()));
```

editing/edit-features-with-feature-linked-annotation/README.metadata.json
```diff
@@ -0,0 +1,34 @@
+{
+    "category": "",
```
JonLavi: Let's not forget to add the category here. 
If you haven't been using the script to generate this, do give it a shot. We're constantly optimising it so it would be helpful to see the output with every new sample
```suggestion
    "category": "Editing",
```

editing/edit-features-with-feature-linked-annotation/src/main/resources/edit_features_with_feature_linked_annotation/edit_attributes_dialog.fxml
```diff
@@ -0,0 +1,40 @@
...
+  ~ License for the specific language governing permissions and limitations under
+  ~ the License.
+  -->
+
+<?import javafx.scene.control.Dialog?>
+<?import javafx.scene.control.DialogPane?>
+<?import javafx.scene.layout.GridPane?>
+<?import javafx.scene.control.Label?>
+<?import javafx.scene.control.TextField?>
+<?import javafx.scene.control.ButtonType?>
```
JonLavi, line 21: Looks like the imports here are out of order. The optimize-imports commands should sort that out easily

JonLavi (commented):
Nice work @sclaridge, this is looking great so far.
A few smaller comments for refactoring and structure.
Also, for this sample, I don't think we'd need the GIF. It's only really useful with samples where there's a lot going on graphically, so that it's easy to see things moving within a small thumbnail in the Sample Viewer.

JonLavi:
Hi @sclaridge ,
To address your questions:
 
>     * the `identifyFeature()` method identifies a feature in the topmost layer. This implementation handles the case where the user moves a point very close to a line, then wants to move either feature later. Otherwise, an exception is thrown when both features are selected, then moved.
> 

We don't have to overcomplicate the sample to account for every single case, so this sounds like a good compromise to me.

>     * I created `updateAtttributes()` to update the feature when it is moved, but the logic to update the user entered attributes is located within `EditAttributesDialog`

That's indeed a bit tricky. We could have the dialog return a `Pair<String, String>` with the information for each edited property, and then do the actual editing within the method `updateAttributes()`. You could have a go at that and see whether it looks clean and readible? 

> 
>     * in the samples review, I received feedback that the sample does not need to prompt the user if they want to move the point, as that is outlined in the README.
> 

That seems fair, and the readme looks OK in that regard to me 👍 

>     * is README Java specific enough?
> 

The Java-fied terminology looks good to me!
For the `how it works` section, I'm personally not a fan of the phrasing 'a dialog opens', since we'd generally write these sections in phrasing along the lines of `create/open/add/ a dialog` (not sure of the grammatical term here). 
However since this has been agreed upon in review, let's just leave as is for consistency with other platforms 👍

editing/edit-features-with-feature-linked-annotation/src/main/java/com.esri.samples.edit_features_with_feature_linked_annotation/EditAttributesDialog.java
JonLavi: As I mentioned in the comment on your questions, we also have the possibility to have the `Dialog` return the data to edit back to the main `Controller` class, and then do the `updateFeatureAsync` call back in there. In hindsight that might actually be the better way forward. Then we'd abstract away the data input through the UI, but keep the Runtime API call to `updateFeatureAsync` within the main class.

editing/edit-features-with-feature-linked-annotation/src/main/java/com.esri.samples.edit_features_with_feature_linked_annotation/EditAttributesDialog.java
```diff
@@ -0,0 +1,83 @@
...
+          selectedFeature.getAttributes().put("AD_ADDRESS", Integer.parseInt(addressTextField.getText()));
+
+          // set ST_STR_NAM value to the string from the text field
+          selectedFeature.getAttributes().put("ST_STR_NAM", streetNameTextField.getText());
+
+          // update feature in the feature table
+          ListenableFuture<Void> editResult = selectedFeature.getFeatureTable().updateFeatureAsync(selectedFeature);
+          editResult.addDoneListener(() -> {
+                    try {
+                      if (editResult.isDone()) {
```
colinanderson: Do you need to check `isDone`? I think just calling `get()` will either work or throw an exception.

editing/edit-features-with-feature-linked-annotation/src/main/java/com.esri.samples.edit_features_with_feature_linked_annotation/EditAttributesDialog.java
sclaridge: If I put the remaining code inside the `try` statement, when the loader fails, the rest of the code doesn't execute, but the code that executes after instantiating the dialog in `showEditableAttributes()` creates an empty dialog: 

image

However, if I remove the try catch entirely and have `EditAttributesDialog` throw the Exception, it can be handled in `showEditableAttributes()`:

```
try {
      // create a dialog to edit the attributes of the selected feature
      EditAttributesDialog editAttributesDialog = new EditAttributesDialog(selectedFeature);

      // show the dialog and wait for the user response
      editAttributesDialog.showAndWait();

      // update the selected feature's feature table
      updateAttributes(selectedFeature);
    } catch (Exception e) {
      new Alert(Alert.AlertType.ERROR, "Error: " + e).show();
    }
```
image

JonLavi: Good spot on these! 
> However, if I remove the try catch entirely and have EditAttributesDialog throw the Exception, it can be handled in showEditableAttributes():

I don't think it's a good idea to handle the exceptions in a different method, at least in the context of a sample. Best we keep the complexity to a minimum there.
So, looks like what you had originally was correct. Let's just leave it as is then 👍

editing/edit-features-with-feature-linked-annotation/src/main/java/com.esri.samples.edit_features_with_feature_linked_annotation/EditAttributesDialog.java
```diff
@@ -0,0 +1,71 @@
...
+    setResultConverter(dialogButton -> {
+      if (dialogButton == continueButton) {
+        try {
+          // set AD_ADDRESS value to the int from the text field
+          selectedFeature.getAttributes().put("AD_ADDRESS", Integer.parseInt(addressTextField.getText()));
+
+          // set ST_STR_NAM value to the string from the text field
+          selectedFeature.getAttributes().put("ST_STR_NAM", streetNameTextField.getText());
+
+          return selectedFeature;
```
JonLavi: Nice work on tidying this up.
I've had a closer look and there's actually no need to return the selected feature here. This is because there's a handle on it in the main sample code already (from when we run mapView.identifyLayersAsync). So for the sake of simplicty let's convert back to `return null;`

editing/edit-features-with-feature-linked-annotation/src/main/java/com.esri.samples.edit_features_with_feature_linked_annotation/EditFeaturesWithFeatureLinkedAnnotationSample.java
```diff
@@ -0,0 +1,357 @@
...
+      mapView.setMap(map);
+
+      // add the map view to stack pane
+      stackPane.getChildren().addAll(mapView);
+      
+      // create and load the geodatabase
+      File geodatabaseFile = new File(System.getProperty("data.dir"),
+        "./src/main/resources/edit_features_with_feature_linked_annotation/loudon/loudoun_anno.geodatabase");
+      Geodatabase geodatabase = new Geodatabase(geodatabaseFile.getAbsolutePath());
+      geodatabase.addDoneLoadingListener(() -> {
```
JonLavi: Let's just give this a tad more breathing space and separate into two blocks. One where we create the GDB, and the other where we add the done loading listener:

```suggestion
      Geodatabase geodatabase = new Geodatabase(geodatabaseFile.getAbsolutePath());
      
      geodatabase.addDoneLoadingListener(() -> {
```

editing/edit-features-with-feature-linked-annotation/src/main/java/com.esri.samples.edit_features_with_feature_linked_annotation/EditFeaturesWithFeatureLinkedAnnotationSample.java
```diff
@@ -0,0 +1,357 @@
...
+   *
+   * @param screenPoint location the user clicked
+   */
+  private void identifyFeature(Point2D screenPoint) {
+
+    // clear any previously selected features
+    clearSelection();
+
+    // identify across all layers
+    ListenableFuture<List<IdentifyLayerResult>> identifyLayerResultsFuture = mapView.identifyLayersAsync(screenPoint, 1, false);
```
JonLavi: We can be a bit more generous here and give a larger tolerance. That way it'll be a bit easier to select the lines.

```suggestion
    ListenableFuture<List<IdentifyLayerResult>> identifyLayerResultsFuture = mapView.identifyLayersAsync(screenPoint, 10, false);
```

editing/edit-features-with-feature-linked-annotation/src/main/java/com.esri.samples.edit_features_with_feature_linked_annotation/EditFeaturesWithFeatureLinkedAnnotationSample.java
```diff
@@ -0,0 +1,357 @@
...
+            selectedFeature = null;
+            // show message reminding user to select straight (single segment) polylines only
+            new Alert(Alert.AlertType.WARNING, "Select straight (single segment) polylines only.").show();
+            // return early, effectively disallowing selection of multi segmented polylines
+            return;
+          } else {
+            // select the polyline feature
+            ((FeatureLayer) layerResult.getLayerContent()).selectFeature(selectedFeature);
+            selectedFeatureIsPolyline = true;
+          }
```
JonLavi: I personally think we can simplify this by turning the if-statment around. Doing an 'early return' is very much a C# pattern, as far as I'm aware.

```suggestion
          // only select single segment lines
          if (part.getPointCount() <= 2) {
            // select the polyline feature
            ((FeatureLayer) layerResult.getLayerContent()).selectFeature(selectedFeature);
            selectedFeatureIsPolyline = true;
          
          } else {
            selectedFeature = null;
            // show message reminding user to select straight (single segment) polylines only
            new Alert(Alert.AlertType.WARNING, "Select straight (single segment) polylines only.").show();
          }
```

editing/edit-features-with-feature-linked-annotation/src/main/java/com.esri.samples.edit_features_with_feature_linked_annotation/EditFeaturesWithFeatureLinkedAnnotationSample.java
```diff
@@ -0,0 +1,357 @@
...
+      });
+    } catch (Exception e) {
+      e.printStackTrace();
+    }
+  }
+
+  /**
+   * Selects the nearest feature, or move the point or polyline vertex to the given screen point.
+   *
+   * @param screenPoint location the user clicked
```
JonLavi: I'm not the documentation expert, but I have the impression we usually start `@param` descriptions with `the`, and describe these parameters based on their context within the function. So we could write here:
```suggestion
   * @param screenPoint the screen point at which to select a feature, or to move a feature to
```

editing/edit-features-with-feature-linked-annotation/src/main/java/com.esri.samples.edit_features_with_feature_linked_annotation/EditFeaturesWithFeatureLinkedAnnotationSample.java
```diff
@@ -0,0 +1,357 @@
...
+        // if the feature is a point, move the point
+        movePoint(mapPoint);
+      }
+    }
+  }
+
+  /**
+   * Identifies a feature near the given screen point.
+   *
+   * @param screenPoint location the user clicked
```
JonLavi: Another documentation suggestion: we can describe what purpose this parameter serves within the method:
```suggestion
   * @param screenPoint the screen point at which to identify a feature
```

editing/edit-features-with-feature-linked-annotation/src/main/java/com.esri.samples.edit_features_with_feature_linked_annotation/EditFeaturesWithFeatureLinkedAnnotationSample.java
```diff
@@ -0,0 +1,357 @@
...
+        e.printStackTrace();
+      }
+    });
+  }
+
+  /**
+   * Checks if the identified feature is a straight polyline or a point, and select the feature.
+   * For a point feature, show a dialog to edit attributes. Future clicks will call move functions.
+   *
+   * @param layerResult identify layer result
```
JonLavi: Just another suggestion:
```suggestion
   * @param layerResult the identify layer result from which to select a feature
```

editing/edit-features-with-feature-linked-annotation/src/main/java/com.esri.samples.edit_features_with_feature_linked_annotation/EditFeaturesWithFeatureLinkedAnnotationSample.java
```diff
@@ -0,0 +1,357 @@
...
+          // check that the result is a feature layer, thereby excluding annotation layers
+          if (layerResult.getLayerContent() instanceof FeatureLayer) {
+            // get a reference to the identified feature
+            selectedFeature = (Feature) layerResult.getElements().get(0);
+            // check the geometry and select the feature
+            selectFeature(layerResult);
+          }
+        }
+      } catch (Exception e) {
+        e.printStackTrace();
```
JonLavi: Let's show an alert here instead of just printing the stack trace:

```suggestion
        new Alert(Alert.AlertType.ERROR, "Error identifying the clicked feature.").show();
```

editing/edit-features-with-feature-linked-annotation/src/main/java/com.esri.samples.edit_features_with_feature_linked_annotation/EditFeaturesWithFeatureLinkedAnnotationSample.java
```diff
@@ -0,0 +1,357 @@
...
+      else if (selectedFeature.getGeometry().getGeometryType() == GeometryType.POINT) {
+        ((FeatureLayer) layerResult.getLayerContent()).selectFeature(selectedFeature);
+        // open a dialog to edit the feature's attributes
+        showEditableAttributes(selectedFeature);
+      } else {
+        new Alert(Alert.AlertType.WARNING, "Feature of unexpected geometry type selected.").show();
+      }
+    } catch (Exception e) {
+      e.printStackTrace();
+    }
```
JonLavi: It doesn't look like there's any need for this try/catch block anymore, since all the asynchronous actions are now being done in other methods? In which case we can just remove it

JonLavi (approved):
Nice work @sclaridge .
I think this looks good now. There's a couple of smaller things to fix and a few suggestions on documentation, but other than that it's nearly finished. I'll approve so that you can move on to get a second review once you've done the changes 👍

editing/edit-features-with-feature-linked-annotation/src/main/java/com.esri.samples.edit_features_with_feature_linked_annotation/EditFeaturesWithFeatureLinkedAnnotationSample.java
sclaridge: Increasing the tolerance forces you to zoom in to select a feature, as annotation layers are identified as the top most layer. A for loop can be added to ignore the annotation layers and select the feature. I think this implementation is cleaner, and is similar to the identify workflow in the Android sample. 

```
// identify across all layers
ListenableFuture<List<IdentifyLayerResult>> identifyLayerResultsFuture = mapView.identifyLayersAsync(screenPoint, 10, false);
    identifyLayerResultsFuture.addDoneListener(() -> {
      try {
        // get the list of results from the future
        List<IdentifyLayerResult> identifyLayerResults = identifyLayerResultsFuture.get();
        // for each layer from which an element was identified
        for (IdentifyLayerResult layerResult : identifyLayerResults) {
          // check if the layer is a feature layer, thereby excluding annotation layers
          if (layerResult.getLayerContent() instanceof FeatureLayer) {
            // get a reference to the identified feature
            selectedFeature = (Feature) layerResult.getElements().get(0);
            // check the geometry and select the feature
            selectFeature(layerResult);
            return;
          }
     }
```

editing/edit-features-with-feature-linked-annotation/src/main/java/com.esri.samples.edit_features_with_feature_linked_annotation/EditFeaturesWithFeatureLinkedAnnotationController.java
Rachael-E: @JonLavi @sclaridge just to note the above path for storing samples data is incorrect!

`/resources/sample_name` should be for UI/style files (.fxml/css) ***only***. Any samples data needs to be stored in a `samples-data` directory. You can see this pattern in the edit and sync features sample that Jon referred to above. 

e.g. for your sample
```gradle
    ant.unzip(
            src: file("./data.zip"),
            dest: file("./samples-data/loudon")
    )
```

editing/edit-features-with-feature-linked-annotation/build.gradle
```diff
@@ -0,0 +1,74 @@
+plugins {
+    id 'application'
+    id 'org.openjfx.javafxplugin' version '0.0.5'
+}
+
+group = 'com.esri.samples'
+
+ext {
+    arcgisVersion = '100.9.0-2852'
```
Rachael-E: we can update this to 100.9.0 now that we're live!

editing/edit-features-with-feature-linked-annotation/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+![Image of edit features with feature-linked annotation](EditFeaturesWithFeatureLinkedAnnotation.png)
+
+## Use case
+
+Annotation is useful for displaying text that you don't want to move or resize when the map is panned or zoomed (unlike labels which will move and resize). Feature-linked annotation will update when a feature attribute referenced by the annotation expression is also updated. Additionally, the position of the annotation will transform to match any transformation to the linked feature's geometry.
+
+## How to use the sample
+
+Pan and zoom the map to see that the text on the map is annotation, not labels. Click one of the address points to update the house number (AD_ADDRESS) and street name (ST_STR_NAM). Click one of the dashed parcel polylines and click another location to change its geometry. NOTE: Selection is only enabled for points and straight (single segment) polylines.
```
Rachael-E: this description doesn't include actually moving an address point on the sample. This threw me initially as I clicked "OK" or "Cancel" and then clicked on the map/another point: wasn't expecting the address point to move too.  We should add in some additional documentation for this: 
```suggestion
Pan and zoom the map to see that the text on the map is annotation, not labels. Click one of the address points to update the house number (AD_ADDRESS) and street name (ST_STR_NAM). Once you have edited the feature attributes, click "OK" and then click again on the map to move the address point to a new location. You can also click one of the dashed parcel polylines and click another location to change its geometry and update its annotation (distance in feet). NOTE: Selection is only enabled for points and straight (single segment) polylines.
```

editing/edit-features-with-feature-linked-annotation/build.gradle
```diff
@@ -0,0 +1,74 @@
...
+        attributes("Main-Class": "$mainClassName")
+    }
+    from {
+        configurations.compile.collect { it.isDirectory() ? it : zipTree(it) }
+    }
+}
+
+task productionZip(type: Zip) {
+    group = 'distribution'
+    from copyNatives
```
Rachael-E, line 69: to make sure the samples data is included within a production zip made from this file, we also here need to include the following code here - this should be done for all samples which require locally saved data to run.

```gradle
    from ("samples-data") {
        into "samples-data"
    }
```

editing/edit-features-with-feature-linked-annotation/settings.gradle
```diff
@@ -0,0 +1,16 @@
...
+def ant = new groovy.util.AntBuilder()
+
+if (!file("./samples-data").exists()) {
+    ant.get(
+            src: "https://arcgisruntime.maps.arcgis.com/sharing/rest/content/items/74c0c9fa80f4498c9739cc42531e9948/data",
+            dest: file("./data.zip")
+    )
+    ant.unzip(
+            src: file("./data.zip"),
+            dest: file("./src/main/resources/edit_features_with_feature_linked_annotation/loudon")
```
Rachael-E: I added a comment in Jon's review, but just in case it gets lost: samples data should always be saved into a samples-data directory. The resources are reserved for UI or style files.

```suggestion
            dest: file("./samples-data/loudon")
```

editing/edit-features-with-feature-linked-annotation/src/main/java/com.esri.samples.edit_features_with_feature_linked_annotation/EditAttributesDialog.java
```diff
@@ -0,0 +1,69 @@
...
+      loader.load();
+    } catch (Exception e) {
+      e.printStackTrace();
+    }
+
+    // populate text fields with current attribute values
+    addressTextField.setText(selectedFeature.getAttributes().get("AD_ADDRESS").toString());
+    streetNameTextField.setText(selectedFeature.getAttributes().get("ST_STR_NAM").toString());
+
+    // convert the result to an address and street name when the ok button is clicked.
```
Rachael-E: in in line javadoc comments (i.e. those not describing a method/class), full stops/periods aren't required.
```suggestion
    // convert the result to an address and street name when the ok button is clicked
```

editing/edit-features-with-feature-linked-annotation/src/main/java/com.esri.samples.edit_features_with_feature_linked_annotation/EditAttributesDialog.java
```diff
@@ -0,0 +1,69 @@
...
+    setResultConverter(dialogButton -> {
+      if (dialogButton == continueButton) {
+        try {
+          // set AD_ADDRESS value to the int from the text field
```
Rachael-E: to avoid a keyboard-happy user like me inadvertently throwing a `java.lang.NumberFormatException` by putting lots of extra numbers in in this field ;) 
we can put in something like the following:

```java
          // ensure input is equal to or less than 5 characters (max length for addresses in area)
          if (addressTextField.getLength() <= 5){
            // set AD_ADDRESS value to the int from the text field
            selectedFeature.getAttributes().put("AD_ADDRESS", Integer.parseInt(addressTextField.getText()));
          } else  {
            new Alert(Alert.AlertType.WARNING, "Field not updated. Integer must be less than 6 characters").showAndWait();
          }
```

See also this [stackoverflow post](https://stackoverflow.com/questions/31846838/integer-parsestring-str-java-lang-numberformatexception-errors) for understanding better why an exception is thrown here when putting in too many digits.

editing/edit-features-with-feature-linked-annotation/src/main/java/com.esri.samples.edit_features_with_feature_linked_annotation/EditAttributesDialog.java
```diff
@@ -0,0 +1,69 @@
...
+import com.esri.arcgisruntime.data.Feature;
+
+/**
+ * Custom dialog for editing feature attributes.
+ */
+public class EditAttributesDialog extends Dialog<Feature>{
+
+  @FXML private TextField addressTextField;
+  @FXML private TextField streetNameTextField;
+  @FXML private ButtonType continueButton;
```
Rachael-E: perhaps this would be better named as "updateButton"? We could also update the button text to show "Update" which is perhaps a bit clearer than "Ok" in terms of what the user is doing if they click to edit the feature.

editing/edit-features-with-feature-linked-annotation/src/main/resources/edit_features_with_feature_linked_annotation/edit_attributes_dialog.fxml
```diff
@@ -0,0 +1,39 @@
...
+        <DialogPane>
+            <content>
+                <GridPane vgap="10" hgap="10">
+                    <Label text="AD_ADDRESS:" GridPane.rowIndex="0" GridPane.columnIndex="0"/>
+                    <TextField fx:id="addressTextField" GridPane.rowIndex="0" GridPane.columnIndex="1"/>
+                    <Label text="ST_STR_NAM:" GridPane.rowIndex="1" GridPane.columnIndex="0"/>
+                    <TextField fx:id="streetNameTextField" GridPane.rowIndex="1" GridPane.columnIndex="1"/>
+                </GridPane>
+            </content>
+            <ButtonType fx:id="continueButton" buttonData="OK_DONE" text="Ok"/>
```
Rachael-E: Usually on our samples desktop applications, I'd expect to see the cancel button on the right hand side of dialog, not the left. Our other samples follow this pattern too, so let's stick with that by switching "OK_DONE" for "FINISH" (looks like JavaFX controls the position of the buttons and the enum seems to do the same thing in this case). I also think "Update" might be clearer than "Ok".

```suggestion
            <ButtonType fx:id="continueButton" buttonData="FINISH" text="Update"/>
```

editing/edit-features-with-feature-linked-annotation/src/main/java/com.esri.samples.edit_features_with_feature_linked_annotation/EditFeaturesWithFeatureLinkedAnnotationSample.java
```diff
@@ -0,0 +1,350 @@
...
+import com.esri.arcgisruntime.mapping.view.IdentifyLayerResult;
+import com.esri.arcgisruntime.mapping.view.MapView;
+
+public class EditFeaturesWithFeatureLinkedAnnotationSample extends Application {
+
+  private MapView mapView;
+  private FeatureLayer addressPointFeatureLayer;
+  private FeatureLayer parcelLinesFeatureLayer;
+  private AnnotationLayer addressPointsAnnotationLayer;
+  private AnnotationLayer parcelLinesAnnotationLayer;
```
Rachael-E: all of these can be local variables in the `start` method, since they are not referenced elsewhere in the class. We don't need to worry about these being garbage collected, since we're not waiting for them to finish loading before doing something with them.

editing/edit-features-with-feature-linked-annotation/src/main/java/com.esri.samples.edit_features_with_feature_linked_annotation/EditFeaturesWithFeatureLinkedAnnotationSample.java
```diff
@@ -0,0 +1,350 @@
...
+      mapView = new MapView();
+      mapView.setMap(map);
+
+      // add the map view to stack pane
+      stackPane.getChildren().addAll(mapView);
+      
+      // create and load the geodatabase
+      File geodatabaseFile = new File(System.getProperty("data.dir"),
+        "./samples-data/loudon/loudoun_anno.geodatabase");
+      Geodatabase geodatabase = new Geodatabase(geodatabaseFile.getAbsolutePath());
```
Rachael-E: we should make this an instance variable. Since it is a loadable, and we are waiting on it to finish loading, there is a small chance it's initial reference will be garbage collected and by the time the addDoneLoadingListener has returned, the initial `geodatabase` reference is lost and the app will hang.  Let's add the following up above the start method:

```java
  private Geodatabase geodatabase; // keep loadable in scope to avoid garbage collection

```
```suggestion
      geodatabase = new Geodatabase(geodatabaseFile.getAbsolutePath());
```

editing/edit-features-with-feature-linked-annotation/src/main/java/com.esri.samples.edit_features_with_feature_linked_annotation/EditFeaturesWithFeatureLinkedAnnotationSample.java
```diff
@@ -0,0 +1,350 @@
...
+          addressPointsAnnotationLayer = new AnnotationLayer(geodatabase.getGeodatabaseAnnotationTable("Loudoun_Address_PointsAnno_1"));
+          parcelLinesAnnotationLayer = new AnnotationLayer(geodatabase.getGeodatabaseAnnotationTable("ParcelLinesAnno_1"));
+
+          // add the feature layers to the map
+          map.getOperationalLayers().add(addressPointFeatureLayer);
+          map.getOperationalLayers().add(parcelLinesFeatureLayer);
+
+          // add the annotation layers to the map
+          map.getOperationalLayers().add(addressPointsAnnotationLayer);
+          map.getOperationalLayers().add(parcelLinesAnnotationLayer);
```
Rachael-E: we can simplify the following with:

```java
// add the annotation and feature layers to the map
map.getOperationalLayers().addAll(Arrays.asList(
     addressPointFeatureLayer, parcelLinesFeatureLayer, addressPointsAnnotationLayer, parcelLinesAnnotationLayer));
```

editing/edit-features-with-feature-linked-annotation/src/main/java/com.esri.samples.edit_features_with_feature_linked_annotation/EditFeaturesWithFeatureLinkedAnnotationSample.java
```diff
@@ -0,0 +1,350 @@
...
+      Geodatabase geodatabase = new Geodatabase(geodatabaseFile.getAbsolutePath());
+
+      geodatabase.addDoneLoadingListener(() -> {
+        if (geodatabase.getLoadStatus() == LoadStatus.LOADED) {
+          // create feature layers from tables in the geodatabase
+          addressPointFeatureLayer = new FeatureLayer(geodatabase.getGeodatabaseFeatureTable("Loudoun_Address_Points_1"));
+          parcelLinesFeatureLayer = new FeatureLayer(geodatabase.getGeodatabaseFeatureTable("ParcelLines_1"));
+          // create annotation layers from tables in the geodatabase
+          addressPointsAnnotationLayer = new AnnotationLayer(geodatabase.getGeodatabaseAnnotationTable("Loudoun_Address_PointsAnno_1"));
+          parcelLinesAnnotationLayer = new AnnotationLayer(geodatabase.getGeodatabaseAnnotationTable("ParcelLinesAnno_1"));
```
Rachael-E: by the time we're using local variables for the FeatureLayer's here (see comment above), that's a lot of references to the word FeatureLayer in this block of code. For situations like this, we can make use of Java 10's keyword "var". This replaces the type information when declaring local variables and clears the readability up a little.

Also since these are quite long lines of code, we should stagger them a little:

 ```suggestion
          // create feature layers from tables in the geodatabase
          var addressPointFeatureLayer =
            new FeatureLayer(geodatabase.getGeodatabaseFeatureTable("Loudoun_Address_Points_1"));
          var parcelLinesFeatureLayer =
            new FeatureLayer(geodatabase.getGeodatabaseFeatureTable("ParcelLines_1"));
          // create annotation layers from tables in the geodatabase
          var addressPointsAnnotationLayer =
            new AnnotationLayer(geodatabase.getGeodatabaseAnnotationTable("Loudoun_Address_PointsAnno_1"));
          var parcelLinesAnnotationLayer =
            new AnnotationLayer(geodatabase.getGeodatabaseAnnotationTable("ParcelLinesAnno_1"));
```

File: editing/edit-features-with-feature-linked-annotation/src/main/java/com.esri.samples.edit_features_with_feature_linked_annotation/EditFeaturesWithFeatureLinkedAnnotationSample.java [hunk 36]
```diff
@@ -0,0 +1,350 @@
...
+
+          // add the annotation layers to the map
+          map.getOperationalLayers().add(addressPointsAnnotationLayer);
+          map.getOperationalLayers().add(parcelLinesAnnotationLayer);
+        } else {
+          // show alert if geodatabase fails to load
+          new Alert(Alert.AlertType.ERROR, "Error loading Geodatabase.").show();
+        }
+      });
+      geodatabase.loadAsync();
```
Rachael-E: I think for a readability perspective, it makes more sense for us to move this `loadAsync` call up (to above the `geodatabase.addDoneLoadingListener()` even though it doesn't make much of a difference from a code execution perspective.

editing/edit-features-with-feature-linked-annotation/src/main/java/com.esri.samples.edit_features_with_feature_linked_annotation/EditFeaturesWithFeatureLinkedAnnotationSample.java
```diff
@@ -0,0 +1,350 @@
...
+
+      // set on click behaviour
```
Rachael-E: with moving the logic from the single use method directly into this listener, we can borrow some of the original method description and use it here instead:

```suggestion
      // select the nearest feature from where the user clicked, or move the selected feature to the given screen point
```

editing/edit-features-with-feature-linked-annotation/src/main/java/com.esri.samples.edit_features_with_feature_linked_annotation/EditFeaturesWithFeatureLinkedAnnotationSample.java
```diff
@@ -0,0 +1,350 @@
...
+      mapView.setOnMouseClicked(event -> {
+        // check that the primary mouse button was clicked
```
Rachael-E: we can remove this comment: the code below it is clear enough to describe it is checking the primary mouse button was clicked.
```suggestion
```

editing/edit-features-with-feature-linked-annotation/src/main/java/com.esri.samples.edit_features_with_feature_linked_annotation/EditFeaturesWithFeatureLinkedAnnotationSample.java
```diff
@@ -0,0 +1,350 @@
...
+        if (event.isStillSincePress() && event.getButton() == MouseButton.PRIMARY) {
+          // create a point where the user clicked
+          Point2D screenPoint = new Point2D(event.getX(), event.getY());
+          // call select or move method to move to the point
+          selectOrMove(screenPoint);
```
Rachael-E: it's common with the Android implementations to have single use methods, but with Java SE samples we don't do this as often. The logic in this method isn't too verbose, so let's just remove the method and insert the code from that method directly here instead. 
```suggestion
    // if a feature hasn't been selected, select the feature
    if (selectedFeature == null) {
      identifyFeature(screenPoint);
    } else {
      // convert the screen point to a map point
      Point mapPoint = mapView.screenToLocation(screenPoint);
      // if the feature is a polyline, move the polyline
      if (selectedFeatureIsPolyline) {
        movePolylineVertex(mapPoint);
      } else {
        // if the feature is a point, move the point
        movePoint(mapPoint);
      }
    }```

**Review by Rachael-E — CHANGES_REQUESTED** [review]
@sclaridge this is a really great effort on this rather complex sample! 

My comments here so far are more from a readability/code tidiness perspective and I think there's some more we can do here to neaten up this sample. Let's have a pair programming call and I can talk you through them. 

To discuss:
- Java SE samples and single use methods
- edge case user input control
- sample bug when moving polylines (it looks like additional polylines are being created, rather than just an individual line being moved)
- samples-data directory and some background there for the settings.gradle and build.gradle file (including removing the actual geodatabase from this repo)
- garbage collection

File: editing/edit-features-with-feature-linked-annotation/src/main/java/com.esri.samples.edit_features_with_feature_linked_annotation/EditFeaturesWithFeatureLinkedAnnotationSample.java [hunk 40]
```diff
@@ -0,0 +1,350 @@
...
+
+    // create a polyline builder to add and remove parts from the polyline
+    PolylineBuilder polylineBuilder = new PolylineBuilder(polyline);
+
+    // get the nearest vertex to the map point on the polyline
+    ProximityResult nearestVertex =
+      GeometryEngine.nearestVertex(polyline, (Point) GeometryEngine.project(mapPoint, polyline.getSpatialReference()));
+
+    // get the part of the polyline nearest to the map point
+    Part part = polylineBuilder.getParts().get((int) nearestVertex.getPartIndex());
```
Rachael-E: rather than casting this `long` to an `int` (which could result in a stack overflow if the `long` is greater than the maximum allowed `int` value), we can use the following method which would throw an informative exception in case of the above happening. Very unlikely in this sample but good practice I believe. 

```suggestion
    Part part = polylineBuilder.getParts().get(Math.toIntExact(nearestVertex.getPartIndex()));
```

editing/edit-features-with-feature-linked-annotation/src/main/java/com.esri.samples.edit_features_with_feature_linked_annotation/EditFeaturesWithFeatureLinkedAnnotationSample.java
```diff
@@ -0,0 +1,350 @@
...
+   */
+  private void selectFeature(IdentifyLayerResult layerResult) {
+
+    // if the selected feature is a polyline
+    if (selectedFeature.getGeometry().getGeometryType() == GeometryType.POLYLINE) {
+      // create a polyline builder from the selected feature
+      PolylineBuilder polylineBuilder = new PolylineBuilder((Polyline) selectedFeature.getGeometry());
+      // get a list of parts of the selected polyline
+      List<Part> parts = polylineBuilder.getParts();
+      parts.forEach(part -> {
```
Rachael-E: since we don't use the variable `parts` more than once, we can directly add the forEach like so:

```suggestion
      polylineBuilder.getParts().forEach(part -> {
```

editing/edit-features-with-feature-linked-annotation/src/main/java/com.esri.samples.edit_features_with_feature_linked_annotation/EditFeaturesWithFeatureLinkedAnnotationSample.java
```diff
@@ -0,0 +1,288 @@
...
+            geodatabase.getGeodatabaseFeatureTable("Loudoun_Address_Points_1"));
+          var parcelLinesFeatureLayer = new FeatureLayer(
+            geodatabase.getGeodatabaseFeatureTable("ParcelLines_1"));
+          // create annotation layers from tables in the geodatabase
+          var addressPointsAnnotationLayer = new AnnotationLayer(
+            geodatabase.getGeodatabaseAnnotationTable("Loudoun_Address_PointsAnno_1"));
+          var parcelLinesAnnotationLayer = new AnnotationLayer(
+            geodatabase.getGeodatabaseAnnotationTable("ParcelLinesAnno_1"));
+
+          // add the feature layers and annotation layers to the map
```
Rachael-E: just some wordsmithing here:

```suggestion
          // add the feature and annotation layers to the map
```

editing/edit-features-with-feature-linked-annotation/src/main/java/com.esri.samples.edit_features_with_feature_linked_annotation/EditFeaturesWithFeatureLinkedAnnotationSample.java
```diff
@@ -0,0 +1,288 @@
...
+
+              // add the map point as the new point on the part
+              part.addPoint((Point) GeometryEngine.project(mapPoint, polyline.getSpatialReference()));
+
+              // set the selected feature's geometry to the new polyline
+              selectedFeature.setGeometry(polylineBuilder.toGeometry());
+
+              // if the feature is a point, move the point
+            } else if (selectedFeature.getGeometry().getGeometryType() == GeometryType.POINT) {
+              // set the selected features' geometry to a new map point
```
Rachael-E: correcting apostrophe
```suggestion
              // set the selected feature's geometry to a new map point
```

editing/edit-features-with-feature-linked-annotation/src/main/java/com.esri.samples.edit_features_with_feature_linked_annotation/EditFeaturesWithFeatureLinkedAnnotationSample.java
```diff
@@ -0,0 +1,288 @@
...
+              selectedFeature.setGeometry(mapPoint);
+            }
+            // update the selected feature's feature table
+            updateAttributes(selectedFeature);
+
+            // clear the selected feature
```
Rachael-E: just a suggestion: I think this might be a littler clearer, but up to you if you want to accept this change! 
```suggestion
            // remove the selection of the feature
```

Rachael-E (approved):
@sclaridge this is looking great. I've suggested just a few really minor comment updates. 

Before merging, please can you ensure the rogue geodatabase is deleted from this PR? https://github.com/Esri/arcgis-runtime-samples-java/blob/924c240f55b27603be178aef26019340df599dc7/editing/edit-features-with-feature-linked-annotation/src/main/resources/edit_features_with_feature_linked_annotation/loudon/loudoun_anno.geodatabase

Once you've done this I'm happy for you to go ahead and merge. Good job!

PR #544, opened 2020-08-11
Updates to Readmes

jenmerritt:
This PR covers required readme updates.

**CS repo issue 2166**
- Animate Images with Image Overlay (Scene)
This was to update the 'How it works' section. However reviewing the common designs, Java was already a bit different before these additional changes, so have left as is for now. But would appreciate this being checked as not sure if there's a reason it was different in the first place. Can update to match the design if not.

- Feature Layer Shapefile (feature layers)
Updated How to Use section

- Token Authentication (portal)
Added Additional information section

**CS repo issue 2155**
- Feature Layer Selection (feature layers)
Reordering of 2 sections done

- List Transformations by suitability (geometry)
In this case I added in the entire Additional information section from the common samples repo, including the recent update from the Issue. Again I can amend this if it is supposed to be different for Java or there's back story that I'm not aware of.

### DIFF
--- a/geometry/list-transformations-by-suitability/README.md
+++ b/geometry/list-transformations-by-suitability/README.md
@@ -27,6 +27,22 @@ If the selected transformation is not usable (has missing grid files) then an er
 * GeometryEngine
 * TransformationCatalog
 
+## Additional information
+
+Some transformations aren't available until transformation data is provided.
+
+This sample uses a `GeographicTransformation`, which extends the `DatumTransformation` class. As of 100.9, ArcGIS Runtime also includes a `HorizontalVerticalTransformation`, which also extends `DatumTransformation`. The `HorizontalVerticalTrasnformation` class is used to transform coordinates of z-aware geometries between spatial references that have different geographic and/or vertical coordinate systems.
+
+This sample can be used with or without provisioning projection engine data to your device. If you do not provision data, a limited number of transformations will be available.
+
+To download projection engine data to your device:
+
+1. Login to the [ArcGIS for Developers site](https://developers.arcgis.com/sign-in/) using your Developer account.
+2. In the Dashboard page, click '[Download APIs and SDKs](https://developers.arcgis.com/downloads/data)' and go to the `Supplemental ArcGIS Runtime Data` tab.
+3. Click the download button next to `Projection Engine Data` to download projection engine data to your computer.
+4. Unzip the downloaded data on your computer.
+5. Create an `~/ArcGIS/Runtime/Data/PEDataRuntime` directory on your device and copy the files to this directory.
+
 ## About the data
 
 The map starts out zoomed into the grounds of the Royal Observatory, Greenwich. The initial point is in the [British National Grid](https://epsg.io/27700) spatial reference, which was created by the United Kingdom Ordnance Survey. The spatial reference after projection is in [web mercator](https://epsg.io/3857).
--- a/geometry/list-transformations-by-suitability/README.md
+++ b/geometry/list-transformations-by-suitability/README.md
@@ -29,19 +29,7 @@ If the selected transformation is not usable (has missing grid files) then an er
 
 ## Additional information
 
-Some transformations aren't available until transformation data is provided.
-
-This sample uses a `GeographicTransformation`, which extends the `DatumTransformation` class. As of 100.9, ArcGIS Runtime also includes a `HorizontalVerticalTransformation`, which also extends `DatumTransformation`. The `HorizontalVerticalTrasnformation` class is used to transform coordinates of z-aware geometries between spatial references that have different geographic and/or vertical coordinate systems.
-
-This sample can be used with or without provisioning projection engine data to your device. If you do not provision data, a limited number of transformations will be available.
-
-To download projection engine data to your device:
-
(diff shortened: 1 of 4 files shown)


geometry/list-transformations-by-suitability/README.md
```diff
@@ -27,6 +27,22 @@ If the selected transformation is not usable (has missing grid files) then an er
 * GeometryEngine
 * TransformationCatalog
 
+## Additional information
+
+Some transformations aren't available until transformation data is provided.
+
+This sample uses a `GeographicTransformation`, which extends the `DatumTransformation` class. As of 100.9, ArcGIS Runtime also includes a `HorizontalVerticalTransformation`, which also extends `DatumTransformation`. The `HorizontalVerticalTrasnformation` class is used to transform coordinates of z-aware geometries between spatial references that have different geographic and/or vertical coordinate systems.
```
JonLavi: Should this be `Z-aware`? Not sure when you capitalise.

```suggestion
This sample uses a `GeographicTransformation`, which extends the `DatumTransformation` class. As of 100.9, ArcGIS Runtime also includes a `HorizontalVerticalTransformation`, which also extends `DatumTransformation`. The `HorizontalVerticalTrasnformation` class is used to transform coordinates of Z-aware geometries between spatial references that have different geographic and/or vertical coordinate systems.
```

JonLavi (approved):
Thanks for updating these @jenmerritt!
The AnimateImagesWithImageOverlay sample should be fine. Particularly with that sample, every SDK has a slightly different approach to rigging up the sample. What we have there represents the code well so we don't have to update.

geometry/list-transformations-by-suitability/README.md
jenmerritt: Not 100%, seems to be z-aware when I did a search on documentation, but refer to a Z when there's not a hyphen. I'll leave as is based on the documentation check.

Rachael-E: It looks like in our sample we don't use the `GeographicTransformation` class: just the `DatumTransformation` class. Whenever I do a ReadMe update which contains new classes, I always do a quick "Find..." within the sample just to sanity check if we use the class from the generic sample description or not: samples can vary slightly between SDKs and also in name, so it's a good habit to get into! 

What is the purpose of the addition of this additional information here: is it intended that this sample use the `HorizontalVerticalTransformation`? Does the code require updating? If not then we can just reword to avoid specifically saying that our samples uses the `GeographicTransformation`: but you should probably double check the wording with the author of the common sample readme to be sure this still delivers the message intended. 

```suggestion
As of 100.9, ArcGIS Runtime also includes a `HorizontalVerticalTransformation`, which extends `DatumTransformation`. The `HorizontalVerticalTrasnformation` class is used to transform coordinates of z-aware geometries between spatial references that have different geographic and/or vertical coordinate systems.
```

geometry/list-transformations-by-suitability/README.md
```diff
@@ -27,6 +27,22 @@ If the selected transformation is not usable (has missing grid files) then an er
...
+
+This sample can be used with or without provisioning projection engine data to your device. If you do not provision data, a limited number of transformations will be available.
+
+To download projection engine data to your device:
+
+1. Login to the [ArcGIS for Developers site](https://developers.arcgis.com/sign-in/) using your Developer account.
+2. In the Dashboard page, click '[Download APIs and SDKs](https://developers.arcgis.com/downloads/data)' and go to the `Supplemental ArcGIS Runtime Data` tab.
+3. Click the download button next to `Projection Engine Data` to download projection engine data to your computer.
+4. Unzip the downloaded data on your computer.
+5. Create an `~/ArcGIS/Runtime/Data/PEDataRuntime` directory on your device and copy the files to this directory.
```
Rachael-E: it's possible we need to refine this further for Java: having a look at our `TransformationCatalog` class, there is guidance there for where the default location for such projection info is:

https://developers.arcgis.com/java/latest/api-reference/reference/com/esri/arcgisruntime/geometry/TransformationCatalog.html#getProjectionEngineDirectory()

Note that it is also possible to set the directory location:
https://developers.arcgis.com/java/latest/api-reference/reference/com/esri/arcgisruntime/geometry/TransformationCatalog.html#setProjectionEngineDirectory(java.lang.String)

in which case we'd need to say something like the following, but I'd like a second opinion from @colinanderson here first please to sanity check what I'm saying! 

```suggestion
5. Find the `<application directory>/resources/pedata` directory on your device and copy the files to this directory.
```

Rachael-E (changes requested):
thanks @jenmerritt, I've left two changes that we need to fix up: basically just ensuring the updated readme instructions do indeed match exactly what we do on Java and to make sure the guidance we give is Java specific. The second one I'd like a second opinion from @colinanderson on before continuing, thank you!

geometry/list-transformations-by-suitability/README.md
jenmerritt: As discussed with Rachael/Mark, the data in that directory is downloaded with the SDK for Java so not needed in the readme

geometry/list-transformations-by-suitability/README.md
jenmerritt: Checked these queries on CS Issue. And amended as required.

Rachael-E (approved):
looks good to me, thanks @jenmerritt !

PR #546, opened 2020-08-18
update portal item id

TADraeseke:
Update Open Scene Portal data:

1. ItemId 31874da8a16d45bfbc1273422f772270
2. README - Additional information link "This sample uses a Berlin, Germany scene hosted on ArcGIS Online."
3. Screenshot in Readme


sclaridge (commented):
This looks a lot better! I think the README just needs an updated image.

jenmerritt:
For me it's like the elevation isn't loading in the Scene, as the buildings appear to be floating... Setting a surface fixes it, however, the previous scene layer elevation was fine and there was no need to set a surface for it.

Not sure if this is Java specific / if any other platforms see this. (It's the same in 100.8.0)

image

Rachael-E:
might be to do with the portal data: when looking at the Geneva data in ArcGISOnline I see a "Topography" entry with "World Topo Map",  but not one for Montreal (instead "Terrain 3D" is listed under "Layers"). To access the terrain in layers, we might have to add in some extra code to add a surface? Or is it possible to update the ArcGISOnline item to include a Topography entry: assuming that is the problem?

image

sclaridge:
The same "floating" behaviour is observed when viewing the Scene in Pro, it looks like the Terrain 3D elevation layer is missing:  
image

jenmerritt:
Confirmed required changes after some additional discussions around layers:

Update Open Scene Portal data:

1. ItemId 31874da8a16d45bfbc1273422f772270
2. README - Additional information link "This sample uses a Berlin, Germany scene hosted on ArcGIS Online." 
3. Screenshot in Readme

Rachael-E (approved):
looks good to me!

JonLavi (approved):

PR #547, opened 2020-08-25
Remove initowner from dialog

Rachael-E:
### Description

As discussed with @mbcoder, removing initOwner from this dialog as it is causing conflicts with the sample viewer. @JonLavi @jenmerritt can you review please? I have tested this both locally via the IDE and via the sample viewer on Mac and it works as expected. Please could you sanity check the dialog still appears as expected for you on Windows? Thanks. 

Checklist:
- [x] Request a reviewer
- [x] Assign a reviewer
- [x] Tag reviewer in comment

### DIFF
--- a/feature_layers/feature-layer-query/src/main/java/com/esri/samples/feature_layer_query/FeatureLayerQuerySample.java
+++ b/feature_layers/feature-layer-query/src/main/java/com/esri/samples/feature_layer_query/FeatureLayerQuerySample.java
@@ -100,7 +100,6 @@ public class FeatureLayerQuerySample extends Application {
 
       // create dialog to display alert information
       dialog = new Alert(AlertType.WARNING);
-      dialog.initOwner(stackPane.getScene().getWindow());
 
       // search for the state that was entered
       searchButton.setOnAction(e -> {


jenmerritt (approved):
Tested Feature Layer Query Sample on Windows.

- IDE - dialog displays when entering blank state or incorrect state (works as it did before)
- SV - dialog now displays when entering blank state or incorrect state. NullPointerException no longer occurrs.

So looks good to me!

JonLavi:
For some weird reason I can't leave a proper review (so just merge if you need to)!

I've tested with adding 
```
dialog.initOwner(Stage.getWindows().stream().filter(Window::isShowing).findFirst().orElse(null));
```

and it seems to work fine in both the SV and the isolated sample.

This probably works because `Stage` can be found when running from the SV as it then grabs the SV's stage.

I don't know if this is necessary at all still, so just putting it out there.

colinanderson:
I doubt this is the right fix. I suspect you were setting the owner to null because the way the sample viewer creates the node means it doesn't have a window when you were calling `stackPane.getScene().getWindow()`. Not having an owner might have changed the modality of your dialog. See https://stackoverflow.com/questions/37181445/initowner-role-javafx.

colinanderson:
At least two differences when there is no owner
- the dialog does not get the app's icon.
- the dialog appears as a separate app on the task bar e.g.
image

PR #554, opened 2020-09-11
Update style WMS layers to use correct spatial reference

sclaridge:
This PR resolves issue #997:

- The service doesn't support the spatial reference of the map. They layer should have failed to load
- The service is generating incorrect images when the view includes polar regions; this is a service issue.

**Resolution:**

- Update sample to set a MinScale of 7,000,000 - avoids zooming to polar regions
- Update sample to use SR 26915
- Update sample to use no basemap (layer draws white over basemap anyway)

### DIFF
--- a/ogc/style-wms-layer/src/main/java/com/esri/samples/style_wms_layer/StyleWmsLayerSample.java
+++ b/ogc/style-wms-layer/src/main/java/com/esri/samples/style_wms_layer/StyleWmsLayerSample.java
@@ -55,9 +55,9 @@ public class StyleWmsLayerSample extends Application {
       stage.setScene(scene);
       stage.show();
 
-      // create a map with spatial reference appropriate for the service
+      // create a map with spatial reference appropriate for the service (North American Datum 83)
       ArcGISMap map = new ArcGISMap(SpatialReference.create(26915));
-      map.setMinScale(7000000.0);
+      map.setMinScale(7000000);
       // set the map to the map view
       mapView = new MapView();
       mapView.setMap(map);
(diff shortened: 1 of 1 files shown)


JonLavi (approved):
Thanks for putting in a fix for this @sclaridge !
Looks all good to me!

ogc/style-wms-layer/src/main/java/com/esri/samples/style_wms_layer/StyleWmsLayerSample.java
```diff
@@ -55,8 +55,10 @@ public void start(Stage stage) {
       stage.setScene(scene);
       stage.show();
 
-      // create a map and add it to the map view
-      ArcGISMap map = new ArcGISMap(Basemap.createImagery());
+      // create a map with spatial reference appropriate for the service
```
Rachael-E: since these spatial reference id's can be quite obscure, I like to give a little more context in the comments:
```suggestion
      // create a map with spatial reference appropriate for the service (North American Datum 83)
```

ogc/style-wms-layer/src/main/java/com/esri/samples/style_wms_layer/StyleWmsLayerSample.java
```diff
@@ -55,8 +55,10 @@ public void start(Stage stage) {
...
+      ArcGISMap map = new ArcGISMap(SpatialReference.create(26915));
+      map.setMinScale(7000000.0);
```
Rachael-E: let's drop the decimal and keep it as a whole number (in keeping with other setting scales in the repo)
```suggestion
      map.setMinScale(7000000);
```

Rachael-E (approved):
@sclaridge thanks, I've made two small suggestions and once you're happy with those you can go ahead and merge!

PR #555, opened 2020-09-16
List related features sample: use correct identify workflow

jenmerritt:
### Description

Original request (CS 2032): Several samples use an identify workflow that is not best practice. We should be using `mapView.identifyLayersAsync(...)` when identifying across multiple layers or `mapView.identifyLayerAsync(...)` when we know which layer we need to run the identify on.

For Java the sample needing to be update is List Related Features. I've made what I think is the correct change, but @TADraeseke  if you could check this over first as per your original issue requirements, I can then get it reviewed for merging.

Thanks!

Checklist:
- [x] Request a reviewer
- [x] Assign a reviewer
- [x] Tag reviewer in comment

### DIFF
--- a/feature_layers/list-related-features/src/main/java/com/esri/samples/list_related_features/ListRelatedFeaturesSample.java
+++ b/feature_layers/list-related-features/src/main/java/com/esri/samples/list_related_features/ListRelatedFeaturesSample.java
@@ -16,7 +16,6 @@
 
 package com.esri.samples.list_related_features;
 
-import java.util.Iterator;
 import java.util.List;
 import java.util.concurrent.ExecutionException;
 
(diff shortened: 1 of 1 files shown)


TADraeseke (approved):
LGTM! Thanks @jenmerritt!

jenmerritt:
@JonLavi @Rachael-E  very small update to list related features sample if you could both take a look, thanks!

feature_layers/list-related-features/src/main/java/com/esri/samples/list_related_features/ListRelatedFeaturesSample.java
```diff
@@ -102,44 +99,43 @@ public void start(Stage stage) {
...
 
               try {
-                FeatureQueryResult result = selectFeatureQuery.get();
+                IdentifyLayerResult identifiedLayer = identifyLayerResult.get();
+                List<GeoElement> identifiedLayerGeoElements = identifiedLayer.getElements();
                 // get the first selected feature
-                Iterator<Feature> iterator = result.iterator();
-                if (iterator.hasNext()) {
-                  ArcGISFeature selectedFeature = (ArcGISFeature) iterator.next();
+                for (var geoElement : identifiedLayerGeoElements) {
```
JonLavi: For the sake of being extra clear which class we are dealing with here, we should write:
```suggestion
                for (GeoElement geoElement : identifiedLayerGeoElements) {
```

There's no harm in using `var` generally, but for samples we want to make sure our users know exactly which class they are dealing with at any point.

JonLavi (approved):
Nice update to this sample. Only thing I'm noticing is the use of `var`, which we shouldn't really do in samples. We want to be as verbose as possible. 
Approving so you can move on to a second/third review 👍

feature_layers/list-related-features/src/main/java/com/esri/samples/list_related_features/ListRelatedFeaturesSample.java
Rachael-E: I agree with Jon here: when using `var`, it's typically to improve readability in the case of a super long class name, or multiple blocks of repeated class types. In this for loop, it's better to err on the side of readability with `GeoElement`.

jenmerritt: Understood, thanks both for the explanation. I've changed that @Rachael-E

Rachael-E (approved):

PR #556, opened 2020-09-17
New sample: Show location history

sclaridge:
This PR is for the Show location history sample, showing off the `SimulatedLocationDataSource` 
 with some test location data. The tracking is toggled on and off with a button. When the user pans away from the tracking, the tracking pauses until they pan back or stop the tracking. 

* inside `simulatedLocationDataSource.addLocationChangedListener`, the next point in the tracking route is set using `locationChangedEvent.getLocation().getPosition()`. This results in the tracking to move ahead of the `LocationDisplay` symbol. Is this ok?
* Alternatively, using `mapView.getLocationDisplay().getMapLocation()` causes the tracking to follow the `LocationDisplay`symbol, however when the user pans away from the location display additional lines are added to the tracking route.
* Does the logic for creating the tracking points and lines need to be in a `try` block?

Thanks!

@JonLavi


map_view/show-location-history/README.md
```diff
@@ -0,0 +1,43 @@
+# Show location history
+
+Display your location history on the map.
+
+![](ShowLocationHistory.png)
```
JonLavi: Don't forget to add a screenshot (and maybe even a fancy gif for the SV?)

map_view/show-location-history/README.md
```diff
@@ -0,0 +1,43 @@
...
+* LocationDataSource.LocationChangedEvent
+* LocationDataSource.LocationChangedListener
+* LocationDisplay.AutoPanMode
+* MapView.LocationDisplay
+* SimulatedLocationDataSource
+* SimulationParameters
+
+## About the data
+
+The sample uses a dark gray basemap with lime and red graphics.
```
JonLavi: I don't think we need this description, it's quite obvious and the type of basemap/graphics aren't really that relevant to the way the API works

map_view/show-location-history/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+![](ShowLocationHistory.png)
+
+## Use case
+
+You can track device location history and display it as lines and points on the map. The history can be used to visualize how the user moved through the world, to retrace their steps, or to create new feature geometry. An unmapped trail, for example, could be added to the map using this technique.
+
+## How to use the sample
+
+Click the button to start tracking your location, which will appear as points on the map. A line will connect the points for easier visualization. Click the button again to stop updating the location history.
```
JonLavi: I might suggest adding a mention of the simulated location data source here already, since this section by itself would give the users the impression they are seeing some sort of 'live' location?

map_view/show-location-history/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+## How it works
+
+1. If necessary, request location permission from the operating system.
```
JonLavi: We don't have this in Java, so I think we can leave this out. A potential sample with the NMEA api would handle passing in 'real' data into `LocationDataSource`

map_view/show-location-history/README.md
```diff
@@ -0,0 +1,43 @@
...
+2. Create a `GraphicsOverlay` to show each point and another `GraphicsOverlay` for displaying the route line.
+3. Create a `SimulatedLocationDataSource` and initialize it with a polyline. Start the `SimulatedLocationDataSource` to begin receiving location updates.
+4. Use a `LocationChangedListener` on the `simulatedLocationDataSource` to get location updates.
+5. When the location updates store that location, display a point on the map at the location, and re-create the route polyline.
+
+## Relevant API
+
+* LocationDataSource.Position
+* LocationDataSource
```
JonLavi: We could put this one at the top of the list, so they are sorted 'alphabetically'.
Not that there's a general tendenct only to put top-level classes in here, but since the inner classes `Location`, `LocaitonChangedEvent`, etc. are all quite important, this is ok.

map_view/show-location-history/README.metadata.json
```diff
@@ -0,0 +1,45 @@
+{
+    "category": "Map view",
+    "description": "Display your location history on the map.",
+    "ignore": false,
+    "images": [
+        "ShowLocationHistory.jpg"
```
JonLavi: This should say 
```suggestion
        "ShowLocationHistory.png"
```

If you haven't been using the metadata script to generate this file, it's worth giving it a try, since it does deal with quite a lot of the things that may have typos in it!

map_view/show-location-history/build.gradle
```diff
@@ -0,0 +1,75 @@
+plugins {
+    id 'application'
+    id 'org.openjfx.javafxplugin' version '0.0.5'
+}
+
+group = 'com.esri.samples'
+
+ext {
+    arcgisVersion = '100.10.0-2902'
```
JonLavi, line 9: I _think_ this API is already available in 100.9.0, so might be able to merge this straight into master. Depends on whether @TADraeseke and @mbcoder think that's worth doing without the NMEA API and sample also being available...

map_view/show-location-history/src/main/java/com/esri/samples/show_location_history/ShowLocationHistorySample.java
```diff
@@ -0,0 +1,201 @@
...
+
+      // set title, size, and add scene to stage
+      stage.setTitle("Show Location History Sample");
+      stage.setWidth(800);
+      stage.setHeight(700);
+      stage.setScene(scene);
+      stage.show();
+
+      // create a map with a dark gray canvas basemap
+      ArcGISMap map = new ArcGISMap("https://www.arcgis.com/home/item.html?id=1970c1995b8f44749f4b9b6e81b5ba45");
```
JonLavi: instead of referring to the URL, you can simply do:`
```suggestion
      ArcGISMap map = new ArcGISMap(Basemap.createDarkGrayCanvasVector());
```

Standard basemaps all have a method on the `Basemap` class to create them:
https://developers.arcgis.com/java/latest/api-reference/reference/com/esri/arcgisruntime/mapping/Basemap.html

map_view/show-location-history/src/main/java/com/esri/samples/show_location_history/ShowLocationHistorySample.java
```diff
@@ -0,0 +1,201 @@
...
+      trackingButton.setDisable(true);
+
+      // enable the button interactions when the map is loaded
+      map.addDoneLoadingListener(() -> {
+        if (map.getLoadStatus() == LoadStatus.LOADED) {
+          trackingButton.setDisable(false);
+        } else {
+          new Alert(Alert.AlertType.ERROR, "Map failed to load").show();
+        }
+      });
```
JonLavi: I'd suggest perhaps moving this beneath the creation of the GraphicsOverlays and Polyline Builder. That way we have all the parts up and ready before want to make use of them inside the map. 

Then, perhaps we should move the parts that set up the `SimulatedLocationDataSource` and configure the `LocationDisplay` (lines 122-154) inside this load-done check. After all, there's not much of a reason to start the LocationDisplay if our map is not working.

map_view/show-location-history/src/main/java/com/esri/samples/show_location_history/ShowLocationHistorySample.java
```diff
@@ -0,0 +1,201 @@
...
+      // configure the map view's location display to follow the simulated location data source
+      mapView.getLocationDisplay().setLocationDataSource(simulatedLocationDataSource);
+      mapView.getLocationDisplay().setAutoPanMode(LocationDisplay.AutoPanMode.RECENTER);
+      mapView.getLocationDisplay().setInitialZoomScale(7000);
+
+      trackingButton.setOnAction(event -> {
+        // if the user has panned away from the location display, turn it on again
+        if (mapView.getLocationDisplay().getAutoPanMode() == LocationDisplay.AutoPanMode.OFF) {
+          mapView.getLocationDisplay().setAutoPanMode(LocationDisplay.AutoPanMode.RECENTER);
+        }
```
JonLavi: I find the interaction with panning and switching the auto-pan on and off is not that intuitive? It doesn't help that the auto-pan functionality is not continuous, but only happens when the dot is approaching the edge of the screen. Not sure if this can be adjusted somehow.

 Perhaps this sample could benefit from a tick-box to enable/disable auto-pan? When the tick-box would be checked, the manual panning with the mouse should be disabled (or, panning with the mouse would de-activate that tickbox). What do you think?

map_view/show-location-history/src/main/java/com/esri/samples/show_location_history/ShowLocationHistorySample.java
```diff
@@ -0,0 +1,201 @@
...
+      // create a simulated location data source
+      SimulatedLocationDataSource simulatedLocationDataSource = new SimulatedLocationDataSource();
+      // set the location of the simulated location data source with simulation parameters to set a consistent velocity
+      simulatedLocationDataSource.setLocations(
+        routePolyline, new SimulationParameters(Calendar.getInstance(), 30.0, 0.0, 0.0));
+
+      // start the simulated location data source
+      simulatedLocationDataSource.startAsync();
+      simulatedLocationDataSource.addLocationChangedListener(locationChangedEvent -> {
+        // if location tracking is turned off, do not add to the polyline
```
JonLavi: Technically we'll be filling the gap of the polyline when we start tracking again. So perhaps this comment should say 'do not draw points or extend the polyline'? Something along those lines

map_view/show-location-history/src/main/java/com/esri/samples/show_location_history/ShowLocationHistorySample.java
```diff
@@ -0,0 +1,201 @@
...
+      // add the graphics overlays to the map view
+      mapView.getGraphicsOverlays().addAll(Arrays.asList(locationHistoryOverlay, locationHistoryLineOverlay));
+
+      // create a polyline builder to connect the location points
+      PolylineBuilder polylineBuilder = new PolylineBuilder(SpatialReference.create(3857));
+      // access the json of the location points
+      String polylineData = IOUtils.toString(getClass().getResourceAsStream("/show_location_history/polyline_data.json"),
+        StandardCharsets.UTF_8);
+      // create a polyline from the location points
+      Polyline routePolyline = (Polyline) Polyline.fromJson(polylineData,
```
JonLavi: This is a bit of a tricky statement. If you're creating a `Polyline.fromJson`, then there's technically no need to cast it to a `Polyline`, since it's an object of that class already.

But, we can't do ` Polyline routePolyline = Polyline.fromJson(),`, because the return type of that is a 'Geometry'. This is because that method is indirectly referring to `Geometry.fromJson`, which is the type from which `Polyline` inherits. 

So to keep with the class structure, it's best to do the following:

```suggestion
      Polyline routePolyline = (Polyline) Geometry.fromJson(polylineData,
```

That way we're accessing the method `fromJson` at it's lowest level, and trying to cast the result to a Polygon. If there's an error somewhere there, the message might also make it a little easier to understand why something broke (either with creating a generic `Geometry` from the Json, or with casting it to a Polyline)

map_view/show-location-history/src/main/java/com/esri/samples/show_location_history/ShowLocationHistorySample.java
```diff
@@ -0,0 +1,201 @@
...
+
+      // configure the map view's location display to follow the simulated location data source
+      mapView.getLocationDisplay().setLocationDataSource(simulatedLocationDataSource);
+      mapView.getLocationDisplay().setAutoPanMode(LocationDisplay.AutoPanMode.RECENTER);
+      mapView.getLocationDisplay().setInitialZoomScale(7000);
+
+      trackingButton.setOnAction(event -> {
+        // if the user has panned away from the location display, turn it on again
+        if (mapView.getLocationDisplay().getAutoPanMode() == LocationDisplay.AutoPanMode.OFF) {
+          mapView.getLocationDisplay().setAutoPanMode(LocationDisplay.AutoPanMode.RECENTER);
```
JonLavi: Instead of repeatedly querying the mapView's LocationDisplay, we can save it to a variable. Although it adds a line of code, it makes the other lines a bit clearer:

```suggestion
	  LocationDisplay locationDisplay = mapView.getLocationDisplay();
      locationDisplay.setLocationDataSource(simulatedLocationDataSource);
      locationDisplay.setAutoPanMode(LocationDisplay.AutoPanMode.RECENTER);
      locationDisplay.setInitialZoomScale(7000);

      trackingButton.setOnAction(event -> {
        // if the user has panned away from the location display, turn it on again
        if (locationDisplay.getAutoPanMode() == LocationDisplay.AutoPanMode.OFF) {
          locationDisplay.setAutoPanMode(LocationDisplay.AutoPanMode.RECENTER);
```

map_view/show-location-history/src/main/java/com/esri/samples/show_location_history/ShowLocationHistorySample.java
```diff
@@ -0,0 +1,201 @@
...
+        }
+        // toggle the location tracking when the button is clicked
+        if (isTrackingEnabled) {
+          trackingButton.setText("Start Tracking");
+          isTrackingEnabled = false;
+        } else {
+          trackingButton.setText("Stop Tracking");
+          isTrackingEnabled = true;
+        }
```
JonLavi: There's absolutely nothing wrong with the way this is done here (arguably it's easier to read), but if you wanted to save yourself one line, a cheeky way of setting a value to it's opposite would be: 
```suggestion
        if (isTrackingEnabled) {
          trackingButton.setText("Start Tracking");
        } else {
          trackingButton.setText("Stop Tracking");
        }
        isTrackingEnabled = !isTrackingEnabled;
```

map_view/show-location-history/src/main/java/com/esri/samples/show_location_history/ShowLocationHistorySample.java
```diff
@@ -0,0 +1,201 @@
...
+      });
+
+      // add the map view and tracking button to the stack pane
+      stackPane.getChildren().addAll(mapView, trackingButton);
+      StackPane.setAlignment(trackingButton, Pos.TOP_LEFT);
+      StackPane.setMargin(trackingButton, new Insets(10, 0, 0, 10));
+    } catch (Exception e) {
```
JonLavi, line 206: I think a newline here would be good to separate this from the block above
```suggestion
    
    } catch (Exception e) {
```

map_view/show-location-history/src/main/java/com/esri/samples/show_location_history/ShowLocationHistorySample.java
```diff
@@ -0,0 +1,201 @@
...
+
+      // create a map with a dark gray canvas basemap
+      ArcGISMap map = new ArcGISMap("https://www.arcgis.com/home/item.html?id=1970c1995b8f44749f4b9b6e81b5ba45");
+
+      // create a map view and set its map
+      mapView = new MapView();
+      mapView.setMap(map);
+
+      // set the map views's viewpoint centered on Los Angeles, California and scaled
+      Point center = new Point(-13185535.98, 4037766.28, SpatialReference.create(3857));
```
JonLavi: This WKID corresponds to WebMercator, which we also have a utility method for (just makes it a bit more clear which reference we'll be using):
```suggestion
      Point center = new Point(-13185535.98, 4037766.28, SpatialReferences.getWebMercator());
```

map_view/show-location-history/src/main/java/com/esri/samples/show_location_history/ShowLocationHistorySample.java
```diff
@@ -0,0 +1,201 @@
...
+      mapView.setViewpoint(new Viewpoint(center, 7000));
```
JonLavi: If the variable `point` above is not being re-used, we could consider writing:
```suggestion
      mapView.setViewpoint(new Viewpoint(new Point(-13185535.98, 4037766.28, SpatialReferences.getWebMercator()), 7000));
```

It's a bit of a mouthfull, but some of the other samples do this as well. Either way works though, so it's just a matter of preference!

map_view/show-location-history/src/main/java/com/esri/samples/show_location_history/ShowLocationHistorySample.java
```diff
@@ -0,0 +1,201 @@
...
+      SimpleRenderer locationHistoryLineRenderer = new SimpleRenderer(locationLineSymbol);
+      locationHistoryLineOverlay.setRenderer(locationHistoryLineRenderer);
+
+      // add the graphics overlays to the map view
+      mapView.getGraphicsOverlays().addAll(Arrays.asList(locationHistoryOverlay, locationHistoryLineOverlay));
+
+      // create a polyline builder to connect the location points
+      PolylineBuilder polylineBuilder = new PolylineBuilder(SpatialReference.create(3857));
+      // access the json of the location points
+      String polylineData = IOUtils.toString(getClass().getResourceAsStream("/show_location_history/polyline_data.json"),
```
JonLavi: As you've already mentioned, it might be worth investigating a try-catch for this particular part of the code. The reason is that if there's every a problem getting this file (for example, the Gradle script didn't download it properly), we'd get a meaningful error in the catch block. At the moment the code would defer to the catch at line 175. Have a try in structuring it in a way that we display an error to the user, saying there was a problem getting the simulated data.

JonLavi (commented):
Nice work @sclaridge .
I've made a few suggestions for re-structuring this a little. 
You're right in thinking the creating of the polyline should happen in a try/catch block. It's generally good practice to do this whenever we're reading/writing from a file or to a network (stuff usually goes wrong there!). 

I see the problems with the AutoPanMode and the point being slightly behind the location marker.

Perhaps adding a checkbox for the auto-pan mode would help with any sort of confusion regarding AutoPan?
 
For the polygon drawing, it looks as if the Location is updated before the marker moves to that spot on the map? Have other APIs implemented this sample in a similar way? Perhaps we'd need to add a small delay for drawing the polyline, but this doesn't seem like a user-friendly or intuitive solution. Let's do some investigating.

map_view/show-location-history/src/main/java/com/esri/samples/show_location_history/ShowLocationHistorySample.java
sclaridge: Sounds good! Do you think I should only catch IOExceptions? I imagine the try-catch with this block of code will be above the `map.addDoneLoadingListener`?

map_view/show-location-history/src/main/java/com/esri/samples/show_location_history/ShowLocationHistorySample.java
sclaridge: I agree, I think a checkbox toggling the auto-pan would be more intuitive. 

However, when the auto-pan is off the issue with the extra polylines still occurs when the user pans away.

map_view/show-location-history/README.md
sclaridge: Perhaps I could also add a line like: "To track a user's real position, use the `LocationDataSource` instead"?

map_view/show-location-history/src/main/java/com/esri/samples/show_location_history/ShowLocationHistorySample.java
JonLavi: Not sure which exceptions we'd need to handle, but the IDE will tell you if you're missing any. If there are more than say 3 specific exception types, we can simply do a generic `catch (Exception e)`, since all we're really doing is displaying an error, not working with the internals of the individual exceptions.

For the structure: Yeah I think above is probably a good place to put it. We'd want that polyline ready to go before we do anything particular with the Map and MapView.

map_view/show-location-history/README.md
JonLavi: Yes, that's a good call. It's important to make clear to the users that they are not actually driving through LA when they run this sample ;)

map_view/show-location-history/src/main/java/com/esri/samples/show_location_history/ShowLocationHistorySample.java
JonLavi: I'll have a closer look at the code. I didn't notice any connection between the drawing and listening to panning on the map, but it's possible that one thing is triggering another?

sclaridge: It happens when `position = mapView.getLocationDisplay().getMapLocation()` so that the LocationDisplay symbol moves ahead of the tracking. When the position is set this way and auto-pan is disabled, if the user pans away then back to the tracking, extra graphics are drawn. 

image

sclaridge: So I think this fixes the issue with the extra polylines, and also sets the `LocationDisplay` marker ahead of the tracking graphics:
```
// get the position as a point from the simulated location data source
Integer nextPoint = simulatedLocationDataSource.getCurrentLocationIndex()-1;
LocationDataSource.Location nextLocation = simulatedLocationDataSource.getLocations().get(nextPoint);
Point position = new Point(nextLocation.getPosition().getX(),nextLocation.getPosition().getY());
```

JonLavi: I was thinking of something similar, but that feels a little counterintuitive from a user perspective.
I'd check whether any of the other platforms are doing something similar. If they aren't specifically working around this, then we may have a small bug in the LocationDisplay API on Java.

map_view/show-location-history/README.md
JonLavi: Sorry, what I meant here was the 'alt' text is missing. It should say:
`Image of show location history`

sclaridge:
@JonLavi Here are the changes as discussed in our chat! I've moved the `catch` block above the code that adds the button and map view to the stack pane, similar to the Map Image Layer Sublayer Visibility sample.

I've set the `position = locationChangedEvent.getLocation().getPosition()` until I receive more clarity on this from the Argo team. I'll hold off on making a gif until then as well.

map_view/show-location-history/build.gradle
sclaridge: After discussing this with @TADraeseke, this sample can be merged into dev so that it can be released alongside the NMEA API.

map_view/show-location-history/README.md
```diff
@@ -6,37 +6,34 @@ Display your location history on the map.
 
 ## Use case
 
-You can track device location history and display it as lines and points on the map. The history can be used to visualize how the user moved through the world, to retrace their steps, or to create new feature geometry. An unmapped trail, for example, could be added to the map using this technique.
+You can track device location history and display it as lines and points on the map. The history can be used to visualize how the user moved through the world, to retrace their steps, or to create new feature geometry. For example, an unmapped trail could be added to the map using this technique.
 
 ## How to use the sample
 
-Click the button to start tracking your location, which will appear as points on the map. A line will connect the points for easier visualization. Click the button again to stop updating the location history.
+Click the button to start tracking your location, which will appear as points on the map. A line will connect the points for easier visualization. Click the button again to stop updating the location history. This sample uses a simulated data source to allow the sample to be useful on desktop/non-mobile devices.
```
JonLavi: I think it'd be more percise to say `This sample uses a simulated data source to allow the sample to be useful on devices without a location data source`. This is since just because a device is desktop, doesn't mean it won't have the means of getting it's location. For example most desktop devices can get a rough location via WiFi, etc.

JonLavi (approved):
Looks good to me @sclaridge ! 
Perhaps we can be a bit more specific about the use case for SimulatedLocationDataSource, but that's being a bit pedantic 😁

map_view/show-location-history/README.md
sclaridge: Sounds good! I could also add this to the use case: `You can use an existing dataset in place of a connection to a GPS device. A simulated data source is useful if you're giving a demonstration or are working with previously collected data.` from [Setting up the simulation](https://desktop.arcgis.com/en/arcmap/10.3/map/display-gps-data/setting-up-the-simulation.html). But that could be redundant!

sclaridge:
@JonLavi I've added a potential fix for the location display symbol issue. A `LocationChangedListener` is created, then when the button is toggled, it is added and removed from the `simulatedLocationDataSource`. 

Let me know what you think! I believe this is similar to how the .NET sample gets around this issue.

map_view/show-location-history/src/main/java/com/esri/samples/show_location_history/ShowLocationHistorySample.java
```diff
@@ -0,0 +1,218 @@
...
+        simulatedLocationDataSource.setLocations(
+          routePolyline, new SimulationParameters(Calendar.getInstance(), 30.0, 0.0, 0.0));
+
+        // configure the map view's location display to follow the simulated location data source
+        LocationDisplay locationDisplay = mapView.getLocationDisplay();
+        locationDisplay.setLocationDataSource(simulatedLocationDataSource);
+        locationDisplay.setAutoPanMode(LocationDisplay.AutoPanMode.RECENTER);
+        locationDisplay.setInitialZoomScale(7000);
+
+        // track previous location to ensure the route line appears behind the location display symbol
```
JonLavi: Maybe we can be a bit more descriptive about what this block does. The second part of the sentence sounds a bit like we're admitting it's a workaround, so perhaps we can reword it slightly to say this is deliberate?

```suggestion
        // create a listener to track the previous location, to draw a route line behind the location display symbol
```

JonLavi (approved):
Nice work @sclaridge .
I think the sample looks good now, just a suggestion for a comment on the LocationChangedListener

map_view/show-location-history/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+![Image of show location history](ShowLocationHistory.png)
+
+## Use case
+
+You can track device location history and display it as lines and points on the map. The history can be used to visualize how the user moved through the world, to retrace their steps, or to create new feature geometry. For example, an unmapped trail could be added to the map using this technique.
+
+## How to use the sample
+
+Click the button to start tracking your location, which will appear as points on the map. A line will connect the points for easier visualization. Click the button again to stop updating the location history. This sample uses a simulated data source to allow the sample to be useful on devices without a location data source.
```
Rachael-E: I think we need to specify/reword this slightly to make clear what happens when the app opens: when it opened I assumed the moving point _was_ tracking location (a moving point on a map) before realising that tracking makes an actual path. 

I've suggested removing the reference to "allow the sample to be useful on devices without a location data source" since that instruction is repeated in the "About the data" section.

```suggestion
The sample loads with a moving simulated location data source. Click the button to start tracking the location, which will appear as red points on the map. A green line will connect the points for easier visualization. Click the button again to stop updating the location history.
```

map_view/show-location-history/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+## How it works
+
+1. Create a `GraphicsOverlay` to show each point and another `GraphicsOverlay` for displaying the route line.
+2. Create a `SimulatedLocationDataSource` and initialize it with a polyline. Start the `SimulatedLocationDataSource` to begin receiving location updates.
```
Rachael-E:
```suggestion
2. Create a `SimulatedLocationDataSource` and initialize it with a `Polyline`. Start the `SimulatedLocationDataSource` to begin receiving location updates.
```

map_view/show-location-history/README.md
```diff
@@ -0,0 +1,40 @@
...
+3. Use a `LocationChangedListener` on the `simulatedLocationDataSource` to get location updates.
+4. When the location updates store that location, display a point on the map at the location, and re-create the route polyline.
+
+## Relevant API
+
+* LocationDataSource
+* LocationDataSource.Location
+* LocationDataSource.LocationChangedEvent
+* LocationDataSource.LocationChangedListener
+* LocationDataSource.Position
```
Rachael-E: This has always been a slightly contentious deviation by Java from the rest of the samples team to not include inner classes/methods in the "Relevant API" section. The idea being that `LocationDataSource` in itself is clear enough for this section, and that the inner workings of it are described within the "How it Works" section and code well. So from that perspective, I'd be in favour of removing these and keeping only `LocationDataSource`.  It's not a biggy if you want to keep them in though.

map_view/show-location-history/README.md
```diff
@@ -0,0 +1,40 @@
...
+* LocationDisplay.AutoPanMode
+* MapView.LocationDisplay
```
Rachael-E, line 26: e.g. this I am in favour of keeping because it specifically links the MapView to a LocationDisplay: if we just had `MapView` it wouldn't be clear its link to the sample other than the usual "display a map on the map view".

map_view/show-location-history/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+## Use case
+
+You can track device location history and display it as lines and points on the map. The history can be used to visualize how the user moved through the world, to retrace their steps, or to create new feature geometry. For example, an unmapped trail could be added to the map using this technique.
+
+## How to use the sample
+
+Click the button to start tracking your location, which will appear as points on the map. A line will connect the points for easier visualization. Click the button again to stop updating the location history. This sample uses a simulated data source to allow the sample to be useful on devices without a location data source.
+
+## How it works
```
Rachael-E, line 14: Kinda feel we might be a bit thin on the ground here instruction wise. E.g. we don't have a mention of getting the `LocationDisplay` from the `MapView`. Is there another sample in the pipeline that will be going over the basics of location displays? If so, we don't need to be too explicit about that here but I think it's worth dropping in mention of `locationDisplay` here since it's referred to in the Relevant API section. 

A good way to think about it: If it's not relevant enough to be in the "How it works", it's not relevant enough to go in the Relevant API section.

Rachael-E: Also the autopan is an interesting feature of the sample, and it's referenced in the Relevant API: some reference to that instruction here would be good to do also.

map_view/show-location-history/README.md
Rachael-E: Also: I guess we're not really using `LocationDataSource` in this sample, but `SimulatedLocationDataSource`? The only place we use `LocationDataSource` is to create a listener, and there's no reference to the class in "How it works": only in the "About the Data" section which is what got me thinking.

In which case do we only want to include `LocationDataSource.LocationChangedEvent` and `LocationDataSource.LocationChangedListener`

map_view/show-location-history/README.md
```diff
@@ -0,0 +1,40 @@
...
+* LocationDataSource.LocationChangedListener
+* LocationDataSource.Position
+* LocationDisplay.AutoPanMode
+* MapView.LocationDisplay
+* SimulatedLocationDataSource
+* SimulationParameters
+
+## About the data
+
+A custom set of points is used to create a `Polyline` and initialize a `SimulatedLocationDataSource`. The simulated location data source enables easier testing and allows the sample to be used on devices without an actively updating GPS signal. To track a user's real position, use the `LocationDataSource` instead.
```
Rachael-E: Is there a sample coming through for `LocationDataSource`? If so, we should come back to this readme when it's completed and hyperlink to it here. Or alternatively I might like to see some API Ref when it's available to hyperlink this class to for users wanting to easily access more info (not available until 100.10.0 release)
```suggestion
A custom set of points (provided in JSON format) is used to create a `Polyline` and initialize a `SimulatedLocationDataSource`. The simulated location data source enables easier testing and allows the sample to be used on devices without an actively updating GPS signal. To track a user's real position, use `LocationDataSource` instead. 
```

map_view/show-location-history/README.metadata.json
```diff
@@ -0,0 +1,45 @@
+{
+    "category": "Map view",
+    "description": "Display your location history on the map.",
+    "ignore": false,
+    "images": [
+        "ShowLocationHistory.png"
```
Rachael-E: I think a gif would look cool here! Both for the website and for the sample viewer. If you create a gif, then ensure the reference both here and in the readme are to the gif. The .png doesn't have to be referenced, it can just sit in the sample directory and the sample viewer will pick it up automatically.

map_view/show-location-history/src/main/java/com/esri/samples/show_location_history/ShowLocationHistorySample.java
```diff
@@ -0,0 +1,218 @@
...
+import com.esri.arcgisruntime.symbology.SimpleMarkerSymbol;
+import com.esri.arcgisruntime.symbology.SimpleRenderer;
+
+import org.apache.commons.io.IOUtils;
+
+
+public class ShowLocationHistorySample extends Application {
+
+  private MapView mapView;
+  private boolean isTrackingEnabled = false;
```
Rachael-E: I think we can handle this better than setting up a boolean to track things. I'm proposing to use a `ToggleButton`. See what you think, and we can have a chat about it too.

map_view/show-location-history/src/main/java/com/esri/samples/show_location_history/ShowLocationHistorySample.java
```diff
@@ -0,0 +1,218 @@
...
+      // create a map view and set its map
+      mapView = new MapView();
+      mapView.setMap(map);
+
+      // set the map views's viewpoint centered on Los Angeles, California and scaled
+      mapView.setViewpoint(new Viewpoint(new Point(-13185535.98, 4037766.28,
+        SpatialReferences.getWebMercator()), 7000));
+
+      // create a button that toggles the location tracking
+      Button trackingButton = new Button("Start tracking");
```
Rachael-E: if we make this a `ToggleButton` we can use the `isSelected()` property to keep a track on the tracking status without needing to add an extra boolean variable.

map_view/show-location-history/src/main/java/com/esri/samples/show_location_history/ShowLocationHistorySample.java
```diff
@@ -0,0 +1,218 @@
...
+          // toggle the location tracking when the button is clicked
+          if (isTrackingEnabled) {
+            trackingButton.setText("Start Tracking");
+            simulatedLocationDataSource.removeLocationChangedListener(locationChangedListener);
+          } else {
+            trackingButton.setText("Stop Tracking");
+            simulatedLocationDataSource.addLocationChangedListener(locationChangedListener);
+          }
+          isTrackingEnabled = !isTrackingEnabled;
+        });
```
Rachael-E: I'm thinking something like the following....but let's have a chat about it as it has its positives and  possible negatives:

```suggestion
          if (trackingButton.isSelected()) {
            trackingButton.setStyle("-fx-background-color: #00ff00; ");
            trackingButton.setText("Tracking Status: ACTIVE");
            simulatedLocationDataSource.addLocationChangedListener(locationChangedListener);
          } else {
            trackingButton.setStyle("-fx-background-color: #F16345; ");
            trackingButton.setText("Tracking Status: STOPPED");
            simulatedLocationDataSource.removeLocationChangedListener(locationChangedListener);
          }
        });
```

map_view/show-location-history/src/main/java/com/esri/samples/show_location_history/ShowLocationHistorySample.java
```diff
@@ -0,0 +1,218 @@
...
+
+            // add the updated polyline to the graphics overlay
+            locationHistoryLineOverlay.getGraphics().add(new Graphic(polylineBuilder.toGeometry()));
+          };
+
+        trackingButton.setOnAction(event -> {
+          // if the user has panned away from the location display, turn it on again
+          if (locationDisplay.getAutoPanMode() == LocationDisplay.AutoPanMode.OFF) {
+            locationDisplay.setAutoPanMode(LocationDisplay.AutoPanMode.RECENTER);
+          }
```
Rachael-E: the only pitfall to this is that of course, if you click and drag around the mapview you could lose the simulated data point unless you click the button again. If you click the button and don't touch the mapview, then the simulated data point is kept nicely within the mapview though. 

Is this as designed? What it be worth considering setting it up so that even if you interact with the mapview it always comes back to keep the point in the mapview?

map_view/show-location-history/src/main/java/com/esri/samples/show_location_history/ShowLocationHistorySample.java
```diff
@@ -0,0 +1,218 @@
...
+      // create a polyline builder to connect the location points
+      PolylineBuilder polylineBuilder = new PolylineBuilder(SpatialReferences.getWebMercator());
+
+      try {
+        // access the json of the location points
+        String polylineData = IOUtils.toString(getClass().getResourceAsStream("/show_location_history/polyline_data.json"),
+          StandardCharsets.UTF_8);
+        // create a polyline from the location points
+        Polyline routePolyline = (Polyline) Geometry.fromJson(polylineData,
+          SpatialReferences.getWebMercator());
```
Rachael-E: this can all go on the one line without exceeding the recommended line limit 
```suggestion
        Polyline routePolyline = (Polyline) Geometry.fromJson(polylineData, SpatialReferences.getWebMercator());
```

map_view/show-location-history/src/main/java/com/esri/samples/show_location_history/ShowLocationHistorySample.java
```diff
@@ -0,0 +1,218 @@
...
+              polylineBuilder.addPoint(position);
+              // add the new point to the graphics overlay
+              locationHistoryOverlay.getGraphics().add(new Graphic(position));
+            }
+            // store the current position
+            position = locationChangedEvent.getLocation().getPosition();
+
+            // add the updated polyline to the graphics overlay
+            locationHistoryLineOverlay.getGraphics().add(new Graphic(polylineBuilder.toGeometry()));
+          };
```
Rachael-E: the indentation is off by two on this block (IntelliJ has a shortcut to auto-indent if you haven't seen that already!)

map_view/show-location-history/src/main/java/com/esri/samples/show_location_history/ShowLocationHistorySample.java
```diff
@@ -0,0 +1,218 @@
...
+            new Alert(Alert.AlertType.ERROR, "Map failed to load").show();
+          }
+        });
+
+        // add the map view and tracking button to the stack pane
+        stackPane.getChildren().addAll(mapView, trackingButton);
+        StackPane.setAlignment(trackingButton, Pos.TOP_LEFT);
+        StackPane.setMargin(trackingButton, new Insets(10, 0, 0, 10));
+
+      } catch (IOException e) {
```
Rachael-E: something isn't quite right with this `try` `catch` block. This set up infers if _anything_ within the block goes wrong, it's because there was an `Error` "loading the simulated data" which wouldn't be accurate. However, if anything does go wrong, it will bubble up to the try/catch initiated at the start of the `start()` method. 

Therefore we can remove this try/catch since anything going wrong with Json would be handled by the `start()` method try/catch. If we wanted to be extra safe for this data, we could put in an if condition to e.g. ensure the string isn't empty/the file exists and if it is return an Alert: but the data should be there since in this instance it is bundled with the sample (e.g. other sample data is not). Looking at our other samples bringing in resources as stream, looks like we don't put that extra check in (doesn't mean we shouldn't going forward, just an observation).

map_view/show-location-history/src/main/java/com/esri/samples/show_location_history/ShowLocationHistorySample.java
```diff
@@ -0,0 +1,218 @@
...
+import com.esri.arcgisruntime.mapping.view.Graphic;
+import com.esri.arcgisruntime.mapping.view.GraphicsOverlay;
+import com.esri.arcgisruntime.mapping.view.LocationDisplay;
+import com.esri.arcgisruntime.mapping.view.MapView;
+import com.esri.arcgisruntime.symbology.SimpleLineSymbol;
+import com.esri.arcgisruntime.symbology.SimpleMarkerSymbol;
+import com.esri.arcgisruntime.symbology.SimpleRenderer;
+
+import org.apache.commons.io.IOUtils;
+
```
Rachael-E, line 61: remove excess line here

Rachael-E (changes requested):
@sclaridge nice work! Interesting wee sample. Nothing major here, some comments fine tuning the readme to be more specific to how the sample is working, question over the AutoPan functionality and an alternative UI suggestion for us to look over and discuss.

map_view/show-location-history/src/main/java/com/esri/samples/show_location_history/ShowLocationHistorySample.java
```diff
@@ -0,0 +1,218 @@
...
+        map.addDoneLoadingListener(() -> {
+
+          if (map.getLoadStatus() == LoadStatus.LOADED) {
+            trackingButton.setDisable(false);
+
+            // start the simulated location data source
+            simulatedLocationDataSource.startAsync();
+
+          } else {
+            new Alert(Alert.AlertType.ERROR, "Map failed to load").show();
```
Rachael-E: we can put in a direct reference to why the map has failed to load with:

```suggestion
            new Alert(Alert.AlertType.ERROR, "Map failed to load" +       map.getLoadError().getCause().getMessage()).show();
```

map_view/show-location-history/README.md
sclaridge: I'm not mentioning `SimulationParameters` in the "How it works" section. Would it be ok to also remove it from the Relevant API section?

sclaridge:
@Rachael-E I've added the highlighted toggle button and accompanying gif!

map_view/show-location-history/README.md
```diff
@@ -2,38 +2,34 @@
...
 ## How to use the sample
 
-Click the button to start tracking your location, which will appear as points on the map. A line will connect the points for easier visualization. Click the button again to stop updating the location history. This sample uses a simulated data source to allow the sample to be useful on devices without a location data source.
-
+The sample loads with a moving simulated location data source. Click the button to start tracking the location, which will appear as red points on the map. A green line will connect the points for easier visualization. Click the button again to stop updating the location history.
 ## How it works
 
 1. Create a `GraphicsOverlay` to show each point and another `GraphicsOverlay` for displaying the route line.
-2. Create a `SimulatedLocationDataSource` and initialize it with a polyline. Start the `SimulatedLocationDataSource` to begin receiving location updates.
+2. Create a `SimulatedLocationDataSource` and initialize it with a `Polyline`. Start the `SimulatedLocationDataSource` to begin receiving location updates.
```
Rachael-E: let's add in `SimulationParameters` here as you were asking about: we can also be a bit clearer about how the `Polyline` and parameters are being added, since technically, we aren't initializing the `SimulatedLocationDataSource` with them: we are creating a new one and then setting its location. 
```suggestion
2. Create a `SimulatedLocationDataSource` and call its `setLocations()` method, passing the route `Polyline` and new `SimulatedParameters` as parameters. Start the `SimulatedLocationDataSource` to begin receiving location updates.
```

map_view/show-location-history/README.md
```diff
@@ -2,38 +2,34 @@
...
-* LocationDataSource.Position
 * LocationDisplay.AutoPanMode
 * MapView.LocationDisplay
 * SimulatedLocationDataSource
 * SimulationParameters
 
 ## About the data
 
-A custom set of points is used to create a `Polyline` and initialize a `SimulatedLocationDataSource`. The simulated location data source enables easier testing and allows the sample to be used on devices without an actively updating GPS signal. To track a user's real position, use the `LocationDataSource` instead. 
+A custom set of points (provided in JSON format) is used to create a `Polyline` and initialize a `SimulatedLocationDataSource`. The simulated location data source enables easier testing and allows the sample to be used on devices without an actively updating GPS signal. To track a user's real position, use `NMEALocationDataSource` instead.
```
Rachael-E: since it's not strictly used in an initializer, slight word tweak here: let me know if you think it doesn't make sense, we can try to find a better word! 
```suggestion
A custom set of points (provided in JSON format) is used to create a `Polyline` and configure a `SimulatedLocationDataSource`. The simulated location data source enables easier testing and allows the sample to be used on devices without an actively updating GPS signal. To track a user's real position, use `NMEALocationDataSource` instead. 
```

map_view/show-location-history/src/main/resources/show_location_history/style.css
```diff
@@ -0,0 +1,27 @@
...
+   -fx-text-fill: white;
+}
+
+.panel-region .radio-button {
+   -fx-text-fill: white;
+}
+
+.color-picker .color-picker-label {
+   -fx-text-fill: black;
+}
```
Rachael-E: we can remove these since we have none of the above in this sample :) 
```suggestion
```

map_view/show-location-history/README.metadata.json
```diff
@@ -3,7 +3,7 @@
     "description": "Display your location history on the map.",
     "ignore": false,
     "images": [
-        "ShowLocationHistory.png"
+        "ShowLocationHistory.gif"
     ],
     "keywords": [
         "GPS",
```
Rachael-E, line 9: we should remove the following classes from keywords and relevant API section here since we have removed them from the readme: 

* LocationDataSource	
* LocationDataSource.Location
* LocationDataSource.Position

map_view/show-location-history/src/main/java/com/esri/samples/show_location_history/ShowLocationHistorySample.java
```diff
@@ -111,84 +120,88 @@ public void start(Stage stage) {
...
+        } else {
+          new Alert(Alert.AlertType.ERROR, "Map failed to load: " + map.getLoadError().getCause().getMessage()).show();
+        }
+      });
+
+      // create a control panel and add the label and button UI components
+      HBox controlsHBox = new HBox(6);
+      controlsHBox.setBackground(new Background(new BackgroundFill(Paint.valueOf("rgba(0,0,0,0.3)"), CornerRadii.EMPTY, Insets.EMPTY)));
+      controlsHBox.setPadding(new Insets(10.0));
+      controlsHBox.setMaxSize(180, 50);
```
Rachael-E, line 197: let's add this in just to make it super neat! 
 ```java
controlsHBox.setAlignment(Pos.CENTER);
```

Rachael-E (approved):
thanks @sclaridge, this looks great! A few minor things that need some attention, but I'll approve so that you can go ahead and merge this once they have been addressed.

PR #558, opened 2020-09-21
Fix cancel button in Edit features with feature-linked annotation

sclaridge:
This PR fixes a bug noticed by @hardikc07 during verification. Previously, when the dialog box 's cancel button was clicked, the feature was still selected.

Now when when the update button is clicked, `editAttributesDialog.showAndWait()` returns true, otherwise it returns null, which is stored in an Optional. When the dialog is created, the features are updated based on the value in the Optional.

### DIFF
--- a/editing/edit-features-with-feature-linked-annotation/src/main/java/com.esri.samples.edit_features_with_feature_linked_annotation/EditAttributesDialog.java
+++ b/editing/edit-features-with-feature-linked-annotation/src/main/java/com.esri.samples.edit_features_with_feature_linked_annotation/EditAttributesDialog.java
@@ -28,7 +28,7 @@ import com.esri.arcgisruntime.data.Feature;
 /**
  * Custom dialog for editing feature attributes.
  */
-public class EditAttributesDialog extends Dialog<Feature>{
+public class EditAttributesDialog extends Dialog<Boolean>{
 
   @FXML private TextField addressTextField;
   @FXML private TextField streetNameTextField;
@@ -51,7 +51,7 @@ public class EditAttributesDialog extends Dialog<Feature>{
     addressTextField.setText(selectedFeature.getAttributes().get("AD_ADDRESS").toString());
     streetNameTextField.setText(selectedFeature.getAttributes().get("ST_STR_NAM").toString());
 
-    // convert the result to an address and street name when the ok button is clicked
+    // convert the result to an address and street name when the update button is clicked
     setResultConverter(dialogButton -> {
       if (dialogButton == updateButton) {
         try {
@@ -67,6 +67,7 @@ public class EditAttributesDialog extends Dialog<Feature>{
         } catch (Exception e) {
           e.printStackTrace();
         }
+        return true;
       }
       return null;
     });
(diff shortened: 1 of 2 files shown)


editing/edit-features-with-feature-linked-annotation/src/main/java/com.esri.samples.edit_features_with_feature_linked_annotation/EditAttributesDialog.java
```diff
@@ -67,6 +67,7 @@
         } catch (Exception e) {
           e.printStackTrace();
         }
+        return true;
       }
       return null;
```
JonLavi, line 72: Perhaps it's possible to return `false` here. That way we can convert the type `Optional<Boolean>` to just `Boolean`.

editing/edit-features-with-feature-linked-annotation/src/main/java/com.esri.samples.edit_features_with_feature_linked_annotation/EditFeaturesWithFeatureLinkedAnnotationSample.java
```diff
@@ -225,10 +226,17 @@ else if (selectedFeature.getGeometry().getGeometryType() == GeometryType.POINT)
       EditAttributesDialog editAttributesDialog = new EditAttributesDialog(selectedFeature);
 
       // show the dialog and wait for the user response
-      editAttributesDialog.showAndWait();
+      Optional<Boolean> updatingFeature = editAttributesDialog.showAndWait();
 
-      // update the selected feature's feature table
-      updateAttributes(selectedFeature);
+      // if the user chose to update the feature's attributes
+      if (updatingFeature.isPresent()) {
```
JonLavi, line 232: If the suggestion of returning just a `Boolean` from the dialog works, we can write here:
```suggestion
      if (updatingFeature) {
```

JonLavi (approved):
Very nice fix @sclaridge 
I've got one suggestion, which would be returning a `Boolean` instead of an `Optional<Boolean>` from the dialog. It could simplify things a tad more. If that doesn't work at all, however, what you have here works well too, so I'll approve so that you can get a second review asap!

editing/edit-features-with-feature-linked-annotation/src/main/java/com.esri.samples.edit_features_with_feature_linked_annotation/EditAttributesDialog.java
sclaridge: `Dialog.showAndWait()` returns an Optional, so I don't think I can use a `Boolean`.

JonLavi: Ah no worries then!

editing/edit-features-with-feature-linked-annotation/src/main/java/com.esri.samples.edit_features_with_feature_linked_annotation/EditFeaturesWithFeatureLinkedAnnotationSample.java
```diff
@@ -225,10 +226,17 @@ else if (selectedFeature.getGeometry().getGeometryType() == GeometryType.POINT)
...
+        // update the selected feature's feature table
+        updateAttributes(selectedFeature);
+      }
+      else{
```
Rachael-E: extra space needed between else and bracket 
```suggestion
      else {
```

Rachael-E (approved):
thanks @sclaridge this change looks good! Just a minor syntax thing to fix up (missing an extra space between else and a bracket) then good for you to go ahead and merge.

PR #562, opened 2020-10-02
Manage Operational Layers: UI not working as described

jenmerritt:
### Description

I've done some work on the Manage Operational Layers sample. In the last couple of releases regression testing of going through all the samples it was picked up that the functionality was not working correctly in this sample when adding/removing layers and altering their order. The issue seemed to be how the list of layers and actual layers on the map were synced. The index is being used and for the list of layers this list reads top to bottom (index 0, 1, 2) whereas for the layers on the map, index 0 is actually the bottom layer.

This sample now works as per the readme. I did a few refactors along with the fixes. There were a few errors not being caught so I've also added in a few if statements to allow for these. 

Another thing I noticed is that there isn't any allowance for the image layers loading (or not). However when I started to look into this it got quite complex with there being 3 separate image layers. So if you break the URL of the any of the layers, it doesn't cause an exception or any errors that I can use to control the list of names, instead the image layer is still created and added to the map, it is just empty, and so the label for that layer in the list of names is also empty. The sample doesn't 'break' per se, it just looks weird with a blank name. I'd be interested in any assistance with this and happy to demo it.

Ref: CS issue 1961

Checklist:
- [x] Request a reviewer
- [x] Assign a reviewer
- [x] Tag reviewer in comment

### DIFF
--- a/map/manage-operational-layers/src/main/java/com/esri/samples/manage_operational_layers/ManageOperationalLayersSample.java
+++ b/map/manage-operational-layers/src/main/java/com/esri/samples/manage_operational_layers/ManageOperationalLayersSample.java
@@ -1,12 +1,12 @@
 /*
  * Copyright 2017 Esri.
- * 
+ *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not
  * use this file except in compliance with the License. You may obtain a copy of
  * the License at
- * 
+ *
  * http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
  * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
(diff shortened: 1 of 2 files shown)


map/manage-operational-layers/src/main/java/com/esri/samples/manage_operational_layers/ManageOperationalLayersSample.java
```diff
@@ -38,131 +38,156 @@
...
+    "https://sampleserver5.arcgisonline.com/arcgis/rest/services/DamageAssessment/MapServer";
+  private static final String ELEVATION_LAYER =
+    "https://sampleserver5.arcgisonline.com/arcgis/rest/services/Elevation/WorldElevations/MapServer";
 
   @Override
   public void start(Stage stage) {
 
     try {
-      // create stack pane and application scene
+      // create the stack pane and the application scene
```
JonLavi, line 55: I like the change to plain english 👍 
Might be worth creating an issue for a 'find-and-replace' effort to make all these uniform?

map/manage-operational-layers/src/main/java/com/esri/samples/manage_operational_layers/ManageOperationalLayersSample.java
```diff
@@ -38,131 +38,156 @@
...
-      controlsVBox.getStyleClass().add("panel-region");
+      // create a ArcGISMap with the topographic basemap
+      ArcGISMap map = new ArcGISMap(Basemap.createTopographic());
 
-      // create labels for add/delete layers
-      Label addLayersLabel = new Label("Layers on the ArcGISMap");
-      Label deleteLayersLabel = new Label("Deleted Layers");
-      addLayersLabel.getStyleClass().add("panel-label");
-      deleteLayersLabel.getStyleClass().add("panel-label");
+      // create a view and added ArcGISMap to it
```
JonLavi:
```suggestion
      // create a view and add the ArcGISMap to it
```

map/manage-operational-layers/src/main/java/com/esri/samples/manage_operational_layers/ManageOperationalLayersSample.java
```diff
@@ -38,131 +38,156 @@
...
-  private static final String ELEVATION_LAYER =
-      "https://sampleserver5.arcgisonline.com/arcgis/rest/services/Elevation/WorldElevations/MapServer";
   private static final String CENSUS_LAYER =
-      "https://sampleserver5.arcgisonline.com/arcgis/rest/services/Census/MapServer";
+    "https://sampleserver5.arcgisonline.com/arcgis/rest/services/Census/MapServer";
   private static final String DAMAGE_LAYER =
-      "https://sampleserver5.arcgisonline.com/arcgis/rest/services/DamageAssessment/MapServer";
+    "https://sampleserver5.arcgisonline.com/arcgis/rest/services/DamageAssessment/MapServer";
+  private static final String ELEVATION_LAYER =
+    "https://sampleserver5.arcgisonline.com/arcgis/rest/services/Elevation/WorldElevations/MapServer";
```
JonLavi: In other samples where strings for services are only used once, we write these URLs directly in the constructor, and don't create a String variable for them. Just an observation regarding consistency. I don't mind either way

map/manage-operational-layers/src/main/java/com/esri/samples/manage_operational_layers/ManageOperationalLayersSample.java
```diff
@@ -38,131 +38,156 @@
...
+      // create the elevation, census, and damage image layers
+      final ArcGISMapImageLayer censusImageLayer = new ArcGISMapImageLayer(CENSUS_LAYER);
+      final ArcGISMapImageLayer damageImageLayer = new ArcGISMapImageLayer(DAMAGE_LAYER);
+      final ArcGISMapImageLayer elevationImageLayer = new ArcGISMapImageLayer(ELEVATION_LAYER);
+
+      // add the image layers to the list of operational layers on the map
+      mapLayers = map.getOperationalLayers();
+      mapLayers.add(elevationImageLayer);
+      mapLayers.add(damageImageLayer);
+      mapLayers.add(censusImageLayer);
```
JonLavi: If it seems useful, you could create a method to refactor this a little (althoug I know methods aren't encouraged in samples):
```
private void addMapLayer(String serviceUrl){
	ArcGISMapImageLayer imageLayer =  new ArcGISMapImageLayer(serviceUrl);
	mapLayers.add(imaageLayer);
}

File: map/manage-operational-layers/src/main/java/com/esri/samples/manage_operational_layers/ManageOperationalLayersSample.java [hunk 5]
```diff
@@ -38,131 +38,156 @@
...
+
+          // get the index of the selected item
+          int selectedIndex = addedLayerNamesList.getSelectionModel().getSelectedIndex();
 
-      deletedLayerNames.setOnMouseClicked(e -> {
-        // index of selected item
-        int selectedIndex = deletedLayerNames.getSelectionModel().getSelectedIndex();
-        if (e.getButton() == MouseButton.SECONDARY) {
+          // if a valid index is selected
+          if (selectedIndex >= 0) {
```
JonLavi, line 119: I'm wondering if the selected index can ever be `selectedIndex < 0`? 
If that's not a technically viable option, would it be possible to write `selectedIndex != null`? Just wondering if there's any implication in one expression or the other

map/manage-operational-layers/src/main/java/com/esri/samples/manage_operational_layers/ManageOperationalLayersSample.java
```diff
@@ -38,131 +38,156 @@
...
+              // if the layer is not already at the top, move it to the top of the list and layers on the map
+              if (selectedIndex != 0) {
+
+                // move the layer's name to the top of the added layers list. Note this is adding to index 0.
+                addedLayerNamesList.getItems().remove(selectedIndex);
+                addedLayerNamesList.getItems().add(0, selectedLayer.getName());
+
+                // move the layer to the top of the layers on the map. Note this is adding to the last index.
+                mapLayers.remove(indexOfSelectedLayer);
+                mapLayers.add(mapLayers.size(), selectedLayer);
```
JonLavi, line 136: Back when I noticed that this sample doesn't quite work as expected, I figured this would be the way to go. Unfortunately I still find it a bit clunky to have to keep two lists in sync manually. Tyler had pointed out that `map.getOperationalLayers()` should ideally be an `ObservableList`, which would solve this problem quite elegantly. @mbcoder not sure whether this is easy to pull off in hindsight, but maybe it could be added to the backlog.

map/manage-operational-layers/src/main/java/com/esri/samples/manage_operational_layers/ManageOperationalLayersSample.java
```diff
@@ -38,131 +38,156 @@
...
+          }
+        }
+      });
+
+      // create a control panel and add the label and list UI components
+      VBox controlsVBox = new VBox(6);
+      controlsVBox.setBackground(new Background(new BackgroundFill(Paint.valueOf("rgba(0,0,0,0.3)"), CornerRadii.EMPTY,
+        Insets.EMPTY)));
+      controlsVBox.setPadding(new Insets(10.0));
+      controlsVBox.setMaxSize(200, 260);
```
JonLavi: These lines are now repeated in `style.css`. I think what we have in the CSS generally works ok with the SV, so we should be able to delete these?

map/manage-operational-layers/src/main/java/com/esri/samples/manage_operational_layers/ManageOperationalLayersSample.java
```diff
@@ -38,131 +38,156 @@
...
+  private LayerList mapLayers;
 
-  private static final String ELEVATION_LAYER =
-      "https://sampleserver5.arcgisonline.com/arcgis/rest/services/Elevation/WorldElevations/MapServer";
   private static final String CENSUS_LAYER =
-      "https://sampleserver5.arcgisonline.com/arcgis/rest/services/Census/MapServer";
+    "https://sampleserver5.arcgisonline.com/arcgis/rest/services/Census/MapServer";
   private static final String DAMAGE_LAYER =
-      "https://sampleserver5.arcgisonline.com/arcgis/rest/services/DamageAssessment/MapServer";
+    "https://sampleserver5.arcgisonline.com/arcgis/rest/services/DamageAssessment/MapServer";
```
JonLavi: When testing the sample, I can't see this layer having any content. I had a look at the service, and it doesn't look like there are actually any interesting features in it:

Snag_2bd09d

https://www.arcgis.com/home/webmap/viewer.html?featurecollection=https%3A%2F%2Fsampleserver5.arcgisonline.com%2Farcgis%2Frest%2Fservices%2FDamageAssessment%2FMapServer%3Ff%3Djson%26option%3Dfootprints&supportsProjection=true&supportsJSONP=true

Might be worth mentioning to the team and seeing if there's more interesting data to put in there, so it's visually clearer. Unless there was a thought behind there being no data in one of the layers?

JonLavi (approved):
Nice work on this fix @jenmerritt !
No major issues here, just a couple of observations and comments. I'll leave you to pick and choose which to implement. Approving so that you can move on to get a second review.

map/manage-operational-layers/src/main/java/com/esri/samples/manage_operational_layers/ManageOperationalLayersSample.java
jenmerritt: Yeah there's heaps of these! Will do that 👍

map/manage-operational-layers/src/main/java/com/esri/samples/manage_operational_layers/ManageOperationalLayersSample.java
jenmerritt: I originally implemented with `!= null` but then if you click in any of the blank spots in the list it returns an index of -1 which was causing an out of range exception, so I'm avoiding that with this if statement.

map/manage-operational-layers/src/main/java/com/esri/samples/manage_operational_layers/ManageOperationalLayersSample.java
jenmerritt: I added to CSS and forgot to remove from the sample - good spot!

map/manage-operational-layers/src/main/java/com/esri/samples/manage_operational_layers/ManageOperationalLayersSample.java
jenmerritt: image

@JonLavi I might be misunderstanding but the damage assessment has some data when you zoom out, so gives another layer to play with with the controls?

map/manage-operational-layers/src/main/java/com/esri/samples/manage_operational_layers/ManageOperationalLayersSample.java
jenmerritt: Would make it more easily extendable to add more layers, especially now it's working for `x` number of layers that you put in. I'll refactor.

map/manage-operational-layers/src/main/java/com/esri/samples/manage_operational_layers/ManageOperationalLayersSample.java
jenmerritt: Refactored as per method below.

jenmerritt:
@Rachael-E ready for review. Be interested in your thoughts on my original comment RE: loading of the layers. Thanks!

map/manage-operational-layers/src/main/java/com/esri/samples/manage_operational_layers/ManageOperationalLayersSample.java
Rachael-E: I would be in favour of having a more zoomed out starting point for this sample: it's not instantly clear you have to zoom out to see the damage assessment layer and like Jon says, it doesn't look too interesting. 

If the map was scaled to `10000000` in the constructor, you could see a few points: however, these points will move around over time so I'd be more in favour of centralising the viewpoint mid-North America (pretty much so it loads looking like the screenshot you have above: though if you do that, you might want to disable the UI until the world elevation layer has finished loading since it might be a little slow).

Rachael-E: or we could manually add lots of points in around Redlands ;)

JonLavi: My bad, I didn't realise you need to zoom out that far... I don't mind either way, but probably sensible to have some data there that shows at the initial viewpoint that we set in the code

map/manage-operational-layers/src/main/java/com/esri/samples/manage_operational_layers/ManageOperationalLayersSample.java
JonLavi: You could also add a check for the loading here (add a `doneLoadingListener` and have it show a message).
There's an argument to be made for leaving all that out (since it's not the core of what the sample tries to do), or putting it in because it offers a better user experience! I'll leave that up to you and @Rachael-E to decide 😉

map/manage-operational-layers/src/main/java/com/esri/samples/manage_operational_layers/ManageOperationalLayersSample.java
jenmerritt: I've updated the initial viewpoint, think it works way better. As the user you're keen to use the functionality and get that damage layer to the top. Thanks both for the discussion! @Rachael-E @JonLavi 

image

map/manage-operational-layers/src/main/java/com/esri/samples/manage_operational_layers/ManageOperationalLayersSample.java
```diff
@@ -38,131 +38,142 @@
 import com.esri.arcgisruntime.mapping.ArcGISMap;
 import com.esri.arcgisruntime.mapping.Basemap;
 import com.esri.arcgisruntime.mapping.LayerList;
+import com.esri.arcgisruntime.mapping.Viewpoint;
 import com.esri.arcgisruntime.mapping.view.MapView;
 
 public class ManageOperationalLayersSample extends Application {
 
+  private LayerList mapLayers;
```
Rachael-E: let's make this a local variable

map/manage-operational-layers/src/main/java/com/esri/samples/manage_operational_layers/ManageOperationalLayersSample.java
```diff
@@ -38,131 +38,142 @@
...
+
+      // set the initial viewpoint for the map view
+      mapView.setViewpoint(new Viewpoint(37.056295, -111.195800, 10000000));
 
-      // create list for deleted layers
+      // add the image layers to the list of operational layers on the map
+      mapLayers = map.getOperationalLayers();
+      addMapImageLayer("https://sampleserver5.arcgisonline.com/arcgis/rest/services/Elevation/WorldElevations/MapServer");
+      addMapImageLayer("https://sampleserver5.arcgisonline.com/arcgis/rest/services/DamageAssessment/MapServer");
+      addMapImageLayer("https://sampleserver5.arcgisonline.com/arcgis/rest/services/Census/MapServer");
```
Rachael-E: in reference to Jon's comment here, it's true that it's not really standard for Java samples to contain abstracted methods, such as here, for what is a pretty straightforward initialiser. Adding map image layers to the map's operational layers are a key part of this sample, so it's better to show the `ArcGISMapImageLayer`s instantiated right up front.

We can shorthand it a little with the following:
```suggestion
      // create and add new image layers to the map's operational layers
      LayerList mapLayers = map.getOperationalLayers();
      mapLayers.addAll(Arrays.asList(
        (new ArcGISMapImageLayer("https://sampleserver5.arcgisonline.com/arcgis/rest/services/Elevation/WorldElevations/MapServer")),
        (new ArcGISMapImageLayer("https://sampleserver5.arcgisonline.com/arcgis/rest/services/DamageAssessment/MapServer")),
        (new ArcGISMapImageLayer("https://sampleserver5.arcgisonline.com/arcgis/rest/services/Census/MapServer"))));
```

map/manage-operational-layers/src/main/java/com/esri/samples/manage_operational_layers/ManageOperationalLayersSample.java
```diff
@@ -38,131 +38,142 @@
...
+
```
Rachael-E, line 93: @jenmerritt, to follow up on your question about loading checks. After we instantiate and add the map image layers to the map's operational layers, we can check to ensure each layer has loaded correctly with the following: 

```suggestion
// check that the layer has loaded correctly and if not display an error message
      mapLayers.forEach(layer ->
        layer.addDoneLoadingListener(() -> {
          if (layer.getLoadStatus() != LoadStatus.LOADED) {
            new Alert(Alert.AlertType.ERROR, "Map image layer failed to load: " +
              layer.getLoadError().getCause().getMessage()).show();}
        })
      );
```

Rachael-E (commented):
@jenmerritt just minor things here, refactoring of the instantiation of the map image layers and a suggestion for handling the loading of the layers as you requested: let me know if you have any questions/thoughts on this pattern!

map/manage-operational-layers/src/main/java/com/esri/samples/manage_operational_layers/ManageOperationalLayersSample.java
jenmerritt: Thanks @Rachael-E , this is ideal and works well, I've tested breaking the layers. Was trying to think the other way round and if they did load, rather than this pattern of if they didn't.

map/manage-operational-layers/src/main/java/com/esri/samples/manage_operational_layers/ManageOperationalLayersSample.java
```diff
@@ -70,13 +72,24 @@ public void start(Stage stage) {
...
+      // create and add new image layers to the map's operational layers
+      LayerList mapLayers = map.getOperationalLayers();
+      mapLayers.addAll(Arrays.asList(
+        (new ArcGISMapImageLayer("https://sampleserver5.arcgisonline.com/arcgis/rest/services/Elevation/WorldElevations/MapServer")),
+        (new ArcGISMapImageLayer("https://sampleserver5.arcgisonline.com/arcgis/rest/services/DamageAssessment/MapServer")),
+        (new ArcGISMapImageLayer("https://sampleserver5.arcgisonline.com/arcgis/rest/services/Census/MapServer"))
+      ));
+
+      // check that the layer has loaded correctly and if not display an error message
```
Rachael-E:
```suggestion
      // check that each layer has loaded correctly and if not, display an error message
```

Rachael-E (approved):
looks great, thanks @jenmerritt for updating that! I should have referenced "each" layer rather than "the" layer in my original suggestion: I've put that in as a suggestion to fix up and then you're good to merge.

PR #563, opened 2020-10-05
Small fixes to samples from testing

jenmerritt:
### Description
The following (very) small changes have been made as picked up in a previous round of regression testing when the samples were all tested through:

1. Project Sample: removed a duplicate from the metadata file, was causing 2 duplicate code examples to display in SV / developer site.

2. Generate offline map with local basemap sample: added copyrights to all relevant files

3. Show callout sample: removed the animation 'duration' functionality from the callout. No other SDKs have this functionality so after discussing with the team it was agreed it was not needed as the sample is about displaying a callout where clicked. However this originally came up because on clicking around the screen rapidly with a duration set, the callout would gradually fade out and no longer be visible. Should this be raised as a Java bug anyway?

Ref: CS 2165

@JonLavi first, thanks :) 

Checklist:
- [x] Request a reviewer
- [x] Assign a reviewer
- [x] Tag reviewer in comment

### DIFF
--- a/display_information/show-callout/README.md
+++ b/display_information/show-callout/README.md
@@ -19,7 +19,7 @@ Click anywhere on the map. A callout showing the WGS84 coordinates for the click
 3. Get the screen point's location on the map using `MapView.screenToLocation(Point2D)`.
 4. Get the `MapView`'s callout.
 5. Set the title and detail of the callout to display the map point's coordinates.
-6. Display the callout at the map point and specify an animation duration by using `callout.showCalloutAt(mapPoint, duration)`.
+6. Display the callout at the map point `callout.showCalloutAt(mapPoint)`.
 
 ## Relevant API
 
--- a/geometry/project/README.metadata.json
+++ b/geometry/project/README.metadata.json
@@ -28,7 +28,6 @@
         "SpatialReference"
     ],
     "snippets": [
-        "ProjectSample.java",
         "src/main/java/com/esri/samples/project/ProjectSample.java"
     ],
     "title": "Project"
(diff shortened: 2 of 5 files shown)


display_information/show-callout/README.md
```diff
@@ -19,7 +19,7 @@ Click anywhere on the map. A callout showing the WGS84 coordinates for the click
 3. Get the screen point's location on the map using `MapView.screenToLocation(Point2D)`.
 4. Get the `MapView`'s callout.
 5. Set the title and detail of the callout to display the map point's coordinates.
-6. Display the callout at the map point and specify an animation duration by using `callout.showCalloutAt(mapPoint, duration)`.
+6. Display the callout at the map point `callout.showCalloutAt(mapPoint)`.
```
JonLavi, line 22: @jenmerritt I think we should definitely raise an issue for the callout animation. Not sure how to describe it, but it's almost as if the animation duration is 'stacking' if we initiate another animation before it's finished. Feel free to log an issue (ideally add a snippet of the sample code and a gif), or give me a shout if there's anything to clarify.

geometry/project/README.metadata.json
```diff
@@ -28,7 +28,6 @@
         "SpatialReference"
     ],
     "snippets": [
-        "ProjectSample.java",
```
JonLavi, line 31: Good catch on this!

JonLavi (approved):
Thanks for fixing these up @jenmerritt !
Let's log an issue for the callout duration, would be good to get that fixed up.

jenmerritt:
@Rachael-E this is a smaller one, ready for your review. Thanks!

Rachael-E (approved):
looks good @jenmerritt, just one thing which is to remember to remove the `Duration` import from the show-callout sample. Other than that you're good to merge!

PR #570, opened 2020-10-14
Manage Operational Layers - VBox styles

jenmerritt:
### Description 

Moving more CSS to the main class after verification. VBox width/height/padding styles were not carried across to the SV. I've made the changes and tested in the SV.

FYI @sclaridge for when we look at the SV more, this is a good example to explore and decide what we want to achieve to improve/avoid this in the future. Seems to be more of an issue when it isn't an fxml sample.

@JonLavi first review please, thanks!

Checklist:
- [x] Request a reviewer
- [x] Assign a reviewer
- [x] Tag reviewer in comment

### DIFF
--- a/map/manage-operational-layers/src/main/java/com/esri/samples/manage_operational_layers/ManageOperationalLayersSample.java
+++ b/map/manage-operational-layers/src/main/java/com/esri/samples/manage_operational_layers/ManageOperationalLayersSample.java
@@ -184,6 +184,9 @@ public class ManageOperationalLayersSample extends Application {
       controlsVBox.setBackground(new Background(new BackgroundFill(Paint.valueOf("rgba(0,0,0,0.3)"), CornerRadii.EMPTY,
         Insets.EMPTY)));
       controlsVBox.getStyleClass().add("panel-region");
+      controlsVBox.setPadding(new Insets(10.0));
+      controlsVBox.setMaxWidth(200);
+      controlsVBox.setMaxHeight(260);
       controlsVBox.getChildren().addAll(addedLayersLabel, addedLayerNamesList, deletedLayersLabel, deletedLayerNamesList);
 
       // add the map view and control panel to stack pane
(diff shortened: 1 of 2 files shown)


map/manage-operational-layers/src/main/java/com/esri/samples/manage_operational_layers/ManageOperationalLayersSample.java
```diff
@@ -184,6 +184,9 @@ public void start(Stage stage) {
       controlsVBox.setBackground(new Background(new BackgroundFill(Paint.valueOf("rgba(0,0,0,0.3)"), CornerRadii.EMPTY,
         Insets.EMPTY)));
       controlsVBox.getStyleClass().add("panel-region");
+      controlsVBox.setPadding(new Insets(10.0));
+      controlsVBox.setMaxWidth(200);
+      controlsVBox.setMaxHeight(260);
```
JonLavi, line 189: Should we not remove these relevant lines from the `styles.css` file?
https://github.com/Esri/arcgis-runtime-samples-java/blob/42751302b377153ee1d5631b5374410eaf8d7822/map/manage-operational-layers/src/main/resources/manage_operational_layers/style.css#L1-L4

I'm also noticing the definition in the css is purely `.panel-region {` 
Have you tried `.panel-region .vbox {` by any chance? I don't have much faith in that suddenly working, buy knowing CSS it can sometimes fix things just by being more specific about which element to target with the rules.

JonLavi (commented):
Thanks for looking into these CSS issues @jenmerritt , just a suggestion on something to try, if you haven't already:
Would the CSS rules catch if we do `.panel-region .vbox{`?

map/manage-operational-layers/src/main/java/com/esri/samples/manage_operational_layers/ManageOperationalLayersSample.java
jenmerritt: @JonLavi tried it but sadly not 👎  I don't think the sample viewer even reads the CSS files from the samples, so the label styling that does apply is from the sample viewer CSS files themselves is my theory. We've been given a bit of scope this sprint to look at the SV. @sclaridge FYI on this conversation.

For now, I'll remove the CSS anyway, as no point duplicating. Thanks!

JonLavi (approved):
Thanks for confirming! Yes, would be great for someone to spend some time trying to get behind the whole SV and CSS interaction 👍

Rachael-E (approved):

PR #574, opened 2020-10-19
Configure Subnetwork Trace: Amend UI sizing

jenmerritt:
### Description

Following the feedback from verification on the different platforms and a discussion at review meeting, I've amended the UI slightly:
- Separated the label for defining new conditions onto a separate line to leave more space for the UI components for selecting the conditions. 
- Adjusted the widths of these combo boxes so that the text within them is always visible. 
- Aligned the whole UI top left

Have tested on Windows and SV and seems to work well - I could previously replicate the same feedback so should hopefully be ok on the other platforms too but will need to be checked again on Mac and Linux too.

Checklist:
- [x] Request a reviewer
- [x] Assign a reviewer
- [x] Tag reviewer in comment


utility_network/configure-subnetwork-trace/src/main/resources/configure_subnetwork_trace/main.fxml
```diff
@@ -28,17 +28,20 @@
...
             <CheckBox fx:id="includeContainersCheckBox" text="Include Containers"/>
         </HBox>
 
-        <HBox alignment="CENTER_LEFT">
-            <Label text="Define new conditions"/>
-            <ComboBox fx:id="comparisonSourcesComboBox" onAction="#onComparisonSourceChanged"/>
-            <ComboBox fx:id="comparisonOperatorsComboBox"/>
+        <HBox>
+            <Label text="Define new conditions:"/>
+        </HBox>
```
JonLavi: Instead of separating this, you might be able to add a CSS rule that puts some padding on the bottom of the label?
The easier alternative would be, since it's only one element, to get rid of the outer HBox here and the behaviour would be the same:

```suggestion
        <Label text="Define new conditions:"/>
```

utility_network/configure-subnetwork-trace/src/main/resources/configure_subnetwork_trace/style.css
```diff
@@ -16,33 +17,39 @@ HBox {
...
-.add-button {
-  -fx-min-width: 40;
+.comparison-sources-combobox {
+  -fx-min-width: 240;
+  -fx-max-width: 240;
 }
 
-.trace-button {
-  -fx-min-width: 545;
+.comparison-operators-combobox {
```
JonLavi: I _think_ it might not be necessary to create a 'styleClass' here, and we could target these using the `fx:id`:
```suggestion
#comparisonOperatorsComboBox {
```

Same would go for the above.

It looks like both parameters are the same for these comboboxes, so it might be possible to create one 'styleClass', or to just do
```
ComboBox{
  -fx-min-width: 240;
  -fx-max-width: 240;
}
```
but keep the extra rule for `.comparison-choice-elements ComboBox`. 
I think in CSS the more specific rule overrides the generic one, so this might be worth trying

utility_network/configure-subnetwork-trace/src/main/resources/configure_subnetwork_trace/style.css
```diff
@@ -16,33 +17,39 @@ HBox {
...
-.reset-button {
+.comparison-choice-elements TextField {
   -fx-min-width: 175;
+  -fx-max-width: 175;
+}
+
+.comparison-choice-elements ComboBox {
+  -fx-min-width: 175;
+  -fx-max-width: 175;
 }
```
JonLavi: If there's no other elements other than the one TextField and the one ComboBox inside the styleClass `.comparison-choice-elements `, you might be able to simply do one rule:
```
.comparison-choice-elements {
  -fx-min-width: 175;
  -fx-max-width: 175;
}
```

utility_network/configure-subnetwork-trace/src/main/resources/configure_subnetwork_trace/style.css
```diff
@@ -16,33 +17,39 @@ HBox {
   -fx-font-weight: normal;
 }
 
-
 TextArea {
   -fx-min-height: 181;
```
JonLavi: I am seeing this old number and thinking, could we make this `180`? 😄

JonLavi (commented):
Thanks for looking into this @jenmerritt! Fine-tuning the CSS is always a challenge, especially if it doesn't resonate with the SV in the end.
Functionality wise, the sample looks good to me. Simply putting everything top-left is probably the easiest solution for us.
Only a couple of ideas for refactoring:

You could add to `ConfigureSubnetworkTraceSample.java`:
```
    stage.setMinHeight(400);
    stage.setMinWidth(800);
```
That way, the user won't be able to make the window smaller than that size, meaning we don't have to deal with the case of the elements being cut by a smaller window.

utility_network/configure-subnetwork-trace/src/main/resources/configure_subnetwork_trace/style.css
jenmerritt: So had to go with using the `fxid` for those 2, as the `ComboBox` was overriding the other element in `comparison-choice-elements`. But it is still much tidier now!

jenmerritt:
Thanks @JonLavi , nice tidy ups. See what you think!

utility_network/configure-subnetwork-trace/src/main/resources/configure_subnetwork_trace/main.fxml
```diff
@@ -65,4 +63,4 @@
         </HBox>
     </VBox>
 
-</StackPane>
+</StackPane>
```
JonLavi: just missing a newline at the end of this file.

JonLavi (approved):
LGTM @jenmerritt

Rachael-E (approved):
looks good @jenmerritt !

PR #575, opened 2020-10-19
Sync main and dev

jenmerritt:
### Description

Syncing main and dev after manage operational layers fix was merged.


jenmerritt:
@Rachael-E One of the build checks has passed and one hasn't, says it's due to Show Location History using olympus which makes sense for `dev`. Unclear why the `master` branch is currently showing as failing too for the same reason when it doesn't even have that sample?

Rachael-E (approved):

PR #577, opened 2020-10-20
Edit with Branch Versioning: remove authentication handler

jenmerritt:
### Description

After some verification on the different platforms it was agreed at the review meeting to remove the authentication handler from the Edit with Branch Versioning samples, as it was proving tricky for users to enter credentials that were in the readme when they needed to complete authentication before they could access the readme. Other samples demonstrate how you can implement authentication.

So I have amended the readme, and removed the `AuthenticationChallengeHandler` in the Controller. Instead the user credentials are hard coded and set directly to the `ServiceGeodatabase`. 

I added a comment that the `AuthenticationChallengeHandler` could be used as an alternative. Any thoughts on this wording, and whether I could/should direct the user to a sample to look at, like 'oauth'? Wasn't sure if that was too much to put in the comments.

Checklist:
- [x] Request a reviewer
- [x] Assign a reviewer
- [x] Tag reviewer in comment

### DIFF
--- a/editing/edit-with-branch-versioning/src/main/java/com/esri/samples/edit_with_branch_versioning/EditWithBranchVersioningController.java
+++ b/editing/edit-with-branch-versioning/src/main/java/com/esri/samples/edit_with_branch_versioning/EditWithBranchVersioningController.java
@@ -90,14 +90,15 @@ public class EditWithBranchVersioningController {
         }
       });
 
-      // set the user credentials required to authenticate with the service geodatabase
-      // alternatively authentication could be handled with an AuthenticationChallengeHandler
-      UserCredential userCredential = new UserCredential("editor01", "editor01.password");
-
-      // create and load a service geodatabase
+      // create a service geodatabase
       serviceGeodatabase = new ServiceGeodatabase("https://sampleserver7.arcgisonline" +
         ".com/arcgis/rest/services/DamageAssessment/FeatureServer");
+
+      // set the user credentials required to authenticate with the service geodatabase
+      UserCredential userCredential = new UserCredential("editor01", "editor01.password");
       serviceGeodatabase.setCredential(userCredential);
+
+      // load the service geodatabase
       serviceGeodatabase.loadAsync();
 
       serviceGeodatabase.addDoneLoadingListener(() -> {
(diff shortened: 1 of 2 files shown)


editing/edit-with-branch-versioning/src/main/java/com/esri/samples/edit_with_branch_versioning/EditWithBranchVersioningController.java
```diff
@@ -91,13 +90,16 @@ public void initialize() {
...
+      // set the user credentials required to authenticate with the service geodatabase
+      // alternatively authentication could be handled with an AuthenticationChallengeHandler
+      UserCredential userCredential = new UserCredential("editor01", "editor01.password");
 
       // create and load a service geodatabase
       serviceGeodatabase = new ServiceGeodatabase("https://sampleserver7.arcgisonline" +
         ".com/arcgis/rest/services/DamageAssessment/FeatureServer");
+      serviceGeodatabase.setCredential(userCredential);
       serviceGeodatabase.loadAsync();
+
```
JonLavi, line 103: Only suggestion I can make here is perhaps creating the `UserCredential` immediately before setting it, so have it in order:
```
// create a service geodatabase
 serviceGeodatabase = new ServiceGeodatabase("https://sampleserver7.arcgisonline" +
".com/arcgis/rest/services/DamageAssessment/FeatureServer");

// set the user credentials required to authenticate with the service geodatabase
UserCredential userCredential = new UserCredential("editor01", "editor01.password");
serviceGeodatabase.setCredential(userCredential);

// load the service geodatabase
serviceGeodatabase.loadAsync();
```

I don't know if it's strictly necessary to mention the `AuthenticationChallengeHandler`, but I haven't been following the discussion. I suppose the authentication itself is not an integral part of what the sample is 
trying to show? In which case maybe we can do without the comment

JonLavi (approved):
LGTM @jenmerritt ! Just a small suggestion for ordering.

editing/edit-with-branch-versioning/src/main/java/com/esri/samples/edit_with_branch_versioning/EditWithBranchVersioningController.java
jenmerritt: @JonLavi Yeah exactly it was decided it wasn't necessary because it's not what the sample is demonstrating and was adding unnecessary steps for the user in showing off the branch versioning functionality. It's not a requirement to mention the authentication, so it's maybe overkill to add it into the comments, as there are clearly samples for authentication if someone wanted to add in the authentication. I'll reorder as you suggested and take that extra line out 👍

Rachael-E (approved):
thanks @jenmerritt these changes look good.

PR #578, opened 2020-10-26
Fix bugs from 100.8 regressions testing 

sclaridge:
This PR includes an enhancement for the viewshed samples and fixes some bugs from 100.8 regression testing:

- Removes the colour picker from Viewshed (location) sample and explicitly sets the colours of the visible, obstructed areas and frustum outline. This prevents the changes from the color picker from effecting the other viewshed samples in the sample viewer.
- Removes the unclickable duplicate cell from the transformations list in List transformations by suitability sample.
- Adds CSS colours to the labels in Control annotation sublayer visibility. Previously in the sample viewer, if you hovered over the open sublayer checkbox label when out of scale, it would switch from gray to white. 

Thanks!

@JonLavi


analysis/viewshed-location/src/main/java/com/esri/samples/viewshed_location/ViewshedLocationController.java
```diff
@@ -81,6 +76,11 @@ public void initialize() {
     LocationViewshed viewshed = new LocationViewshed(location, headingSlider.getValue(), pitchSlider.getValue(),
         horizontalAngleSlider.getValue(), verticalAngleSlider.getValue(), minDistanceSlider.getValue(),
         maxDistanceSlider.getValue());
+    // set the colors of the visible and obstructed areas
+    Viewshed.setVisibleColor(0xCC00FF00);
+    Viewshed.setObstructedColor(0xCCFF0000);
+    // set the color of the frustum outline
+    Viewshed.setFrustumOutlineColor(0xCC0000FF);
```
JonLavi: Good solution to the issue, to just go ahead and simplify it 👍 
We only need to update the screenshot to use the colours that we're setting here.

display_information/control-annotation-sublayer-visibility/src/main/resources/control_annotation_sublayer_visibility/style.css
```diff
@@ -0,0 +1,7 @@
+.panel-region .label {
+  -fx-text-fill: white;
+}
+
+.panel-region .check-box {
```
JonLavi: This only really applies to `closedSublayerCheckbox` right now, so I think we could specify by writing:
```suggestion
#closedSublayerCheckbox {
```
Of course this depends on whether this works ok with the SV, so best test it out!

geometry/list-transformations-by-suitability/src/main/java/com/esri/samples/list_transformations_by_suitability/ListTransformationsBySuitabilitySample.java
```diff
@@ -100,10 +100,13 @@ public void start(Stage stage) {
...
-        protected void updateItem(DatumTransformation transformation, boolean bln) {
+        protected void updateItem(DatumTransformation transformation, boolean empty) {
 
-          super.updateItem(transformation, bln);
-          if (transformation != null) {
+          super.updateItem(transformation, empty);
+          if (empty) {
+            setText(null);
+          }
+          else if (transformation != null) {
```
JonLavi: Good spot on this, it was quite a sneaky bug!

JonLavi (approved):
Thanks for fixing these @sclaridge .
Just a couple of suggestions. The only thing really needs updating is the screenshot for `ViewshedLocation`. 
Approving so that you can move on with the second review 👍

display_information/control-annotation-sublayer-visibility/src/main/resources/control_annotation_sublayer_visibility/style.css
sclaridge: This works in the SV!

sclaridge:
@JonLavi Thanks for the feedback, I've update the screenshot! 

Based on feedback from the samples team, I've added a small change so the frustum outline is visible when the sample starts up.

display_information/control-annotation-sublayer-visibility/src/main/java/com/esri/samples/control_annotation_sublayer_visibility/ControlAnnotationSublayerVisibilitySample.java
```diff
@@ -67,10 +68,9 @@ public void start(Stage stage) {
       // create checkboxes for toggling the sublayer visibility manually
       CheckBox closedSublayerCheckbox = new CheckBox();
       closedSublayerCheckbox.setSelected(true);
-      closedSublayerCheckbox.setTextFill(Color.WHITE);
       CheckBox openSublayerCheckbox = new CheckBox();
       openSublayerCheckbox.setSelected(true);
-      openSublayerCheckbox.setTextFill(Color.DARKGRAY);
+      openSublayerCheckbox.setStyle("-fx-text-fill: darkgrey");
```
Rachael-E: since this is the colour being set to the checkbox as the sample loads, this should be able to go into the .css file in the same way that the closedSublayerCheckbox is, e.g. 

```
#openSublayerCheckbox  {
   -fx-text-fill: darkgrey;
}
```

geometry/list-transformations-by-suitability/src/main/java/com/esri/samples/list_transformations_by_suitability/ListTransformationsBySuitabilitySample.java
```diff
@@ -100,10 +100,13 @@ public void start(Stage stage) {
       transformationsListView.setCellFactory(list -> new ListCell<DatumTransformation>() {
 
         @Override
-        protected void updateItem(DatumTransformation transformation, boolean bln) {
+        protected void updateItem(DatumTransformation transformation, boolean empty) {
```
Rachael-E: naming conventions for booleans can be vague with Java, but we can try and be more succinct with their naming here. something like `isEmpty` would be more readable here.
```suggestion
        protected void updateItem(DatumTransformation transformation, boolean isEmpty) {
```

Rachael-E (commented):
hey @sclaridge , thanks for pulling this together! I have two minor suggestions, and I'll come back to the Viewshed colour part after we hear back from the review tomorrow after our chat. Thanks!

geometry/list-transformations-by-suitability/src/main/java/com/esri/samples/list_transformations_by_suitability/ListTransformationsBySuitabilitySample.java
sclaridge: @Rachael-E At our review meeting on Friday, the samples team confirmed that we'll remove the colour picking functionality. Setting the colours explicitly in this sample shows off the ability to configure the colours of the viewshed areas without impacting the other samples.

Rachael-E (approved):
thanks @sclaridge for taking that to review, this is good to go ahead and merge then!

PR #579, opened 2020-10-27
Add draggable graphics to Update Graphics sample

sclaridge:
This PR adds the [enhancements](https://github.com/Esri/arcgis-runtime-samples-java/pull/507) suggested by Tyler to the Update Graphics sample. Now you can select a graphic and drag it to a new location. The cursor changes from default to `Cursor.MOVE`. To finish moving the graphic, you can press Escape to then select another graphic. 

I would like some feedback on:

-  renamed `enableUI()` to `disableUI()`, which I think is more intuitive? 
- I added  `mapView.requestFocus()` as I otherwise could not generate key events in the SV. I think this is because the mapView was not in focus. 

Thanks! 

@JonLavi


display_information/update-graphics/README.md
```diff
@@ -11,8 +11,8 @@ A field worker may want to update a graphic's properties to adjust which real-wo
 ## How to use the sample
 
 Update Graphic's Location:
-  - click on a graphic, then click the update location button
-  - now click a new location on the map to move graphic
+  - click on a graphic, then drag the graphic to a new location
```
JonLavi: It might be really good to have the ability to directly drag graphics without clicking them first. It might complicate the code a bit, but worth looking into.

I've notice that if I try to drag without the initial click to select the graphics first, there's a `NullPointerException`, meaning we need to check on drag whether there's a graphics selected.

display_information/update-graphics/README.md
```diff
@@ -11,8 +11,8 @@ A field worker may want to update a graphic's properties to adjust which real-wo
...
+  - press ESCAPE to stop moving the graphic
```
JonLavi: Technically escape doesn't stop the moving, it de-selects the graphic. Also, it's not just escape, but a single click outside the graphic will also de-select it.

display_information/update-graphics/README.md
```diff
@@ -26,13 +26,13 @@ Update Graphic's Symbol:
 1. Capture clicks on the `MapView` by using `.setOnMouseClicked()`. Then use the event to create a `Point` from the clicked location. 
 2. Identify the clicked graphics using `MapView.identifyGraphicsOverlayAsync(graphicsOverlay, pointClicked, tolerance, max results)`.
 3. To select a clicked graphics, iterate through the list of graphics returned by the identification method, and set each graphic's selection property to `true`.
-4. To update graphic's location, create a new `Point` from the user's clicked point, and use `Graphic.setGeometry(point)` to modify the graphic's geometry.
+4. To update a graphic's location, capture drags on the `MapView` using `.setOnMouseDragged()`, and use `Graphic.setGeometry(point)` to modify the graphic's geometry from the dragged location.
```
JonLavi: We've already used the term `MapView` in step 1, so I think it'd be ok to write `map view` in plain English?

display_information/update-graphics/src/main/java/com/esri/samples/update_graphics/UpdateGraphicsSample.java
```diff
@@ -90,18 +92,15 @@ public void start(Stage stage) {
...
       // create buttons for user interaction
-      updateLocationButton = new Button("Update Location");
       updateDescriptionButton = new Button("Update Description");
-      updateLocationButton.setMaxWidth(Double.MAX_VALUE);
       updateDescriptionButton.setMaxWidth(Double.MAX_VALUE);
-      updateLocationButton.setDisable(true);
       updateDescriptionButton.setDisable(true);
 
-      // create section for combo box
+      // create combo box
```
JonLavi: While we're at it, lets make these comments plain English?
Peraps `// create the combo box for the UI`?

display_information/update-graphics/src/main/java/com/esri/samples/update_graphics/UpdateGraphicsSample.java
```diff
@@ -117,8 +116,8 @@ public void start(Stage stage) {
         }
       });
 
-      // add label, dropdown and buttons to the control panel
-      controlsVBox.getChildren().addAll(updateLocationButton, updateDescriptionButton, symbolLabel, symbolBox);
+      // add label, combo box and buttons to the control panel
```
JonLavi: Another suggestion for plain english? Also, we only have one button now, by the looks of it? Would also be nice to order the items in the comment in the same order as they are in the code:

```suggestion
      // add the button, label, and combo box to the control panel
```

display_information/update-graphics/src/main/java/com/esri/samples/update_graphics/UpdateGraphicsSample.java
```diff
@@ -260,10 +281,14 @@ private void createGraphics() {
...
-  private void enableUI(boolean enable) {
-    updateDescriptionButton.setDisable(enable);
-    updateLocationButton.setDisable(enable);
-    symbolBox.setDisable(enable);
+  /**
+   * Disables the visibility of the UI controls
+   *
+   * @param disable visibility of the UI
+   */
+  private void disableUI(boolean disable) {
```
JonLavi: Good catch on renaming this, definitely makes more sense 👍

display_information/update-graphics/src/main/java/com/esri/samples/update_graphics/UpdateGraphicsSample.java
```diff
@@ -167,39 +159,68 @@ public void start(Stage stage) {
...
-          }
+            } catch (Exception x) {
+              new Alert(Alert.AlertType.ERROR, "Error identifying clicked graphic").show();
+            }
+          });
+        }
+      });
+
+      mapView.setOnMouseDragged(e -> {
+        if (selectedGraphic.isSelected()) {
```
JonLavi: It looks like we need a null check here, to avoid a `NullPointerException` when no graphics is selected and we're dragging around:
```suggestion
        if (selectedGraphic != null && selectedGraphic.isSelected()) {
```

display_information/update-graphics/src/main/java/com/esri/samples/update_graphics/UpdateGraphicsSample.java
```diff
@@ -167,39 +159,68 @@ public void start(Stage stage) {
...
-            });
-          }
+            } catch (Exception x) {
+              new Alert(Alert.AlertType.ERROR, "Error identifying clicked graphic").show();
+            }
+          });
+        }
+      });
+
+      mapView.setOnMouseDragged(e -> {
```
JonLavi: It would be really cool if we could identify any potential graphic at the clicked point, select it, and then allow to drag if the mouse pointer is held down. We might be able to do that by extracting a method e.g. `identifyGraphicAtClickedPoint`, and then allowing to select it. Could be a bit tricky to do it elegantly, so I would suggest don't spend too much time on it if the code starts growing arms and legs!

display_information/update-graphics/src/main/java/com/esri/samples/update_graphics/UpdateGraphicsSample.java
```diff
@@ -167,39 +159,68 @@ public void start(Stage stage) {
...
+        if (selectedGraphic.isSelected()) {
+          // set the cursor to the move
+          mapView.setCursor(Cursor.MOVE);
+
+          // disable to UI while moving the graphic
+          disableUI(true);
```
JonLavi: Ideally we should be enabling the UI again after we stop dragging, because the graphics is still selected and the user might want to update it's properties?
Strictly speaking, we don't have to disable the UI while dragging at all, since we have to stop dragging anyway to interact with the UI properly? Might be worth a play around with it just to see which works best.

display_information/update-graphics/src/main/java/com/esri/samples/update_graphics/UpdateGraphicsSample.java
```diff
@@ -51,9 +54,7 @@
 
 public class UpdateGraphicsSample extends Application {
 
-  private boolean isUpdateLocationActive;
   private List<SimpleMarkerSymbol> markers;
-  private Button updateLocationButton;
   private Button updateDescriptionButton;
   private ComboBox<String> symbolBox;
```
JonLavi, line 59: @Rachael-E on the line below we have the comment regarding keeping `map` in scope to avoid garbage collection. I can't see any part of the code where this might happen here? The only async operation is on `mapView`:
```
mapView.identifyGraphicsOverlayAsync
```

Was it decided to leave these comments in as a standard, or can we remove this one and make `map` a local member?

JonLavi (commented):
Nice work taking this one on @sclaridge 
There's a couple of suggestions to perhaps expand the functionality even further, as long as we use our judgement when the code is getting too complicated.
Only immenent thing to fix is the null pointer exception when trying to drag on an area in the map view which doesn't have a graphic, the rest is up to you to take or leave as you see fit 👍

display_information/update-graphics/src/main/java/com/esri/samples/update_graphics/UpdateGraphicsSample.java
```diff
@@ -167,39 +159,68 @@ public void start(Stage stage) {
...
+          // disable to UI while moving the graphic
+          disableUI(true);
+
+          // create a point from the dragged location
+          mapViewPoint = new Point2D(e.getX(), e.getY());
+          Point mapPoint = mapView.screenToLocation(mapViewPoint);
+
+          // update the location of the graphic to the dragged location
+          selectedGraphic.setGeometry(mapPoint);
+          mapView.requestFocus();
```
JonLavi: I had to look this one up a bit. Not sure which element of the SV takes focus away from the map view, but I don't think we should include the code in the sample if it's just a workaround for the SV. Can't provide an alternative off the top of my head, but we might want to do a session to check what we can do to avoid adding it here.

sclaridge: Based on this [link](https://docs.oracle.com/javafx/2/events/processing.htm), for key events the target is the node that has focus, but for mouse events the target is the node at the location of the cursor. I tried to use `stackpane.onKeyPressed()` like the Create and Save KML File sample, but I didn't have any luck.

Would you recommended scrapping the ESC key functionality? A second click will de-select and stop moving the point.

JonLavi: That's probably a good idea, considering the ESC functionality it an added bonus, but that extra line of code is a bit confusing!

sclaridge:
@JonLavi I've removed the ESC key action. Now you can select and move the graphic, and a second click de selects the graphic. The UI is visible while the graphic is selected so you can change the symbol after a move. 

I had a hard time finding a way to drag the graphic without first selecting it. Let me know what you think of the current workflow, and I always give it another go if you think it can be improved.

display_information/update-graphics/README.md
```diff
@@ -12,7 +12,6 @@ A field worker may want to update a graphic's properties to adjust which real-wo
 
 Update Graphic's Location:
   - click on a graphic, then drag the graphic to a new location
```
JonLavi: Should we add a mention now to click outside of a graphic to de-select it?

JonLavi (approved):
Looks great @sclaridge !
No worries about direct click&drag, I also think it's best to keep it simple to just show how the principle works.

display_information/update-graphics/README.md
```diff
@@ -11,8 +11,8 @@ A field worker may want to update a graphic's properties to adjust which real-wo
 ## How to use the sample
 
 Update Graphic's Location:
```
Rachael-E: From a readme style perspective, this should be a " plain english" approach rather than a list. e.g. something like

```suggestion
To update the graphic's location, click on a graphic and then drag it to a new location. Use the "Update Description" button to provide a new description for the graphic, and use the "Update Symbol" drop down menu to choose a new symbol. Click away from the graphic to de-select it. 
```

display_information/update-graphics/src/main/java/com/esri/samples/update_graphics/UpdateGraphicsSample.java
```diff
@@ -51,17 +53,14 @@
...
-  private boolean isUpdateLocationActive;
   private List<SimpleMarkerSymbol> markers;
-  private Button updateLocationButton;
   private Button updateDescriptionButton;
   private ComboBox<String> symbolBox;
 
-  private ArcGISMap map; // keep loadable in scope to avoid garbage collection
   private MapView mapView;
-  private Graphic selectedGraphic;
+  private Graphic identifiedGraphic = null;
```
Rachael-E: we don't have to assert it's null here, the compiler will auto set all member variables to a reasonable default, generally null in this case.
```suggestion
  private Graphic identifiedGraphic;
```

display_information/update-graphics/src/main/java/com/esri/samples/update_graphics/UpdateGraphicsSample.java
```diff
@@ -167,39 +156,52 @@ public void start(Stage stage) {
...
+              if (!identifyGraphics.get().getGraphics().isEmpty()) {
+                // get the first identified graphic
+                identifiedGraphic = identifyGraphics.get().getGraphics().get(0);
+                // select the identified graphic
+                identifiedGraphic.setSelected(true);
+                // update the drop down box with the identified graphic's current symbol
+                String style = ((SimpleMarkerSymbol) identifiedGraphic.getSymbol()).getStyle().toString();
+                symbolBox.getSelectionModel().select(style);
+                // show the UI
+                disableUI(false);
```
Rachael-E, line 183: A few things in this suggestion. 

1. When the sample loads, and the user clicks and drags on the map view, a null pointer exception is thrown. This continues until a user clicks to select a graphic, and then it's not seen again. We could handle that by showing an error in the mapView.setOnMouseDragged listener, or, by simply moving that listener in the identifyGraphics.addDoneListener (see suggestion).

The benefit to this approach is that we know the identifiedGraphic should never be null (since it's already identified), and a null pointer shouldn't be able to be thrown. Also when the sample loads, a user could click and drag the map view if they wanted to before it gets locked when selecting a graphic.

2.  If a user selects a graphic but clicks and drags on the other side of the screen, the point is dragged. I'd expect the graphic only to be dragged if the cursor was actually on the graphic. I think Jon has suggested something similar above to address this but might have too many code arms and legs to follow phrase! So to make it clear to the user that the graphic is still selected and in drag mode, I've suggested a simple `Cursor.CLOSED_HAND` here to make it visually obvious we're still dragging. 

```suggestion
                disableUI(false);

                // enable dragging of the identified graphic to move its location
                mapView.setOnMouseDragged(event -> {

                    if (identifiedGraphic.isSelected() && identifiedGraphic != null) {
                      // set the cursor to closed hand to indicate graphic dragging is active
                      mapView.setCursor(Cursor.CLOSED_HAND);
                      // create a point from the dragged location
                      mapViewPoint = new Point2D(event.getX(), event.getY());
                      Point mapPoint = mapView.screenToLocation(mapViewPoint);
                      // update the location of the graphic to the dragged location
                      identifiedGraphic.setGeometry(mapPoint);
                    } else new Alert(Alert.AlertType.ERROR, "Error selecting graphic").show();
                });
```

Rachael-E: Alternatively, it is possible to simultaneously click and drag the graphic with the following. 

However, I had to bump the tolerance up quite high since there is a bit of lag with the drag and if the mouse travels faster than the graphic, it leaves it behind. With this bigger tolerance though, more than one graphic can be selected and it results in a fun dance between the two graphics moving...if you'd like to take this code and work around it to avoid that behaviour, see what you think: it might take a fair bit of time, I've had to leave it for now. Otherwise the workflow I've suggested above is just fine! 

```java
      mapView.setOnMouseDragged(e -> {
        if (e.getButton() == MouseButton.PRIMARY) {
//           create a point where the user clicked
          mapViewPoint = new Point2D(e.getX(), e.getY());

          // identify graphics on the graphics overlay
          ListenableFuture<IdentifyGraphicsOverlayResult> identifyGraphics =
            mapView.identifyGraphicsOverlayAsync(graphicsOverlay, mapViewPoint, 50, false);

          identifyGraphics.addDoneListener(() -> {
            try {
              if (!identifyGraphics.get().getGraphics().isEmpty()) {
                // get the first identified graphic
                identifiedGraphic = identifyGraphics.get().getGraphics().get(0);
                // select the identified graphic
                identifiedGraphic.setSelected(true);
                // update the drop down box with the identified graphic's current symbol
                String style = ((SimpleMarkerSymbol) identifiedGraphic.getSymbol()).getStyle().toString();
                symbolBox.getSelectionModel().select(style);
                // show the UI
                disableUI(false);

                // enable dragging of the identified graphic to move its location
                if (identifiedGraphic.isSelected() && identifiedGraphic != null) {
                  // set the cursor to closed hand to indicate graphic dragging is active
//                      mapView.setCursor(Cursor.CLOSED_HAND);
                  // create a point from the dragged location
                  mapViewPoint = new Point2D(e.getX(), e.getY());
                  Point mapPoint = mapView.screenToLocation(mapViewPoint);
                  // update the location of the graphic to the dragged location
                  identifiedGraphic.setGeometry(mapPoint);
                } else new Alert(Alert.AlertType.ERROR, "Error selecting graphic").show();

                // clear any selected graphic when clicking on the map view
                mapView.setOnMouseClicked(mouseEvent -> {
                  if (mouseEvent.isStillSincePress()) {
                    graphicsOverlay.clearSelection();
                  }
                });

              } else {
                disableUI(true);
              }
            } catch (Exception x) {
              new Alert(Alert.AlertType.ERROR, "Error identifying clicked graphic").show();
            }
          });
        }
      });
```

display_information/update-graphics/src/main/java/com/esri/samples/update_graphics/UpdateGraphicsSample.java
```diff
@@ -260,10 +262,14 @@ private void createGraphics() {
     }
   }
 
-  private void enableUI(boolean enable) {
-    updateDescriptionButton.setDisable(enable);
-    updateLocationButton.setDisable(enable);
-    symbolBox.setDisable(enable);
+  /**
+   * Disables the visibility of the UI controls
```
Rachael-E: Missing a period from this javadoc 
```suggestion
   * Disables the visibility of the UI controls.
```

display_information/update-graphics/src/main/java/com/esri/samples/update_graphics/UpdateGraphicsSample.java
```diff
@@ -260,10 +262,14 @@ private void createGraphics() {
...
+   *
+   * @param disable visibility of the UI
+   */
+  private void disableUI(boolean disable) {
```
Rachael-E: let's give the boolean parameter a slightly more meaningful name

```suggestion
  private void disableUI(boolean isDisabled) {
```

display_information/update-graphics/src/main/java/com/esri/samples/update_graphics/UpdateGraphicsSample.java
Rachael-E: @sclaridge perhaps Jon's suggestion combined with my attempt below might work? Like Jon says though, don't spend too long on it, we have a workable solution from the first suggestion I put up.

display_information/update-graphics/src/main/java/com/esri/samples/update_graphics/UpdateGraphicsSample.java
```diff
@@ -167,39 +156,52 @@ public void start(Stage stage) {
...
+            } catch (Exception x) {
+              new Alert(Alert.AlertType.ERROR, "Error identifying clicked graphic").show();
+            }
+          });
+        }
+      });
+
+      mapView.setOnMouseDragged(e -> {
+        if (identifiedGraphic.isSelected() && identifiedGraphic != null) {
+          // set the cursor to the move
```
Rachael-E:
```suggestion
          // set the cursor to the move icon
```

Rachael-E (changes requested):
@sclaridge nice updates to this sample, thanks. I've put up a suggestion to avoid triggering a null pointer exception when the sample opens, and a suggestion for the cursor icon change when dragging a graphic. 

I've also put up an alternative but not water-tight suggestion for a simultaneous click and drag on the graphic. It has its flaws, you may have more time to play around with it. See what you think but don't lose too much time: we can go for my first suggestion if time creeps on, no problem. Give me a shout if you have any questions.

sclaridge:
@Rachael-E After some trial and error, I think your [implementation](https://github.com/Esri/arcgis-runtime-samples-java/pull/579#discussion_r518258579) is clear, logical, and prevents the null pointer exception from being thrown. I've added those changes and updated the image and gif.

Rachael-E (approved):
this looks great! thanks @sclaridge

PR #581, opened 2020-10-30
Add Exclusive Mode to Group Layers Sample

jenmerritt:
### Description

I've added in Exclusive mode to the Group Layers Sample. 

The grouping of the layers has changed into 2 groups:
- `projectAreaGroupLayer`
- `buildingsGroupLayer`

The latter has `GroupVisibilityMode.EXCLUSIVE`. To accommodate this this group also has a toggle group and radio buttons in the `TreeView`. 

There was also an issue noted for the Java sample in relation to this sample, about the pathways layer showing underneath the development project area. This has been fixed by the way the 2 layers are now grouped and layered on the operational layers.

One thing I noted is that the layers themselves are loaded in the `buildLayersView` method, so I think they are added to the `TreeView` at the rate they are loaded, which can vary so they aren't always displayed in the list in the same order. If we want this to be consistent between the order on the scene and UI, then a bit like with the operational layers sample, we'd need to load each layer before that method and then reverse the order they're added to the `TreeView`. Personally not sure the additional code is worth is to keep the focus on the Group Layers but let me know if you want to discuss this more. I'll also show at Review later.

Checklist:
- [x] Request a reviewer
- [x] Assign a reviewer
- [x] Tag reviewer in comment


jenmerritt:
@JonLavi up first :)

group_layers/group-layers/src/main/java/com/esri/samples/group_layers/GroupLayersSample.java
```diff
@@ -78,40 +82,59 @@ public void start(Stage stage) {
...
+        (new ArcGISSceneLayer("https://tiles.arcgis.com/tiles/P3ePLMYs2RVChkJx/arcgis/rest/services/DevA_Trees/SceneServer"))
+      ));
+
+      buildingsGroupLayer = new GroupLayer();
+      buildingsGroupLayer.setName("Buildings group");
+      buildingsGroupLayer.setVisibilityMode(GroupVisibilityMode.EXCLUSIVE);
+      buildingsGroupLayer.getLayers().addAll(Arrays.asList(
+        (new ArcGISSceneLayer("https://tiles.arcgis.com/tiles/P3ePLMYs2RVChkJx/arcgis/rest/services/DevA_BuildingShells/SceneServer")),
+        (new ArcGISSceneLayer("https://tiles.arcgis.com/tiles/P3ePLMYs2RVChkJx/arcgis/rest/services/DevB_BuildingShells/SceneServer"))
+      ));
```
JonLavi: Maybe it's worth considering (two?) method(s) here?
`private void FeatureLayer(url URL)`
`private void addArcGISSceneLayer(url URL)`
I'll let you and @Rachael-E discuss whether that's applicable here.

group_layers/group-layers/src/main/java/com/esri/samples/group_layers/GroupLayersSample.java
```diff
@@ -78,40 +82,59 @@ public void start(Stage stage) {
...
+
+      // Display an alert if the child layers in the group layers fail to load
+      projectAreaGroupLayer.getLayers().forEach(layer ->
+        layer.addDoneLoadingListener(() -> {
```
JonLavi: Perhaps adding a `.loadAsync` above this line (and below when iterating over buildingsGroupLayer) would help with the wait happening when switching between the different group layers

group_layers/group-layers/src/main/java/com/esri/samples/group_layers/LayerTreeCell.java
```diff
@@ -1,32 +1,71 @@
...
+      if (layer.getName().equals("DevA_BuildingShells") || layer.getName().equals("DevB_BuildingShells")) {
+
+        // if the layer is a building shell create a radio button and assign to the toggle group
+        RadioButton radioButton = new RadioButton();
+        setGraphic(radioButton);
+        radioButton.setToggleGroup(GroupLayersSample.buildingsToggleGroup);
+
+        // toggle the layer's visibility when the radio button is selected
+        radioButton.setSelected(layer.isVisible());
+        radioButton.selectedProperty().addListener(e -> layer.setVisible(radioButton.isSelected()));
```
JonLavi, line 48: There might be a possibility to use property binding here, but no need to spend too much time trying to optimise it if you find it's getting out of hand a bit.

JonLavi (commented):
A general issue I see with the sample is that there's a noticable delay between ticking the checkbox and the respective layer appearing/disappearing. Not sure whether there's anything we can do about this?

Also wondering whether it'd be sensible to have both buildings layers load at startup. The reason is that at the moment, it may take a while for anything to appear when switching to 'Buildings B', because the loading is only triggered at that point.

Happy to chat through these in more detail!

group_layers/group-layers/src/main/java/com/esri/samples/group_layers/GroupLayersSample.java
jenmerritt: I followed what we did in another sample adding this way, so will leave as is to be checked/discussed if needed 👍

group_layers/group-layers/src/main/java/com/esri/samples/group_layers/GroupLayersSample.java
jenmerritt: @JonLavi The last few days I've tested a load of strategies in terms of how the layers are loaded and also added to the UI (which is currently in the order at which the layers are loaded, rather than the order they are on the map) to see if/what improvements can be made.  @TADraeseke and I ran through it and compared with Android too. Conclusion has been that I've moved the loading out of this method (though the switch speed still seems to just be the layers themselves). I've also left the ordering pattern as it is for the UI.

I've saved the testing on a branch so if you or @Rachael-E  wanted me to share/discuss those just let me know.

group_layers/group-layers/src/main/java/com/esri/samples/group_layers/LayerTreeCell.java
jenmerritt: I had a look at this but couldn't figure out how I would do that. Could look at this together, or leave it if it's ok with the listeners.

JonLavi: Listeners might be the way forward actually, because it doesn't look like we can bind `layer.isVisible`

JonLavi (approved):
LGTM @jenmerritt, nice improvements

group_layers/group-layers/src/main/java/com/esri/samples/group_layers/GroupLayersSample.java
```diff
@@ -43,10 +46,11 @@
 
 public class GroupLayersSample extends Application {
 
+  private GroupLayer buildingsGroupLayer; // keep loadable in scope to avoid garbage collection
+  private GroupLayer projectAreaGroupLayer; // keep loadable in scope to avoid garbage collection
+
```
Rachael-E: I've re-reviewed this from the garbage collection perspective and we don't actually need these up here (I missed this the first time round!)

We don't need to worry about garbage collection since we add them to the scene's operational layers, and since the scene is attached to the sceneview, the sceneview can't be garbage collected.

group_layers/group-layers/src/main/java/com/esri/samples/group_layers/GroupLayersSample.java
```diff
@@ -78,40 +82,63 @@ public void start(Stage stage) {
...
-
-      // add the group layer and other layers to the scene as operational layers
-      scene.getOperationalLayers().addAll(Arrays.asList(groupLayer, devBBuildings, devProjectArea));
-
-      // zoom to the extent of the group layer when the child layers are loaded
-      groupLayer.getLayers().forEach(childLayer ->
-        childLayer.addDoneLoadingListener(() -> {
-          if (childLayer.getLoadStatus() == LoadStatus.LOADED) {
-            sceneView.setViewpointCamera(new Camera(groupLayer.getFullExtent().getCenter(), 700, 0, 60, 0));
+      // create two group layers, set the visibility modes and add layers
```
Rachael-E:
```suggestion
      // create a group layer for project areas, set the visibility mode and add layers to it
```

group_layers/group-layers/src/main/java/com/esri/samples/group_layers/GroupLayersSample.java
```diff
@@ -78,40 +82,63 @@ public void start(Stage stage) {
...
+      projectAreaGroupLayer = new GroupLayer();
+      projectAreaGroupLayer.setName("Project area group");
+      projectAreaGroupLayer.setVisibilityMode(GroupVisibilityMode.INDEPENDENT);
+      projectAreaGroupLayer.getLayers().addAll(Arrays.asList(
+        (new FeatureLayer(new ServiceFeatureTable("https://services.arcgis.com/P3ePLMYs2RVChkJx/arcgis/rest/services/DevelopmentProjectArea/FeatureServer/0"))),
+        (new FeatureLayer(new ServiceFeatureTable("https://services.arcgis.com/P3ePLMYs2RVChkJx/arcgis/rest/services/DevA_Pathways/FeatureServer/1"))),
+        (new ArcGISSceneLayer("https://tiles.arcgis.com/tiles/P3ePLMYs2RVChkJx/arcgis/rest/services/DevA_Trees/SceneServer"))
+      ));
+
```
Rachael-E, line 95:
```suggestion
  // create a group layer for buildings, set the visibility mode and add layers to it
```

group_layers/group-layers/src/main/java/com/esri/samples/group_layers/GroupLayersSample.java
Rachael-E: happy with this as is, although the urls are going way over our line limit (around 110 chars) recommendation. adding a line break before the string part of each code line should get over that more or less.

group_layers/group-layers/src/main/java/com/esri/samples/group_layers/GroupLayersSample.java
```diff
@@ -78,40 +82,63 @@ public void start(Stage stage) {
...
+      buildingsGroupLayer.getLayers().addAll(Arrays.asList(
+        (new ArcGISSceneLayer("https://tiles.arcgis.com/tiles/P3ePLMYs2RVChkJx/arcgis/rest/services/DevA_BuildingShells/SceneServer")),
+        (new ArcGISSceneLayer("https://tiles.arcgis.com/tiles/P3ePLMYs2RVChkJx/arcgis/rest/services/DevB_BuildingShells/SceneServer"))
+      ));
+
+      // load the group layers
+      projectAreaGroupLayer.loadAsync();
+      buildingsGroupLayer.loadAsync();
+
+      // Display an alert if the child layers in the group layers fail to load
```
Rachael-E:
```suggestion
      // display an alert if the child layers in the group layers fail to load
```

group_layers/group-layers/src/main/java/com/esri/samples/group_layers/GroupLayersSample.java
```diff
@@ -78,40 +82,63 @@ public void start(Stage stage) {
...
+      buildingsGroupLayer.setName("Buildings group");
+      buildingsGroupLayer.setVisibilityMode(GroupVisibilityMode.EXCLUSIVE);
+      buildingsGroupLayer.getLayers().addAll(Arrays.asList(
+        (new ArcGISSceneLayer("https://tiles.arcgis.com/tiles/P3ePLMYs2RVChkJx/arcgis/rest/services/DevA_BuildingShells/SceneServer")),
+        (new ArcGISSceneLayer("https://tiles.arcgis.com/tiles/P3ePLMYs2RVChkJx/arcgis/rest/services/DevB_BuildingShells/SceneServer"))
+      ));
+
+      // load the group layers
+      projectAreaGroupLayer.loadAsync();
+      buildingsGroupLayer.loadAsync();
```
Rachael-E: the pattern for group layers is that they are "loaded" by default when they are brought into the map, so they don't need this additional load call.
```suggestion
```

group_layers/group-layers/src/main/java/com/esri/samples/group_layers/GroupLayersSample.java
```diff
@@ -78,40 +82,63 @@ public void start(Stage stage) {
...
+
+      // Display an alert if the child layers in the group layers fail to load
+      projectAreaGroupLayer.getLayers().forEach(layer ->
```
Rachael-E: I think it's more readable to name each layer childLayer as we had before: 

```suggestion
      projectAreaGroupLayer.getLayers().forEach(childLayer ->
```

group_layers/group-layers/src/main/java/com/esri/samples/group_layers/GroupLayersSample.java
```diff
@@ -78,40 +82,63 @@ public void start(Stage stage) {
...
+
+      buildingsGroupLayer.getLayers().forEach(layer ->
+        layer.addDoneLoadingListener(() -> {
+          if (layer.getLoadStatus() != LoadStatus.LOADED) {
+            new Alert(Alert.AlertType.ERROR, "Layer failed to load:\n" + layer.getLoadError().getCause().getMessage()).show();
           }
         })
       );
 
+      // add the group layers to the scene operational layers
```
Rachael-E:
```suggestion
      // add the group layers to the scene as operational layers
```

group_layers/group-layers/src/main/java/com/esri/samples/group_layers/GroupLayersSample.java
```diff
@@ -78,40 +82,63 @@ public void start(Stage stage) {
...
+      scene.getOperationalLayers().addAll(Arrays.asList(projectAreaGroupLayer, buildingsGroupLayer));
+
```
Rachael-E: from a workflow order perspective, let's move this up after the buildingsGroupsLayer set up, and before the layer is done loading to set the viewpoint camera.

group_layers/group-layers/src/main/java/com/esri/samples/group_layers/GroupLayersSample.java
```diff
@@ -78,40 +82,63 @@ public void start(Stage stage) {
...
+        })
+      );
+
+      buildingsGroupLayer.getLayers().forEach(layer ->
+        layer.addDoneLoadingListener(() -> {
+          if (layer.getLoadStatus() != LoadStatus.LOADED) {
+            new Alert(Alert.AlertType.ERROR, "Layer failed to load:\n" + layer.getLoadError().getCause().getMessage()).show();
           }
         })
       );
```
Rachael-E: I think we can add the alert into `buildLayersView` where it's already looping through all the layers in the sample anyway and saves us repeating some code. If that doesn't work out, perhaps an additional method to save repeating the code? 
```suggestion
      // zoom to the extent of the group layer when the child layers are loaded
      projectAreaGroupLayer.getLayers().forEach(childLayer ->
        childLayer.addDoneLoadingListener(() -> {
          if (childLayer.getLoadStatus() == LoadStatus.LOADED) {
            // zoom to the extent of the project area group layer when the child layers are loaded
            sceneView.setViewpointCamera(new Camera(projectAreaGroupLayer.getFullExtent().getCenter(), 700, 0, 60, 0));
          }
        })
      );
```

group_layers/group-layers/src/main/java/com/esri/samples/group_layers/GroupLayersSample.java
```diff
@@ -132,17 +159,18 @@ public void start(Stage stage) {
    */
   private void buildLayersView(TreeItem<Layer> parentItem, List<Layer> operationalLayers) {
     for (Layer layer : operationalLayers) {
-      // load each layer before adding to ensure all metadata is ready for display
-      layer.loadAsync();
+      // check the layer has loaded before creating UI
```
Rachael-E: reinstate the loadAsync after this line: as discussed with Gunther, for the current sample workflow this is required although it isn't best practice. We can update the design later.

group_layers/group-layers/README.md
```diff
@@ -18,7 +18,11 @@ The layers in the map will be displayed in a table of contents. Toggle the check
...
 1. Create an empty `GroupLayer`.
 2. Add a child layer to the group layer's layers collection.
-3. To toggle the visibility of the group, simply change the group layer's visibility property using `layer.setVisible(boolean)`.
+3. Set the group layer's `GroupVisibilityMode` to change its behavior.
+- `GroupVisibilityMode.INDEPENDENT` allows each sublayer to change its visibility independently.
+- `GroupVisibilityMode.EXCLUSIVE` allows only one sublayer to be visible at a time.
+- `GroupVisibilityMode.INHERITED` treats the group layer as if it is one merged layer.
+4. To toggle the visibility of the group, simply change the group layer's visibility property using `layer.setVisible(boolean)`.
 
 ## Relevant API
```
Rachael-E, line 28: I think it would be appropriate to add `GroupVisibilityMode` in here as a new relevant API: might need updating in the common-samples repo

group_layers/group-layers/build.gradle
```diff
@@ -6,7 +6,7 @@ plugins {
 group = 'com.esri.samples'
 
 ext {
-    arcgisVersion = '100.9.0'
+    arcgisVersion = '100.10.0-2950'
```
Rachael-E, line 9: might get a merge conflict in this file with the updated gradle/jfrog changes: or hopefully it just updates no problem!

Rachael-E (commented):
looks good @jenmerritt! Following our chat with Gunther about loading patterns, I've made comments to put the loadAsync back where it was until we have a new design. Otherwise, minor things to consider.

group_layers/group-layers/src/main/java/com/esri/samples/group_layers/GroupLayersSample.java
jenmerritt: Added this and tested and it works well 👍

Rachael-E (approved):
looks great @jenmerritt !

PR #582, opened 2020-10-30
Remove deprecated Gradle features from samples 

sclaridge:
This PR makes the following changes to the build.gradle files to remove the features that will be deprecated in Gradle 7.0 :

- replace `id 'org.openjfx.javafxplugin' version '0.0.5`' with `id 'org.openjfx.javafxplugin' version '0.0.8'` 

- replace `compile` with `implementation` under dependencies `compile "com.esri.arcgisruntime:arcgis-java:$arcgisVersion"`

- replace the `http` with `https` for `maven {url 'https://olympus.esri.com/artifactory/arcgisruntime-repo'}`

- under `task productionZip(type: Zip)` replace `baseName` with `archiveBaseName` and replace `destinationDir` with `destinationDirectory` 

Thanks! 

@JonLavi

### DIFF
--- a/analysis/analyze-hotspots/build.gradle
+++ b/analysis/analyze-hotspots/build.gradle
@@ -64,7 +64,7 @@ jar {
 task productionZip(type: Zip) {
     group = 'distribution'
     from copyNatives
-    from jar.destinationDir
+    from jar.destinationDirectory
     into (project.name)
     baseName = project.name
 }
--- a/analysis/analyze-hotspots/build.gradle
+++ b/analysis/analyze-hotspots/build.gradle
@@ -66,7 +66,7 @@ task productionZip(type: Zip) {
     from copyNatives
     from jar.destinationDirectory
     into (project.name)
-    baseName = project.name
+    archiveBaseName = project.name
 }
 
 wrapper {
--- a/analysis/analyze-hotspots/build.gradle
+++ b/analysis/analyze-hotspots/build.gradle
@@ -1,6 +1,6 @@
 plugins {
     id 'application'
-    id 'org.openjfx.javafxplugin' version '0.0.5'
+    id 'org.openjfx.javafxplugin' version '0.0.8'
 }
 
 group = 'com.esri.samples'
--- a/analysis/analyze-hotspots/build.gradle
+++ b/analysis/analyze-hotspots/build.gradle
@@ -31,7 +31,7 @@ configurations {
 }
 
 dependencies {
-    compile "com.esri.arcgisruntime:arcgis-java:$arcgisVersion"
+    implementation "com.esri.arcgisruntime:arcgis-java:$arcgisVersion"
     natives "com.esri.arcgisruntime:arcgis-java-jnilibs:$arcgisVersion"
     natives "com.esri.arcgisruntime:arcgis-java-resources:$arcgisVersion"
 }
--- a/analysis/analyze-hotspots/build.gradle
+++ b/analysis/analyze-hotspots/build.gradle
@@ -22,7 +22,7 @@ repositories {
         url 'https://esri.bintray.com/arcgis'
     }
     maven {
-        url 'http://olympus.esri.com/artifactory/arcgisruntime-repo'
+        url 'https://olympus.esri.com/artifactory/arcgisruntime-repo'
     }
 }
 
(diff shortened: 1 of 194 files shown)


analysis/analyze-hotspots/build.gradle
```diff
@@ -1,6 +1,6 @@
 plugins {
     id 'application'
-    id 'org.openjfx.javafxplugin' version '0.0.5'
+    id 'org.openjfx.javafxplugin' version '0.0.8'
```
TADraeseke, line 3: Could we make a new variable in the version.gradle file to avoid having to update this again in the future?

Something like $javaFxVersion?

JonLavi: This is a good shout, but I think it'd require creating a `version.gradle` file for each sample. In the Java samples, each sample is it's own 'gradle project', so we wouldn't have a central place to change this variable from? Might be wrong though

JonLavi (approved):
Nice bit of gradle wizardry 👍 
If it wasn't done already, we should look at the Sample Viewer and see if anything there needs to be updated, just to make sure there's no problems when putting the samples and SV together as one big gradle project.

sclaridge:
> Nice bit of gradle wizardry 👍
> If it wasn't done already, we should look at the Sample Viewer and see if anything there needs to be updated, just to make sure there's no problems when putting the samples and SV together as one big gradle project.

@JonLavi The changes seem to be working with the SV as is. However the SV's `build.gradle` file also has deprecated features that could be removed.

analysis/analyze-hotspots/build.gradle
TADraeseke: Hmm. We only have one on Android which we share across several projects.

Just need to add a line to the gradle file to tell it where the version.gradle can be found with something like:

`apply from: '../version.gradle'` for a situation where the version.gradle file is up one level from the current gradle file location.

sclaridge: If I put the JavaFX version in a project level `version.gradle` file, then use a `defaultConfig` block inside the sample's `build.gradle` file, I get an error that the plugins should contain a string literal or a string with property replacement. 

I think the plugin version has to be [string literal](https://docs.gradle.org/current/userguide/plugins.html#sec:constrained_syntax). If we wanted to use a [variable](https://docs.gradle.org/current/userguide/plugins.html#sec:plugin_version_management) we would need to create a `pluginManagement` block in the `settings.gradle` file. I'm not sure how this would work when the samples are individual gradle projects, but I can look into it!

TADraeseke: Looks like you're on the right track and if the intention is to keep all samples on the same javafx version going forward I reckon it's probably worth the effort

Rachael-E (approved):
I had a quick scan through of all the files: looks good to me!

network_analysis/find-service-areas-for-multiple-facilities/build.gradle
```diff
@@ -67,9 +67,9 @@ task productionZip(type: Zip) {
         into "samples-data"
     }
     from copyNatives
-    from jar.destinationDir
+    from jar.destinationDirectory
```
Rachael-E, line 70: the gradle build check seems to be throwing a build failure here

```
Caused by: groovy.lang.MissingPropertyException: Could not get unknown property 'destinationDirectory' for task ':feature-collection-layer:jar' of type org.gradle.api.tasks.bundling.Jar.
	at org.gradle.internal.metaobject.AbstractDynamicObject.getMissingProperty(AbstractDynamicObject.java:83)
	at org.gradle.internal.metaobject.AbstractDynamicObject.getProperty(AbstractDynamicObject.java:61)
	at org.gradle.api.tasks.bundling.Jar_Decorated.getProperty(Unknown Source)
```

Rachael-E (changes requested):
looks like we need to find out what's causing these errors: let me know if you need any help.

network_analysis/find-service-areas-for-multiple-facilities/build.gradle
Rachael-E: @sclaridge I found the culprit! Thanks to Jon's thoughts on why things were going awry a while back, we need to upgrade the highest level gradle-wrapper.properties file to ensure it references 6.5.1 and not 5.0.

sclaridge: Thanks for noticing this, the build is successful now!

Rachael-E (approved):

PR #586, opened 2020-11-12
Readme Status Checker Scripts

jenmerritt:
### Description 

This PR is for the Readme status checker scripts that check the README.md and README.metada.json files are as per the requirements for Java Samples and are formatted correctly. The scripts will run as a GitHub Action when a Pull Request is made to either the `master` or `dev` branches, as set in the main.yml file on this branch.

The two scripts README_style_checker.py and metadata_style_checker.py can also be run locally on a sample directory.

It has been tested on my fork for PRs into the jen/test-status-checks branch (as set in the main.yml file on that branch).


.github/workflows/main.yml
```diff
@@ -0,0 +1,61 @@
+# The name of the job, which will be displayed for the status check in PR.
+name: README and metadata formatting
+
+# Controls when the action will run.
+# Below triggers the workflow on pull requests to the specified branches.
+on:
+  pull_request:
+    branches:
+      - jen/test-status-checks
```
JonLavi: This should be updated to `main` and we should add `dev`?

.github/workflows/main.yml
```diff
@@ -0,0 +1,61 @@
...
+
+# A workflow run is made up of one or more jobs that can run sequentially or
+# in parallel.
+jobs:
+  # This workflow contains a single job called "changes"
+  changes:
```
JonLavi: Can we change the name of the job to something more meaningful, e.g. `verify-readme`?

.github/workflows/main.yml
```diff
@@ -0,0 +1,61 @@
...
+      # https://github.com/actions/checkout
+      - name: Checkout the repository
+        uses: actions/checkout@v2
+
+      # Print out changed file paths for debugging purposes.
+      - name: Print changed file paths
+        run: |
+          echo 'Below shows a list of changed file paths'
+          echo '${{ steps.file_changes.outputs.files }}'
+      #    echo '${{ steps.file_changes.outputs.files_modified }}'
```
JonLavi, line 49: Printing out `files` and not `files_modified` here seems a bit counterintuitive, but the step `id: file_changes` might just pass the result into `files`? or does `files` include all the three commented-out categories? (modified/added/removed)?

.github/workflows/main.yml
```diff
@@ -0,0 +1,61 @@
...
+      # Run the style checks using the scripts in the './Scripts/CI/README_Metadata_StyleCheck' directory
+      # The scripts will run on any detected changed files from the 'file_changes' step
+      - name: Run style checks
+        uses: ./Scripts/CI/README_Metadata_StyleCheck
+        with:
+          FILE_PATHS: ${{ steps.file_changes.outputs.files }}
+
+      # Runs a single command using the runners shell
+      # - name: Run a script with Python
+      #   run: python ./Scripts/README-formatter.py
```
JonLavi: Just missing a newline here

.github/workflows/main.yml
```diff
@@ -0,0 +1,61 @@
...
+      # Print out changed file paths for debugging purposes.
+      - name: Print changed file paths
+        run: |
+          echo 'Below shows a list of changed file paths'
+          echo '${{ steps.file_changes.outputs.files }}'
+      #    echo '${{ steps.file_changes.outputs.files_modified }}'
+      #    echo '${{ steps.file_changes.outputs.files_added }}'
+      #    echo '${{ steps.file_changes.outputs.files_removed }}'
+
+      # Run the style checks using the scripts in the './Scripts/CI/README_Metadata_StyleCheck' directory
```
JonLavi: I think folder names in unix are traditionally written lowercase, so maybe we should change to `./scripts`

Scripts/CI/README_Metadata_StyleCheck/Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+ADD metadata_style_checker.py /metadata_style_checker.py
+ADD README_style_checker.py /README_style_checker.py
+# Install dependencies.
+RUN echo "**** Install Ruby and mdl ****" && \
+    apk add --update --no-cache ruby-full && \
+    gem install mdl --no-document && \
+    echo "**** Install Python ****" && \
+    apk add --no-cache python3 && \
+    if [ ! -e /usr/bin/python ]; then ln -sf python3 /usr/bin/python ; fi
+ENTRYPOINT ["python3", "/entry.py"]
```
JonLavi: newline missing here as well

Scripts/CI/README_Metadata_StyleCheck/README_style_checker.py
```diff
@@ -0,0 +1,435 @@
...
+    'ArcGIS Pro',
+    'GeoPackage',
+    'loadStatus',
+    'Integrated Windows Authentication',
+    'GeoElement',
+    'Network Link',
+    'Network Link Control',
+    'Open Street Map',
+    'OpenStreetMap',
+    'Play KML Tour',
```
JonLavi: Is this list for sample names only, specific sections of the readme, or any word inside a readme?
In case of the latter, we should add KML as well, but there's probably many more that we'd add along the way

Scripts/CI/README_Metadata_StyleCheck/README_style_checker.py
```diff
@@ -0,0 +1,435 @@
...
+
+
+# region Main wrapper functions
+def run_check(path: str, count: int) -> int:
+    checker = ReadmeStyleChecker(path)
+    # 1. Populate from README.
+    try:
+        checker.populate_from_readme()
+    except Exception as err:
+        count += 1
```
JonLavi: I can't tell if the actual number of exceptions is being used anywhere, so maybe it's possible to switch to `true`/`false`? 
I do recall there was an exit code 0/1 though, so maybe there's a technical reason for this as well.

Scripts/CI/README_Metadata_StyleCheck/README_style_checker.py
```diff
@@ -0,0 +1,435 @@
...
+# endregion
+
+
+if __name__ == '__main__':
+    try:
+        main()
+    except Exception as error:
+        print(f'{error}')
+        # Abort with failure if any exception occurs.
+        exit(1)
```
JonLavi: Missing newline here as well

Scripts/CI/README_Metadata_StyleCheck/action.yml
```diff
@@ -0,0 +1,20 @@
+# The name of the step just as a reference.
+name: "mdl and Python style checks"
+description: "This check will run several scripts to ensure the formatting."
+author: "tchen@esri.com"
```
JonLavi: Should we update the email here at all?

Scripts/CI/README_Metadata_StyleCheck/action.yml
```diff
@@ -0,0 +1,20 @@
...
+runs:
+  using: "docker"
+  image: "Dockerfile"
+  args:
+    - "-s"
+    - ${{ inputs.FILE_PATHS }}
+
+branding:
+  icon: "mic"
+  color: "purple"
```
JonLavi: Another missing newline, I won't comment on any other cases

Scripts/CI/README_Metadata_StyleCheck/entry.py
```diff
@@ -0,0 +1,143 @@
...
+    'network_analysis',
+    'ogc',
+    'portal',
+    'raster',
+    'scene',
+    'search',
+    'symbology',
+    'tiled_layers',
+    'utility_network'
+}
```
JonLavi: I remember we discussed a solution for this. At the moment this list is repeated in `README_style_checker.py`.
Perhaps we can store it in a text file `categories.txt` and read it in in both scripts, that way there's only one source of truth?

Scripts/CI/README_Metadata_StyleCheck/entry.py
```diff
@@ -0,0 +1,143 @@
...
+    return code
+
+def run_style_check(dirname: str):
+    print("**** README_style_checker ****")
+    code1 = sp.call(f'python3 /README_style_checker.py -s "{dirname}"', shell=True)
+    print("**** metadata_style_checker ****")
+    code2 = sp.call(f'python3 /metadata_style_checker.py -s "{dirname}"', shell=True)
+    return code1 + code2
+
+
```
JonLavi: Some inconsistent whitespace here, might be nice to tidy it up

Scripts/CI/README_Metadata_StyleCheck/entry.py
```diff
@@ -0,0 +1,143 @@
...
+    'tiled_layers',
+    'utility_network'
+}
+
+def run_mdl(readme_path: str):
+    print("**** markdownlint style checker (mdl) ****")
+    code = sp.call(f'mdl --style /style.rb "{readme_path}"', shell=True)
+    return code
+
+def run_style_check(dirname: str):
```
JonLavi: I think it would be helpful to add some comments here as to why we add together `code1 + code2` and what the return value is used for?

Scripts/CI/README_Metadata_StyleCheck/metadata_style_checker.py
```diff
@@ -0,0 +1,338 @@
...
+import re
+import json
+import typing
+import argparse
+import glob
+from pathlib import Path
+
+# region Global sets
+# A set of category folder names in current sample viewer.
+categories = {
```
JonLavi: Same here, if we could read this in from a file, it would save having to update the list in three places

Scripts/CI/README_Metadata_StyleCheck/metadata_style_checker.py
```diff
@@ -0,0 +1,338 @@
...
+}
+# endregion
+
+def sub_special_char(string: str) -> str:
+    """
+    Check and substitute if a string contains special characters.
+    :param string: The input string.
+    :return: A new string
+    """
+    # regex = re.compile('[@_!#$%^&*<>?|/\\}{~:]')
```
JonLavi: Should we remove the commented out regex? It looks identical to the one below anyway?

Scripts/CI/README_Metadata_StyleCheck/metadata_style_checker.py
```diff
@@ -0,0 +1,338 @@
...
+
+    # Specific check to ensure redirect_from is not [""] which breaks the website build
+    if json_data['redirect_from'] != [""]:
+        single_updater.redirect_from = json_data['redirect_from']
+    else:
+        raise Exception('Error in metadata: redirect_from cannot be [""]')
+
+    new = single_updater.flush_to_json_string()
+    original = json.dumps(json_data, indent=4, sort_keys=True)
+    if new != original:
```
JonLavi: I think if we use the output of this script https://devtopia.esri.com/runtime/common-samples/tree/master/tools/java
as the source of truth, and compare it against the README.metadata.json that is uploaded in the PR, then maybe it'd be easier to run the script and just overwrite whichever comes from the PR? It does feel a bit invasive, but there wouldn't be a way to pass this check if the metadata file isn't identical to the script output anyway? 
At this point we maybe need to consider what/where we need to make changes in case this script is stopping us from merging a metadata file as is?

JonLavi (commented):
Great work on this! Definitely a very useful thing to have, and as we fine-tune it it'll save us a lot of manual labour. 
I haven't gotten around to test it myself just yet, but adding a few comments for consideration.
If we could have a central file for the list of categories, and read that in at the right place, it will make dealing with those lists a bit easier.
Also noted some other points where it might be possible to do things differently (but not necessarily better - I'll leave that up to you to decide what is worht pursuing!)

Scripts/CI/README_Metadata_StyleCheck/README_style_checker.py
JonLavi: In hindsight, might be good to have this in a separate file as well, so that we have a central place to refer to the keywords

.github/workflows/main.yml
jenmerritt: Yes - helps if I push the change! Thanks

.github/workflows/main.yml
jenmerritt: yes - `files` is all three: modified / added / removed. I'll improve the commenting here.

Scripts/CI/README_Metadata_StyleCheck/README_style_checker.py
jenmerritt: This is specifically for the "tags" section and "title". It is for when it checks casing by comparing uppercase and lowercase so KML is ok as it will equal KML (but not kml) whereas e.g. GeoPackage will not equal GEOPACKAGE or geopackage. So in this list I've picked out all the titles that don't fit this rule (e.g. Play KML Tour) and any tags identified as not fitting this rule.

Other sections are left alone, e.g. Relevant API where they contain a lot / almost exclusively these proper nouns.

I've added to the commenting here too.

JonLavi: Ah now I follow! Thanks 😄

Scripts/CI/README_Metadata_StyleCheck/action.yml
jenmerritt: I'll run this by Ting and how we can handle across each SDK

Scripts/CI/README_Metadata_StyleCheck/entry.py
jenmerritt: Definitely, I've added this along with a few of the other improvements that we discussed to an issue in CS repo for future work to improve. I'll do the tidy ups from the code reviews and leave anything that will need more thought/time to that issue for another iteration.

Scripts/CI/README_Metadata_StyleCheck/metadata_style_checker.py
jenmerritt: Definitely sounds like a good extra step towards automating - I'll add these to the notes in the issue for the next iteration, thanks 👍

Scripts/CI/README_Metadata_StyleCheck/README_style_checker.py
jenmerritt: Yeah I think you're right. I think this is being output with the exception to show where the exception is e.g.
`1. map/display-map - Error title - README should contain title, description and image.` 
`1.` being the count. I guess because count is only 0/1 here where it compares `exception_count` to 0 it could be feasible to do true/false. But it maybe works with the exit code style? I'm hesitant to break this right now but could make a note in the CS issue if you think it'd be a good tidy up / improvement to explore?

.github/workflows/main.yml
jenmerritt: Good idea. Changed to `readme-metadata-stylecheck`

Scripts/CI/README_Metadata_StyleCheck/entry.py
jenmerritt: Added in 👍

Scripts/CI/README_Metadata_StyleCheck/README_style_checker.py
JonLavi: No worries, it's nothing major. Might be worth having a play with it, but if the errors are printed with a 'count' then that's ok too.

jenmerritt:
Thanks for the comments @JonLavi 
I made the changes initially on my testing branch in my fork so I could check everything still ran ok (also tested locally). Then merged into this branch - and corrected the branch names!!! So now the check is also visible in this PR, as expected.

scripts/ci/readme_metadata_style_check/README_style_checker.py
```diff
@@ -0,0 +1,435 @@
...
+        raise Exception('Tags are not sorted.')
+    return s
+
+
+def check_sentence_case(string: str) -> None:
+    """
+    Check if a sentence follows 'sentence case'. A few examples below.
+    Hello world! -> YES
+    I'm a good guy. -> YES
+    a man and a gun. -> NO
```
Rachael-E: minor comment but stood out for me...could we make these examples a bit more inclusive and less masculine/weaponised? :D some esri related suggestions below! 

```suggestion
    We're a mapping company. -> YES
    a map and a globe. -> NO
```

scripts/ci/readme_metadata_style_check/README_style_checker.py
```diff
@@ -0,0 +1,435 @@
...
+
+    essential_headers = {
+        'Use case',
+        'How to use the sample',
+        'How it works',
+        'Relevant API',
+        'Tags'
+    }
+
+    possible_headers = [
```
Rachael-E: there are some repeats in here from essential_headers: does that matter? just wondering if there's a way to avoid repetition.

Rachael-E (approved):
great work @jenmerritt. I haven't had the bandwidth to be able to test this out, but I reckon you and Simone have put it through the wringer a good amount of times now that that's a good enough test. When it goes "live" we'll be able to make tweaks as we find them. I've just put in two comments, one is more a question and doesn't block anything and the other is a suggestion for more inclusive language.

Only thought is that the new scripts directory will sit at the highest level of this repo, right? I thought it would live in the hidden .github/workflows folder, like the main.yml file and our existing gradle.yml. Is there a way the scripts directory could be hidden in the .github directory too?

.github/workflows/main.yml
```diff
@@ -0,0 +1,62 @@
+# The name of the job, which will be displayed for the status check in PR.
+name: README and metadata formatting
+
+# Controls when the action will run.
+# Below triggers the workflow on pull requests to the specified branches.
+on:
+  pull_request:
+    branches:
+      - master
```
sclaridge, line 9: Will we need to eventually change this when master changes to main?

sclaridge (approved):

.github/workflows/main.yml
jenmerritt: Yeah correct - it would just be a case of making that manual update. All that would happen if the branch was renamed and we didn't add it, is the scripts just would not run on that branch, so that should help us remember / it wouldn't cause any direct issues.

scripts/ci/readme_metadata_style_check/README_style_checker.py
jenmerritt: Thanks I'd skimmed the detail in these - and agree these should be updated. Adding these changes now.

scripts/ci/readme_metadata_style_check/README_style_checker.py
jenmerritt: Will add this as a refactor opportunity to the open issue, as it stands the lists are both being used and compared so will leave for now 👍

jenmerritt:
Just coming back to this having left it for the Citra project.

Thanks @Rachael-E for the suggestion to move the files, think that's a great shout and better usability. I've made that change.

@JonLavi I'll re-request your review incase there's anything else you wanted to run over.

Otherwise once everyone has approved I'll merge them in and @Rachael-E can see how useful they are in the coming months and identify any potential issues in practice!

JonLavi (approved):
LGTM!

PR #593, opened 2020-11-24
Add API key Gradle task to build.gradle files in samples

sclaridge:
This PR adds a Gradle task to all of the sample build.gradle files. The API key task creates a gradle.properties file in the .gradle folder if one doesn't already exist. The API key is set as a system property in the `run` task to allow the key to be used within in the samples.

Some things I would like some feedback on:
- Added some instructions to the `README` on API keys. Are the IDE instructions clear?
- For the terminal instructions, I followed the existing convention and added the step `./gradlew build`, which would create the  gradle.properties file. When testing out that command or `./gradlew run`, I got the error `Could not find or load main class org.gradle.wrapper.GradleWrapperMain`. To run `./gradlew run`, I had to run the `wrapper` task before hand, which also creates the gradle.properties file. The issue is documented [here](https://stackoverflow.com/questions/29805622/could-not-find-or-load-main-class-org-gradle-wrapper-gradlewrappermain). Should I replace `./gradlew build` with `gradle wrapper`?

@JonLavi


README.md
```diff
@@ -6,28 +6,36 @@ Browse the category directories to explore the samples. Each sample is an indivi
...
+5. Set your API key in the `gradle.properties` file located in the `/.gradle` folder within your home directory.
+6. Open the Gradle Tasks view with _Window > Show View > Other... > Gradle > Gradle Tasks_.
+7. In the Gradle Tasks view, double-click the `run` task under _{project_name} > application_ to run the app.
 
 ### Terminal
 
 1. `cd` into one of the sample project directories (not the category folder).
+3. Build the sample project with `./gradlew build`
+4. Set your API key in the `gradle.properties` file located in the `/.gradle` folder within your home directory.
 2. Run `./gradlew run` on Linux/Mac or `gradlew.bat run` on Windows to run the app.
```
JonLavi: The numbering here is out of order. 
Regarding your comment in the main issue, I was under the impression that we `gitignore` the wrapper file, but that doesn't actually seem to be the case:
https://github.com/Esri/arcgis-runtime-samples-java/tree/master/analysis/analyze-hotspots/gradle/wrapper
All samples have the wrapper and properties file on github, I think, so it seems like users _should_ have the wrapper and could just go ahead.. Is it possible that the wrappers are simply out of date? There's a change to one of the wrappers on the `display map` sample in this PR, which may be the only thing that was needed.

analysis/analyze-hotspots/build.gradle
```diff
@@ -36,6 +36,18 @@ dependencies {
     natives "com.esri.arcgisruntime:arcgis-java-resources:$arcgisVersion"
 }
 
+task apiKey {
+    description = "Creates a new gradle.properties file with an empty API key variable in the user home ./gradle folder if the file doesn't already exist."
```
JonLavi: I think another comma here would be good to break the sentence up:
```suggestion
    description = "Creates a new gradle.properties file with an empty API key variable in the user home ./gradle folder, if the file doesn't already exist."
```

analysis/analyze-hotspots/build.gradle
```diff
@@ -36,6 +36,18 @@ dependencies {
...
+    group = "build"
+    def apiKeyFile = new File("${System.properties.getProperty("user.home")}/.gradle/gradle.properties")
+    if (!apiKeyFile.exists()) {
+        print("Go to " + new URL("https://developers.arcgis.com") + " to get an API key.")
+        print(" Add your API key to ${System.properties.getProperty("user.home")}\\.gradle\\gradle.properties.")
+        String apiKeyFileContents = "apiKey = "
+        apiKeyFile.write(apiKeyFileContents)
```
JonLavi: Technically it's a properties file, and it could have other uses than just the Api key. Should we rename perhaps?

```suggestion
    def propertiesFile = new File("${System.properties.getProperty("user.home")}/.gradle/gradle.properties")
    if (!propertiesFile.exists()) {
        print("Go to " + new URL("https://developers.arcgis.com") + " to get an API key.")
        print(" Add your API key to ${System.properties.getProperty("user.home")}\\.gradle\\gradle.properties.")
        String propertiesFileContents = "apiKey = "
        propertiesFile.write(propertiesFileContents)
```

The last two lines could also possibly be written in one line?
`propertiesFile.write("apiKey = ")`?
I'm not 100% it works though, so worth trying it out.

analysis/analyze-hotspots/build.gradle
```diff
@@ -36,6 +36,18 @@ dependencies {
     natives "com.esri.arcgisruntime:arcgis-java-resources:$arcgisVersion"
 }
 
+task apiKey {
```
JonLavi: In the future this might be a bit more of a generic task, and mostly it creates the properties file, so if you choose to rename the variables inside, maybe we could call it:
```suggestion
task createProperties {
```

map/display-map/gradlew
```diff
@@ -1,5 +1,21 @@
 #!/usr/bin/env sh
 
+#
```
JonLavi: There's an update here to this wrapper file which I'm not sure was intentional?
Most of the wrappers are probably out of date, so might be good to have a separate PR that updates them all?

map/display-map/gradlew.bat
```diff
@@ -1,3 +1,19 @@
+@rem
```
JonLavi: Same here, not sure whether the update here was intentional (gradle updates these automatically when you run the wrapper task, I think)

map/display-map/gradlew
JonLavi: (mind the wrapper.jar file above as well for this sample, it's been changed too)

JonLavi (commented):
Nice work @sclaridge .
I'm still a bit unclear on whether we actually need to create the wrapper (it is included in the source code when pulling from github), or it's simply out of date? We did update the gradle version a while back so that might be a problem, with the build.gradle file specifying v 6.5.1 but the wrapper `jar` file and `gradlew` scripts still being on the old version (v5-ish?). This is just a wild guess though!

analysis/analyze-hotspots/build.gradle
sclaridge: It does work!

sclaridge:
@JonLavi I've added the instructions to create the Gradle wrapper to the Readme. I renamed the variables in the `apiKey` task, but I found if I renamed the task itself the build was failing if the `gradle.properties` file did not already exist. 

I've also added back the Gradle badge that was taken out when the reference to the bintray repository was removed.

README.md
```diff
@@ -1,33 +1,43 @@
...
+5. Set your API key in the `gradle.properties` file located in the `/.gradle` folder within your home directory.
+6. Open the Gradle Tasks view with _Window > Show View > Other... > Gradle > Gradle Tasks_.
+7. In the Gradle Tasks view, double-click the `run` task under _{project_name} > application_ to run the app.
 
 ### Terminal
 
 1. `cd` into one of the sample project directories (not the category folder).
+3. Run `gradle wrapper` to create the Gradle Wrapper
+4. Set your API key in the `gradle.properties` file located in the `/.gradle` folder within your home directory.
 2. Run `./gradlew run` on Linux/Mac or `gradlew.bat run` on Windows to run the app.
```
Rachael-E:
```suggestion
2. Run `gradle wrapper` to create the Gradle Wrapper.
3. Set your API key in the `gradle.properties` file located in the `/.gradle` folder within your home directory.
4. Run `./gradlew run` on Linux/Mac or `gradlew.bat run` on Windows to run the app.
```

analysis/analyze-hotspots/build.gradle
Rachael-E: this is a really good point: and I might be tempted to break it up even more or give it an even more precise name.

So either:
1. One task to `createGradleProperties` and another to `writeApiKey` that depends on the `createGradleProperties` task;
2. Rename the task `createGradlePropertiesAndWriteApiKey`

README.md
```diff
@@ -1,33 +1,43 @@
...
 ## Instructions
 
 ### IntelliJ IDEA
 
 1. Open IntelliJ IDEA and select _File > Open..._.
 2. Choose one of the sample project directories (not the category folder) and click _OK_.
 3. Select _File > Project Structure..._ and ensure that the Project SDK and language level are set to use Java 11.
-4. Open the Gradle view with _View > Tool Windows > Gradle_.
-5. In the Gradle view, double-click the `run` task under _Tasks > application_ to run the app.
+4. Set your API key in the `gradle.properties` file located in the `/.gradle` folder within your home directory.
```
Rachael-E, line 24: a small note here to say if the `gradle.properties` file doesn't already exist here, there is a gradle task in the samples which automatically generate a file for you would be helpful.

analysis/analyze-hotspots/build.gradle
```diff
@@ -47,6 +58,7 @@ task copyNatives(type: Copy) {
 }
 
 run {
+    systemProperty 'apiKey', apiKey
```
Rachael-E: can we add a comment above here to describe what is going on? just a brief description as to what the two different references are to apiKey here, a small comment would be useful to new users of the api key. could the two be slightly differently named? perhaps the property could be named "yourApiKey". I recognise that would have a knock on effect in Jen's PR, so perhaps with the task name being updated that will naturally provide a difference in naming.

sclaridge:
@Rachael-E @JonLavi I've renamed the task to `createGradlePropertiesAndWriteApiKey` instead of creating two separate tasks, as Gradle seems to default to writing `apiKey =` in the properties file if I attempt to have a separate task just to create the file. 

Strangely, I found that if I renamed the task apiKey to something like createProperties, if the gradle.properties did not exist, the build process failed with the error: Build file 'C:\Users\sim11117\samples\arcgis-runtime-samples-java\analysis\analyze-hotspots\build.gradle' line: 61
A problem occurred evaluating project ':analyze-hotspots'.
> Could not get unknown property 'apiKey' for task ':analyze-hotspots:run' of type org.gradle.api.tasks.JavaExec.

I've added a note in the Readme about the Gradle task, and added a comment to the run task explaining the apiKey as a system variable.

sclaridge:
@Rachael-E @JonLavi I've renamed the task to `createGradlePropertiesAndWriteApiKey` instead of creating two separate tasks, as Gradle seems to default to writing `apiKey =` in the properties file if I attempt to have a separate task just to create the file. 

Strangely, I found that if I renamed the task, if the gradle.properties did not exist, the build process failed with the error:  
```
> Could not get unknown property 'apiKey' for task ':analyze-hotspots:run' of type org.gradle.api.tasks.JavaExec.
```

I resolved this by adding the system property line in the run task to a `doFirst` block.   

I've added a note in the Readme about the Gradle task, and added a comment to the run task explaining the apiKey as a system variable.

Rachael-E (approved):
looks great @sclaridge !

JonLavi (approved):
Looks good to me!
I personally think `createGradlePropertiesAndWriteApiKey` is a hefty name, and any method/task name with an `And` in it just means it needs separating into a single responsibility 😁 But before we go in circles, let's just keep it as is

PR #597, opened 2020-11-30
Fix Hbox width in Display device location sample

sclaridge:
Another small PR to fix the size of the Hbox size to be correct on Linux.  This one escaped my attention, I should have included it in my previous PR! 

I've also changed the project name in the settings.gradle file.


jenmerritt (approved):

Rachael-E (approved):
thanks @sclaridge , nice catch!

PR #598, opened 2020-12-03
Fix to Take Screenshot Sample

jenmerritt:
Another small fix spotted in testing with the Basemap changes. Take Screenshot isn't running from the IDE due to the Launcher class being incorrectly referenced in the `build.gradle` file. This PR amends it.


JonLavi (approved):
Good find, thanks for fixing this!

Rachael-E (approved):
nice catch, can't believe this has been lurking for a while!

PR #612, opened 2020-12-10
Update data in Integrated Mesh Layer Sample

jenmerritt:
The following changes have been made to the Integrated Mesh Layer Sample (Scene category)

- Data replaced to a layer from Girona, Spain
- Camera position updated
- Scene basemap and Surface removed
- README updated
- PNG replaced to reflect the new data


JonLavi (approved):
Looks great @jenmerritt 👍

Rachael-E (approved):
@jenmerritt thanks looks good! just out of curiosity why was the Yosemite Lidar replaced with this new data?

jenmerritt:
> just out of curiosity why was the Yosemite Lidar replaced with this new data?

@Rachael-E it was to do with permissions on the original data source, so we had to change it.

PR #613, opened 2020-12-11
New Sample: creating symbols from a web style

jenmerritt:
### Description

This is my first shot at working on the first implementation of a sample - so any initial feedback, particularly on the README before I create the Design PR is welcomed. I'll then make any changes from that process before it is "officially reviewed". 

The issue within CS (2331) is called "Save feature layer with symbol style". However the requirements changed and the actual requirements became just following the existing JavaScript sample to create symbols on the LA County Points of Interest service as a feature layer. So functionality wise it's fairly straightforward and clear, it's more the write up where I may need some tweaks.

So that also means I've come up with the current name - "Create Symbols from Web Styles" - so I welcome feedback / ideas on that too, and then I'm sure it'll also then get debated within the samples team. I had thought maybe "Create symbol style from web style" could work but it felt like there were too many styles?!

JS sample: https://developers.arcgis.com/javascript/latest/sample-code/webstylesymbol-2d/index.html
Web Style being used: https://developers.arcgis.com/javascript/latest/guide/esri-web-style-symbols-2d/

The key point of the sample is demonstrating that the constructor for a `SymbolStyle` can now be from a portal (AGO / Enterprise) where a style file is hosted. I can send any internal links from my reading if you want any more links to info.

Thanks!

Additional note - I'll probably extract some of the UI styling out into a CSS file when things are sured up.


symbology/create-symbols-from-web-styles/README.md
```diff
@@ -0,0 +1,46 @@
+# Create symbols from web styles
```
JonLavi: I think for the name of the sample, it really depends on what the sample is trying to show. Is the main thing to show how to create a `SymbolStyle` class, or is the `Symbol` the end goal (rather than just what we use to show it on the map?)
We have a whole bunch of samples saying `simple X symbol`, so perhaps the point of this sample is to show how to make specifically a `SymbolStyle` from a `Web Style`? In which case I think `Create symbol style from web style` is a good name

symbology/create-symbols-from-web-styles/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+Create symbols from a style file hosted on a portal.
```
JonLavi: If we agree on the purpose of this sample, we might have to update this to something along the lines of 
`Create a symbol style from a style file hosted on a portal`
In case the actual displaying of the `Symbol` objects to the map isn't the main subject here?

symbology/create-symbols-from-web-styles/README.md
```diff
@@ -0,0 +1,46 @@
...
+Style files hosted on an ArcGIS Online or Enterprise portal are known as web styles. They can be used to style symbols on a feature layer or graphic overlay.
+
+## How to use the sample
+
+The sample displays a map with a set of symbols that represent the categories of the features within the dataset. Pan and zoom on the map and view the legend to explore the appearance and names of the different symbols from the selected symbol style.
+
+## How it works
+
+1. Create a `FeatureLayer` and add it to the map.
+2. Create a `UniqueValueRenderer` and set it to the feature layer.
```
JonLavi: Might be up to the decision of the samples team, but perhaps the whole instructions on the `UniqueValueRenderer` are slightly distracting? Although I'm not sure if they are an integral part to creating `SymbolStyle`s from web styles?

symbology/create-symbols-from-web-styles/src/main/java/com/esri/samples/create_symbols_from_web_styles/CreateSymbolsFromWebStylesSample.java
```diff
@@ -0,0 +1,316 @@
...
+        "http://services.arcgis.com/V6ZHFr6zdgNZuVG0/arcgis/rest/services/LA_County_Points_of_Interest/FeatureServer/0"));
+
+      // add the feature layer to the map's operational layers
+      map.getOperationalLayers().add(featureLayer);
+
+      // display an error if the feature layer fails to load
+      featureLayer.addDoneLoadingListener(() -> {
+        if (featureLayer.getLoadStatus() == LoadStatus.FAILED_TO_LOAD) {
+          new Alert(Alert.AlertType.ERROR, "Feature layer failed to load. Details: \n"
+            + featureLayer.getLoadError().getCause().getMessage()).show();
```
JonLavi: We've not previously been putting out the contents of the error message (after all, this would be for the end-user who probably won't benefit much from the technical details).
However if this is now the established pattern going forward, we should make an issue t update old samples with the same style of error messages

symbology/create-symbols-from-web-styles/src/main/java/com/esri/samples/create_symbols_from_web_styles/CreateSymbolsFromWebStylesSample.java
```diff
@@ -0,0 +1,316 @@
...
+        break;
+      case "post-office":
+        categories.addAll(Arrays.asList("DHL Locations", "Federal Express Locations"));
+        break;
+      case "school":
+        categories.addAll(Arrays.asList("Public High Schools", "Public Elementary Schools", "Private and Charter Schools"));
+        break;
+      case "trail":
+        categories.add("Trails");
+        break;
```
JonLavi: We should add a `default` case for this switch statement - is there any symbol that can be used as a catch-all for data that doesn't have a category that's in this list?

symbology/create-symbols-from-web-styles/src/main/java/com/esri/samples/create_symbols_from_web_styles/CreateSymbolsFromWebStylesSample.java
```diff
@@ -0,0 +1,316 @@
...
+        }
+      });
+
+      // setup the UI for the legend
+      setupLegend();
+
+      // create a list of the required symbol names from the web style
+      ArrayList<String> symbolNames = new ArrayList<>();
+      symbolNames.addAll(Arrays.asList("atm", "beach", "campground", "city-hall", "hospital", "library", "park",
+        "place-of-worship", "police-station", "post-office", "school", "trail"));
```
JonLavi: It's possible to instantiate the ArrayList and populate it with values in one statement:
```suggestion
      ArrayList<String> symbolNames = new ArrayList<>(Arrays.asList("atm", "beach", "campground", "city-hall", 
              "hospital", "library", "park", "place-of-worship", "police-station", "post-office", "school", "trail"));
```

symbology/create-symbols-from-web-styles/src/main/java/com/esri/samples/create_symbols_from_web_styles/CreateSymbolsFromWebStylesSample.java
```diff
@@ -0,0 +1,316 @@
...
+            // on any error, display the stack trace.
+            e.printStackTrace();
+          }
+        });
+      }
+
+      // add a map scale changed listener on the map view to control the symbol sizes at different scales
+      mapView.addMapScaleChangedListener(mapScaleChangedEvent -> {
+        featureLayer.setScaleSymbols(!(mapView.getMapScale() < 80000));
+      });
```
JonLavi: There's two things we can improve here: since it's a single line within the lambda, we can get rid of the curly braces and the `;` at the end of that single line. Also, instead of doing `!(x < y)`, we can just do (x >= y). I means one less set of parentheses, and the `!` is a bit easier to miss than seeing `>=` 
```suggestion
      mapView.addMapScaleChangedListener(mapScaleChangedEvent ->
              featureLayer.setScaleSymbols(mapView.getMapScale() >= 80000));
```

symbology/create-symbols-from-web-styles/src/main/java/com/esri/samples/create_symbols_from_web_styles/CreateSymbolsFromWebStylesSample.java
```diff
@@ -0,0 +1,316 @@
...
+
+      // set an initial reference scale on the map for controlling symbol size
+      map.setReferenceScale(100000);
+
+      // create a map view and set the map to it
+      mapView = new MapView();
+      mapView.setMap(map);
+
+      // set a viewpoint on the map view
+      mapView.setViewpoint(new Viewpoint(34.28301, -118.44186, 7000));
```
JonLavi: I personally think it'd be more elegant to set the viewpoint programatically instead of hard-code it:
`mapView.setViewpoint(new Viewpoint(featureLayer.getFullExtent()));`
It means we don't have to worry about updating this if the service changes somehow.
The big downside is, that this won't work unless the `FeatureLayer` isn't loaded, so this line would have to go inside a `addDoneLoadingListener`. So up to you and @Rachael-E to decide if this is worth doing

symbology/create-symbols-from-web-styles/src/main/java/com/esri/samples/create_symbols_from_web_styles/CreateSymbolsFromWebStylesSample.java
```diff
@@ -0,0 +1,316 @@
...
+      });
+
+      // add the map view and grid pane to the stack pane
+      stackPane.getChildren().addAll(mapView, gridPane);
+      StackPane.setAlignment(gridPane, Pos.TOP_LEFT);
+      StackPane.setMargin(gridPane, new Insets(10, 0, 0, 10));
+
+    } catch (Exception e) {
+      // on any error, display the stack trace.
+      e.printStackTrace();
```
JonLavi: We should add a clearer error message here (plus `e.getCause().getMessage()`, if that's the way forward

symbology/create-symbols-from-web-styles/src/main/java/com/esri/samples/create_symbols_from_web_styles/CreateSymbolsFromWebStylesSample.java
```diff
@@ -0,0 +1,316 @@
...
+    }
+
+  }
+
+  /**
+   * Creates a legend on the UI for the symbol styles.
+   *
+   */
```
JonLavi: There's an extra line here that can go
```suggestion
  /**
   * Creates a legend on the UI for the symbol styles.
   */
```

symbology/create-symbols-from-web-styles/src/main/java/com/esri/samples/create_symbols_from_web_styles/CreateSymbolsFromWebStylesSample.java
```diff
@@ -0,0 +1,316 @@
...
+
+      // setup the UI for the legend
+      setupLegend();
+
+      // create a list of the required symbol names from the web style
+      ArrayList<String> symbolNames = new ArrayList<>();
+      symbolNames.addAll(Arrays.asList("atm", "beach", "campground", "city-hall", "hospital", "library", "park",
+        "place-of-worship", "police-station", "post-office", "school", "trail"));
+
+      for (String symbolName : symbolNames) {
```
JonLavi: Could we add a comment about what this for loop's purpose is? Something like
`// build unique values for the renderer and construct a symbol for each feature`
?

symbology/create-symbols-from-web-styles/src/main/java/com/esri/samples/create_symbols_from_web_styles/CreateSymbolsFromWebStylesSample.java
```diff
@@ -0,0 +1,316 @@
...
+        break;
+      case "trail":
+        categories.add("Trails");
+        break;
+    }
+    return categories;
+  }
+
+  /**
+   * Returns an imageView populated with a symbol.
```
JonLavi: We should probably either use plain-english, or the full class names for these comments? So either
```suggestion
   * Returns an ImageView populated with a Symbol.
```
or:
```suggestion
   * Returns an image view populated with a symbol.
```

symbology/create-symbols-from-web-styles/src/main/java/com/esri/samples/create_symbols_from_web_styles/CreateSymbolsFromWebStylesSample.java
```diff
@@ -0,0 +1,316 @@
...
+      ArrayList<String> symbolNames = new ArrayList<>();
+      symbolNames.addAll(Arrays.asList("atm", "beach", "campground", "city-hall", "hospital", "library", "park",
+        "place-of-worship", "police-station", "post-office", "school", "trail"));
+
+      for (String symbolName : symbolNames) {
+
+        // search for each symbol in the symbol style
+        List<String> searchKey = new ArrayList<>();
+        searchKey.add(symbolName);
+        ListenableFuture<Symbol> searchResult = symbolStyle.getSymbolAsync(searchKey);
```
JonLavi: I think we can avoid creating a new `List<String` and populating it on separate lines, by just passing `Collections.singeltonList(item)` directly into the method call:
```suggestion
        ListenableFuture<Symbol> searchResult = symbolStyle.getSymbolAsync(Collections.singletonList(symbolName));
```

symbology/create-symbols-from-web-styles/src/main/java/com/esri/samples/create_symbols_from_web_styles/CreateSymbolsFromWebStylesSample.java
```diff
@@ -0,0 +1,316 @@
...
+            // get a list of all categories to be mapped to the symbol
+            List<String> categories = mapSymbolNameToField(symbolName);
+
+            for (String category : categories) {
+
+              // create a unique value for each category
+              List<Object> categoryValue = new ArrayList<>();
+              categoryValue.add(category);
+              UniqueValueRenderer.UniqueValue uniqueValue = new UniqueValueRenderer.UniqueValue(
+                "", symbolName, symbol, categoryValue);
```
JonLavi: I haven't been able to test this, but we might be able to do the same trick above and simply pass a `Collections.singletonList(item)` into the method call. 
```suggestion
              UniqueValueRenderer.UniqueValue uniqueValue = new UniqueValueRenderer.UniqueValue(
                "", symbolName, symbol, Collections.singletonList(category));
```

It's possibly a bit less easy to read though, so up to you whether you choose to do it this way

symbology/create-symbols-from-web-styles/src/main/java/com/esri/samples/create_symbols_from_web_styles/CreateSymbolsFromWebStylesSample.java
```diff
@@ -0,0 +1,316 @@
...
+
+              // add each unique value to the unique value renderer
+              uniqueValueRenderer.getUniqueValues().add(uniqueValue);
+            }
+
+            // create and add an image view and a label for the symbol to the legend on the UI
+            ImageView imageView = createImageView(symbol);
+            Label gridPaneLabel = new Label(symbolName);
+            gridPane.add(imageView, 0, symbolNames.indexOf(symbolName) + 2);
+            gridPane.add(gridPaneLabel, 1, symbolNames.indexOf(symbolName) + 2);
```
JonLavi: Maybe a small comment to say here that the image views and labels are added according to the item's position in the list of `symbol names`? Particularly the +2 is a little puzzling, I'm guessing it's go to do with the headers for this grid pane. Might be easier to extract a variable `Int something = symbolNames.indexOf(symbolName) + 2` 
Alternatively maybe have 2 grid panes, one for the table headers `(Label symbolTitle = new Label("Symbol");` and `Label nameTitle = new Label("Name");`, and then one for the 'data'? We can discuss over a call, might be easier to explain

JonLavi (commented):
Looks very neat @jenmerritt ! 
I've made just a few comments on some small refactoring that might be possible. It's a matter of taste though, so don't feel like you have to sacrifice legibility for any of it. Pick and choose as you see fit.
For the main structure of the code, I think it looks good, but perhaps it would be useful to have a call and you could walk me through it so I can better understand the intentions?

symbology/create-symbols-from-web-styles/src/main/java/com/esri/samples/create_symbols_from_web_styles/CreateSymbolsFromWebStylesSample.java
jenmerritt: Looking at this more, I only want the specific symbol names and categories in the switch statement to be used so it is very selective. I want to ignore all other categories (as there is so much data in the dataset). So as long as that is justification for not having a default, I think I'd want to leave it without?

symbology/create-symbols-from-web-styles/src/main/java/com/esri/samples/create_symbols_from_web_styles/CreateSymbolsFromWebStylesSample.java
jenmerritt: When I set to the feature layer extent it is too far zoomed out. The viewpoint set manually is zoomed in on a specific area with a nice spread of symbols. So if this is an ok justification for that, it would be good to leave it. Unless there is a way to be more selective about where on the feature layer you can create the viewpoint? Will leave this comment open for Rachael's review!

symbology/create-symbols-from-web-styles/src/main/java/com/esri/samples/create_symbols_from_web_styles/CreateSymbolsFromWebStylesSample.java
jenmerritt: I've separated this into 2 `GridPanes`.  I tried to move some styling into a css file, but of course the sample viewer doesn't read the individual sample's style file (until we are able to put some time into that) so the unique styles won't carry over. So for now, I've left all the required styling where it is.

symbology/create-symbols-from-web-styles/src/main/java/com/esri/samples/create_symbols_from_web_styles/CreateSymbolsFromWebStylesSample.java
JonLavi: Yep that seems fair! Just leave it as is in this case.

symbology/create-symbols-from-web-styles/src/main/java/com/esri/samples/create_symbols_from_web_styles/CreateSymbolsFromWebStylesSample.java
JonLavi: Looks much nicer with the two grid panes 👍 
As for the CSS, I wouldn't worry about that too much. As long as it works in the SV then we are good to go.

symbology/create-symbols-from-web-styles/src/main/java/com/esri/samples/create_symbols_from_web_styles/CreateSymbolsFromWebStylesSample.java
JonLavi: I see your point, and that would make sense to leave it hard-coded then 👍

symbology/create-symbols-from-web-styles/src/main/java/com/esri/samples/create_symbols_from_web_styles/CreateSymbolsFromWebStylesSample.java
```diff
@@ -0,0 +1,319 @@
...
+      UniqueValueRenderer uniqueValueRenderer = new UniqueValueRenderer();
+
+      // add the name of a field from the feature layer data that symbols will be mapped to
+      uniqueValueRenderer.getFieldNames().add("cat2");
+
+      // set the unique value renderer on the feature layer
+      featureLayer.setRenderer(uniqueValueRenderer);
+
+      // create a symbol style from a web style
+      symbolStyle = new SymbolStyle("Esri2DPointSymbolsStyle", null);
```
Rachael-E: I think an expansion on the comment here to explain what null defaults to would be good here to help the user understand the constructor better:

```suggestion
      // create a symbol style from a web style. ArcGIS Online is used as the default portal when null is passed as the portal parameter
      symbolStyle = new SymbolStyle("Esri2DPointSymbolsStyle", null);
```

symbology/create-symbols-from-web-styles/src/main/java/com/esri/samples/create_symbols_from_web_styles/CreateSymbolsFromWebStylesSample.java
```diff
@@ -0,0 +1,319 @@
...
+
+      // create a feature layer from a service
+      FeatureLayer featureLayer = new FeatureLayer(new ServiceFeatureTable(
+        "http://services.arcgis.com/V6ZHFr6zdgNZuVG0/arcgis/rest/services/LA_County_Points_of_Interest/FeatureServer/0"));
+
+      // add the feature layer to the map's operational layers
+      map.getOperationalLayers().add(featureLayer);
+
+      // display an error if the feature layer fails to load
+      featureLayer.addDoneLoadingListener(() -> {
```
Rachael-E: I'm not sure how I feel about these add done loading listeners for if something has failed to load. I seem to remember this was something that was brought in a few releases ago to show the loadable pattern consistenly in our samples though.  It just seems like they get in the way a bit of the flow of the sample. If they have to stay due to a previous decision, could we separate out a method that takes a `Loadable` and call that method in place of this, and the later SymbolStyle loading check?

symbology/create-symbols-from-web-styles/src/main/java/com/esri/samples/create_symbols_from_web_styles/CreateSymbolsFromWebStylesSample.java
```diff
@@ -0,0 +1,319 @@
...
+        if (featureLayer.getLoadStatus() == LoadStatus.FAILED_TO_LOAD) {
+          new Alert(Alert.AlertType.ERROR, "Feature layer failed to load. Details: \n"
+            + featureLayer.getLoadError().getCause().getMessage()).show();
+        }
+      });
+
+      // create a unique value renderer
+      UniqueValueRenderer uniqueValueRenderer = new UniqueValueRenderer();
+
+      // add the name of a field from the feature layer data that symbols will be mapped to
```
Rachael-E: this is a really helpful comment! sometimes the strings passed into these parameters can be quite obscure as to where they come from.

symbology/create-symbols-from-web-styles/README.md
Rachael-E: I'd be happy with "Create symbol styles from web styles" if we want to be more explicit, probs a good one to bounce around the samples team too.

symbology/create-symbols-from-web-styles/src/main/java/com/esri/samples/create_symbols_from_web_styles/CreateSymbolsFromWebStylesSample.java
Rachael-E: since this is at the end of the larger try/catch block for the `start` method, I'm happy with leaving this as generic and for the exception to be printed within the console log for the developer to debug. 

Where we are being more explicit is for showing dialog pop up messages, where an end user who isn't a developer may wonder why something didn't load (e.g. feature layer not loading due to no internet connection).

Rachael-E (approved):
hey @jenmerritt really nice sample design here! No major comments here, just a few tidy ups I think are all that are needed. The sample flows well and is really well commented. I wondered initially why we weren't programmatically accessing a list of the web styles but I see from the JavaScript implementation we only want a subset and not the full list of 64 styles or so! So that's fine having the values hardcoded into a list to then map to later.

symbology/create-symbols-from-web-styles/src/main/java/com/esri/samples/create_symbols_from_web_styles/CreateSymbolsFromWebStylesSample.java
jenmerritt: Agreed, have updated

symbology/create-symbols-from-web-styles/src/main/java/com/esri/samples/create_symbols_from_web_styles/CreateSymbolsFromWebStylesSample.java
jenmerritt: @Rachael-E Yeah I get your point - we don't have to have them, I guess it's the balance with making the sample flow better vs silently failing and showing how to avoid that - but I guess if we know the data works in our sample we can leave them out... Guess there are other samples (or tutorials etc) for learning loading patterns and that's not necessarily the point of this sample? Shall I just remove these?

jenmerritt:
@JonLavi this design is now merged in CS 🎉  Could you re-check this PR following your initial comments? Thanks!

JonLavi (approved):
Nice work @jenmerritt ! 
Nothing to add from my side here.

PR #614, opened 2020-12-14
New sample: Apply mosiac rule to rasters

sclaridge:
This sample was previously waiting for two issues to be fixed for the `LOCK_RASTER` method. I've set the target branch to dev so that the lock raster method will work correctly when the sample is publicly released. 

Currently, the `mapView` is instantiated and the `DrawStatusChangedListener` set before the map is created so that `COMPLETED` draw status events will be generated and the progress indicator can be used, as advised by the Java team. 

@jenmerritt I have added the new SS7 URL for the image raster service. 

Thanks!

@JonLavi


raster/apply-mosaic-rule-to-rasters/README.md
```diff
@@ -0,0 +1,43 @@
+# Apply mosaic rule to rasters
```
JonLavi, line 1: I think it would make sense to keep `raster` as singular throughout, unless there's a mechanism to apply a rule to a whole group of rasters in bulk? 
However if the design for this is already finalised, just disregard this comment 😁 
```suggestion
# Apply mosaic rule to raster
```

raster/apply-mosaic-rule-to-rasters/src/main/java/com/esri/samples/apply_mosaic_rule_to_rasters/ApplyMosaicRuleToRastersSample.java
```diff
@@ -0,0 +1,212 @@
...
+      String yourAPIKey = System.getProperty("apiKey");
+      ArcGISRuntimeEnvironment.setApiKey(yourAPIKey);
+
+      // create a label
+      Label mosaicRuleLabel = new Label("Choose a mosaic rule: ");
+      mosaicRuleLabel.setTextFill(Color.WHITE);
+
+      // create a combo box
+      ComboBox<String> comboBox = new ComboBox<>();
+      comboBox.setMaxWidth(Double.MAX_VALUE);
```
JonLavi: Maybe we can be a bit more specific and name this `mosaicRuleComboBox`?

raster/apply-mosaic-rule-to-rasters/src/main/java/com/esri/samples/apply_mosaic_rule_to_rasters/ApplyMosaicRuleToRastersSample.java
```diff
@@ -0,0 +1,212 @@
...
+
+      // create a map with the light gray basemap style
+      ArcGISMap map = new ArcGISMap(BasemapStyle.ARCGIS_LIGHT_GRAY);
+
+      // set the map to the map view
+      mapView.setMap(map);
+
+      // create a raster layer from the image service raster
+      String imageServiceURL = "https://sampleserver7.arcgisonline.com/server/rest/services/amberg_germany/ImageServer";
+      ImageServiceRaster imageServiceRaster = new ImageServiceRaster(imageServiceURL);
```
JonLavi: Should we just put the URL right in the constructor to avoid making a `String` variable only used once?
```suggestion
      ImageServiceRaster imageServiceRaster =
       new ImageServiceRaster("https://sampleserver7.arcgisonline.com/server/rest/services/amberg_germany/ImageServer");
```

raster/apply-mosaic-rule-to-rasters/src/main/java/com/esri/samples/apply_mosaic_rule_to_rasters/ApplyMosaicRuleToRastersSample.java
```diff
@@ -0,0 +1,212 @@
...
+          mapView.setViewpoint(new Viewpoint(imageServiceRaster.getServiceInfo().getFullExtent().getCenter(), 25000.0));
+
+          // enable UI interaction once the raster layer has loaded
+          controlsVBox.setVisible(true);
+
+          // add the mosaic rules to the combo box
+          comboBox.getItems().addAll("Default", "Northwest", "Center", "By attribute", "Lock raster");
+
+          // set the default combo box value
+          comboBox.setValue("Default");
```
JonLavi: I've never actually seen `.setValue`. Should maybe double check that it's doing the same as `          comboBox.getSelectionModel().select(0);`?

JonLavi: I think we can do these things independently from waiting for the `RasterLayer` to load, and it might make sense to move them closer to line 152 (where we add the listener: `comboBox.getSelectionModel().selectedItemProperty().addListener`)? That way it's clearer what these added values are going to be used for

raster/apply-mosaic-rule-to-rasters/src/main/java/com/esri/samples/apply_mosaic_rule_to_rasters/ApplyMosaicRuleToRastersSample.java
```diff
@@ -0,0 +1,212 @@
...
+
+      // set up the control panel UI
+      VBox controlsVBox = new VBox(6);
+      controlsVBox.setBackground(new Background(new BackgroundFill(Paint.valueOf("rgba(0, 0, 0, 0.3)"),
+        CornerRadii.EMPTY, Insets.EMPTY)));
+      controlsVBox.setPadding(new Insets(10.0));
+      controlsVBox.setMaxSize(210, 50);
+      controlsVBox.setVisible(false);
+      // add the label and combo box to the control panel
+      controlsVBox.getChildren().addAll(mosaicRuleLabel, comboBox);
```
JonLavi: In @jenmerritt new sample, there is a method `setupUI` (or similarly named), that hides all this javaFX stuff a bit further to the bottom of the file. Might be worth considering this as a new pattern? Perhaps a discussion with @Rachael-E and Mark on this would be good? After all, it just distracts a bit from the ArcGIS runtime goodness

raster/apply-mosaic-rule-to-rasters/src/main/java/com/esri/samples/apply_mosaic_rule_to_rasters/ApplyMosaicRuleToRastersSample.java
```diff
@@ -0,0 +1,212 @@
...
+          comboBox.setValue("Default");
+
+          // create a mosaic rule for the image service raster if one does not already exist
+          if (imageServiceRaster.getMosaicRule() == null) {
+            mosaicRule = new MosaicRule();
+            imageServiceRaster.setMosaicRule(mosaicRule);
+          }
+        } else {
+          // show alert if raster layer fails to load.
+          new Alert(Alert.AlertType.ERROR, "Error loading raster layer.").show();
```
JonLavi, line 119: Double check with @Rachael-E if we want to show the contents of the exception/load error as a standard for samples now?

raster/apply-mosaic-rule-to-rasters/src/main/java/com/esri/samples/apply_mosaic_rule_to_rasters/ApplyMosaicRuleToRastersSample.java
```diff
@@ -0,0 +1,212 @@
...
+        else {
+          progressIndicator.setVisible(false);
+        }
+      });
+
+      // create a map with the light gray basemap style
+      ArcGISMap map = new ArcGISMap(BasemapStyle.ARCGIS_LIGHT_GRAY);
+
+      // set the map to the map view
+      mapView.setMap(map);
```
JonLavi: I think the established order for this after the recent PRs was something along the lines of:
(not 100% about the comments though, @jenmerritt could advise?)
```suggestion
      // create a map with the light gray basemap style
      ArcGISMap map = new ArcGISMap(BasemapStyle.ARCGIS_LIGHT_GRAY);
	
	  // create a map view and set its map
      mapView = new MapView();
      mapView.setMap(map);
```

We'd then need to move `mapView.addDrawStatusChangedListener` down below this

raster/apply-mosaic-rule-to-rasters/build.gradle
```diff
@@ -0,0 +1,89 @@
...
+        from(zipTree(it))
+    }
+    // store native libraries in a common location shared with other samples
+    into "${System.properties.getProperty("user.home")}/.arcgis/$arcgisVersion"
+}
+
+run {
+    doFirst {
+        // sets the API key from the gradle.properties file as a Java system property
+        systemProperty 'apiKey', apiKey
```
JonLavi, line 63: I've just noticed this doesn't run the task `createGradlePropertiesAndWriteApiKey`, meaning it just kind of fails weirdly with a mention of `apiKey`, if it isn't set inside the `gradle.properties` already.
Should `run` also do that task first, or `dependsOn` it?
I know we've recently been implementing this, so just wondering if/where any users would find documentation when they get stuck?

raster/apply-mosaic-rule-to-rasters/src/main/java/com/esri/samples/apply_mosaic_rule_to_rasters/ApplyMosaicRuleToRastersSample.java
```diff
@@ -0,0 +1,212 @@
...
+      // add draw status listener to the map view
+      mapView.addDrawStatusChangedListener (drawStatusChangedEvent -> {
+        // show progress indicator while map view is drawing
+        if (drawStatusChangedEvent.getDrawStatus() == DrawStatus.IN_PROGRESS) {
+          progressIndicator.setVisible(true);
+        }
+        else {
+          progressIndicator.setVisible(false);
+        }
+      });
```
JonLavi: We can simplify this a fair bit by passing the result of the comparison `==` straight into `.setVisible`.
Up to you and @Rachael-E to decide if this is a sacrifice of legibility though

Other than that, I think the inside comment does explain the whole block, so we could get rid of `// add draw status lisener...`
```suggestion
      // show progress indicator while map view is drawing
      mapView.addDrawStatusChangedListener(drawStatusChangedEvent -> 
        progressIndicator.setVisible(drawStatusChangedEvent.getDrawStatus() == DrawStatus.IN_PROGRESS)
      );
```

raster/apply-mosaic-rule-to-rasters/src/main/java/com/esri/samples/apply_mosaic_rule_to_rasters/ApplyMosaicRuleToRastersSample.java
```diff
@@ -0,0 +1,212 @@
...
+
+      // create a raster layer from the image service raster
+      String imageServiceURL = "https://sampleserver7.arcgisonline.com/server/rest/services/amberg_germany/ImageServer";
+      ImageServiceRaster imageServiceRaster = new ImageServiceRaster(imageServiceURL);
+      rasterLayer = new RasterLayer(imageServiceRaster);
+
+      // add raster layer as an operational layer to the map
+      map.getOperationalLayers().add(rasterLayer);
+
+      // listen for the raster layer to finish loading
```
JonLavi: I think `wait` is a bit more plain english than the technical term `listen`
```suggestion
      // wait for the raster layer to finish loading
```

raster/apply-mosaic-rule-to-rasters/src/main/java/com/esri/samples/apply_mosaic_rule_to_rasters/ApplyMosaicRuleToRastersSample.java
```diff
@@ -0,0 +1,212 @@
...
+
+      // set the map to the map view
+      mapView.setMap(map);
+
+      // create a raster layer from the image service raster
+      String imageServiceURL = "https://sampleserver7.arcgisonline.com/server/rest/services/amberg_germany/ImageServer";
+      ImageServiceRaster imageServiceRaster = new ImageServiceRaster(imageServiceURL);
+      rasterLayer = new RasterLayer(imageServiceRaster);
+
+      // add raster layer as an operational layer to the map
```
JonLavi: Let's add an `a` to make it more plain english?
```suggestion
      // add a raster layer as an operational layer to the map
```

raster/apply-mosaic-rule-to-rasters/src/main/java/com/esri/samples/apply_mosaic_rule_to_rasters/ApplyMosaicRuleToRastersSample.java
```diff
@@ -0,0 +1,212 @@
...
+      map.getOperationalLayers().add(rasterLayer);
+
+      // listen for the raster layer to finish loading
+      rasterLayer.addDoneLoadingListener(() -> {
+        if (rasterLayer.getLoadStatus() == LoadStatus.LOADED) {
+          // when loaded, set map view's viewpoint to the image service raster's center
```
JonLavi: Some more possibility for plain english here as well?
```suggestion
          // when loaded, set the map view's viewpoint to the image service raster's center
```

raster/apply-mosaic-rule-to-rasters/src/main/java/com/esri/samples/apply_mosaic_rule_to_rasters/ApplyMosaicRuleToRastersSample.java
```diff
@@ -0,0 +1,212 @@
...
+        } else {
+          // show alert if raster layer fails to load.
+          new Alert(Alert.AlertType.ERROR, "Error loading raster layer.").show();
+        }
+      });
+
+      // set the mosaic rule of the image service raster based on rule chosen from the combo box
+      comboBox.getSelectionModel().selectedItemProperty().addListener(e -> {
+        // create a new mosaic rule
+        mosaicRule = new MosaicRule();
```
JonLavi, line 126: There's an argument here for checking out whether it's necessary to overwrite our member variable `MosaicRule mosaicRule`, and then explicitly set it `imageServiceRaster.setMosaicRule(mosaicRule);` every time (line 179).
I'm wondering if it'd be possible to set the mosaic rule once when the app starts up, and then just modify the contents of the member variable `MosaicRule mosaicRule`, and the rule would update properly in the view every time? Let's discuss this on a call, it might be easier to explain (and we can test it!)

JonLavi (commented):
Nice work @sclaridge , this looks great.
I've made a few suggestions for re-ordering and re-wording of comments.
One interesting thing to look at together would be the use of the member variable `mosaicRule`, and whether we need to set that explicitly to the `imageServiceRaster` every time we change it's contents. 
Unfortunately I am not able to test this right now as the service seems to be down. Let's discuss it over a call and see if there's any improvements to make

raster/apply-mosaic-rule-to-rasters/build.gradle
sclaridge: The `createGradlePropertiesAndWriteApiKey` task should create the `gradle properties` file when the sample is first built. Does this happen if the sample is built, the file is deleted and the sample is run after?

JonLavi: It doesn't seem to do that for me, I have to run the task `createGradlePropertiesAndWriteApiKey` manually. However it might just be something iffy with my setup. If it works for other people then there's nothing to worry about

raster/apply-mosaic-rule-to-rasters/README.md
sclaridge: The design has been finalized, so I'll stick with rasters!

sclaridge:
@JonLavi I've created a `setupUI()` method to hide the hides the JavaFX controls, let me know what you think!

raster/apply-mosaic-rule-to-rasters/src/main/java/com/esri/samples/apply_mosaic_rule_to_rasters/ApplyMosaicRuleToRastersSample.java
```diff
@@ -0,0 +1,214 @@
...
+
+    // create a combo box
+    mosaicRuleComboBox = new ComboBox<>();
+    mosaicRuleComboBox.setMaxWidth(Double.MAX_VALUE);
+
+    // add the mosaic rules to the combo box
+    mosaicRuleComboBox.getItems().addAll("Default", "Northwest", "Center", "By attribute", "Lock raster");
+
+    // set the default combo box value
+    mosaicRuleComboBox.getSelectionModel().select(0);
```
JonLavi: If you wanted, you could instantiate the combobox directly with the items, instead of adding them in a separate line,
```suggestion
    // create a combo box with the mosaic rules
    mosaicRuleComboBox = new ComboBox<>(
            FXCollections.observableArrayList("Default", "Northwest", "Center", "By attribute", "Lock raster"));
    mosaicRuleComboBox.setMaxWidth(Double.MAX_VALUE);

    // set the default combo box value
    mosaicRuleComboBox.getSelectionModel().select(0);

```

raster/apply-mosaic-rule-to-rasters/src/main/java/com/esri/samples/apply_mosaic_rule_to_rasters/ApplyMosaicRuleToRastersSample.java
```diff
@@ -0,0 +1,214 @@
+/*
+ * Copyright 2020 Esri.
```
JonLavi: 2021 by now? 😁

raster/apply-mosaic-rule-to-rasters/src/main/java/com/esri/samples/apply_mosaic_rule_to_rasters/ApplyMosaicRuleToRastersSample.java
```diff
@@ -0,0 +1,214 @@
...
+import com.esri.arcgisruntime.raster.MosaicRule;
+import com.esri.arcgisruntime.raster.MosaicMethod;
+import com.esri.arcgisruntime.raster.MosaicOperation;
+
+public class ApplyMosaicRuleToRastersSample extends Application {
+
+  private ComboBox<String> mosaicRuleComboBox;
+  private MapView mapView;
+  private MosaicRule mosaicRule;
+  private RasterLayer rasterLayer; // keep loadable in scope to avoid garbage collection
```
JonLavi: I don't think there's any risk of RasterLayer going out of scope (it's inside the Map), but I might be wrong and misremembering the reason for this fix. Might be worth looking up again?

raster/apply-mosaic-rule-to-rasters/src/main/java/com/esri/samples/apply_mosaic_rule_to_rasters/ApplyMosaicRuleToRastersSample.java
```diff
@@ -0,0 +1,214 @@
...
+
+      // show progress indicator while map view is drawing
+      mapView.addDrawStatusChangedListener(drawStatusChangedEvent ->
+        progressIndicator.setVisible(drawStatusChangedEvent.getDrawStatus() == DrawStatus.IN_PROGRESS));
+
+      // set the map to the map view
+      mapView.setMap(map);
+
+      // create a raster layer from the image service raster
+      ImageServiceRaster imageServiceRaster = new ImageServiceRaster("https://sampleserver7.arcgisonline.com/server/rest/services/amberg_germany/ImageServer");
```
JonLavi: Should we stick to the character limit? I think shortening it like this would be enough
```suggestion
      ImageServiceRaster imageServiceRaster = new ImageServiceRaster(
      "https://sampleserver7.arcgisonline.com/server/rest/services/amberg_germany/ImageServer");
```

raster/apply-mosaic-rule-to-rasters/src/main/java/com/esri/samples/apply_mosaic_rule_to_rasters/ApplyMosaicRuleToRastersSample.java
```diff
@@ -0,0 +1,214 @@
...
+
+      // authentication with an API key or named user is required to access basemaps and other location services
+      String yourAPIKey = System.getProperty("apiKey");
+      ArcGISRuntimeEnvironment.setApiKey(yourAPIKey);
+
+      // setup the UI
+      setupUI();
+
+      // create a progress indicator
+      ProgressIndicator progressIndicator = new ProgressIndicator(ProgressIndicator.INDETERMINATE_PROGRESS);
```
JonLavi: Leaving out the type also works here.
Considering it's UI I'd be tempted to shove it in `setupUI` as well, but it would require making it a class member, so that decision is up to you!
```suggestion
      ProgressIndicator progressIndicator = new ProgressIndicator();
```

raster/apply-mosaic-rule-to-rasters/src/main/java/com/esri/samples/apply_mosaic_rule_to_rasters/ApplyMosaicRuleToRastersSample.java
```diff
@@ -0,0 +1,214 @@
...
+            break;
+        }
+        // set the mosaic rule of the image service raster
+        imageServiceRaster.setMosaicRule(mosaicRule);
+      });
+
+      // add the map view, control panel and progress indicator to the stack pane
+      stackPane.getChildren().addAll(mapView,controlsVBox, progressIndicator);
+      StackPane.setAlignment(controlsVBox, Pos.TOP_LEFT);
+      StackPane.setMargin(controlsVBox, new Insets(10, 0, 0, 10));
```
JonLavi: perhaps we can put some of this in `setupUI` as well?
It's good to show that the `MapView` belongs in the `StackPane` though.
It would be quite a big pattern change, so maybe worth a discussion with the team, however it'd make away with a lot of JavaFX code and leave the Runtime code front an centre, so just an idea.

raster/apply-mosaic-rule-to-rasters/src/main/java/com/esri/samples/apply_mosaic_rule_to_rasters/ApplyMosaicRuleToRastersSample.java
```diff
@@ -0,0 +1,214 @@
...
+            mosaicRule.getLockRasterIds().clear();
+            mosaicRule.getLockRasterIds().addAll(Arrays.asList(1L, 7L, 12L));
+            break;
+        }
+        // set the mosaic rule of the image service raster
+        imageServiceRaster.setMosaicRule(mosaicRule);
+      });
+
+      // add the map view, control panel and progress indicator to the stack pane
+      stackPane.getChildren().addAll(mapView,controlsVBox, progressIndicator);
```
JonLavi: Missing a space here:
```suggestion
      stackPane.getChildren().addAll(mapView, controlsVBox, progressIndicator);
```

JonLavi (approved):
Nice work @sclaridge . Approving so that you can move on to the second review.
Just a few ideas for you.

One thing I noticed is that the progress spinner appears even when right-clicking, which can be a bit misleading as right-click doesn't do anything in this sample. It also appears when panning, which is technically correct (the `MapView` is in 'drawing' state), but also not really representative for what's going on.
One way to get around that would be to add the listener when making the selection in the combobox, and remove it again once the drawing is finished. If this complicates the code too much though, perhaps it would be better to leave as is.

sclaridge:
@Rachael-E I've moved the JavaFX set up and progress indicator into a separate  `setupUI` method, let me know what you think!

raster/apply-mosaic-rule-to-rasters/build.gradle
Rachael-E: we should look into this more: if Jon is coming across this as a first time using this task others will too. I can take a look on a fresh windows setup!

raster/apply-mosaic-rule-to-rasters/src/main/java/com/esri/samples/apply_mosaic_rule_to_rasters/ApplyMosaicRuleToRastersSample.java
```diff
@@ -0,0 +1,214 @@
+/*
+ * Copyright 2021 Esri.
```
Rachael-E, line 2: this would have been ok to leave as 2020 since that's when the code was originally authored: no harm to update it to 2021 but just a note for future!

raster/apply-mosaic-rule-to-rasters/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+This sample uses a [raster image service](https://sampleserver7.arcgisonline.com/server/rest/services/amberg_germany/ImageServer) hosted on *ArcGIS Online* that shows aerial images of Amberg, Germany.
+
+## Additional information
+
+For more information, see [Understanding the mosaicking rules](https://desktop.arcgis.com/en/arcmap/latest/manage-data/raster-and-images/understanding-the-mosaicking-rules-for-a-mosaic-dataset.htm) from *ArcGIS Desktop* documentation. To learn more about how to define certain mosaic rules, see [Mosaic rule objects](https://developers.arcgis.com/documentation/common-data-types/mosaic-rules.htm) from *ArcGIS for Developers*.
+
+## Tags
+
+image service, mosaic rule, mosaic method, raster
```
Rachael-E: I think the checker is picking up that these aren't sorted alphabetically, gosh I wouldn't have noticed this at all otherwise! 

```suggestion
image service, mosaic method, mosaic rule, raster
```

raster/apply-mosaic-rule-to-rasters/README.md
Rachael-E: I agree with Jon here, though checking the design I now understand why it's plural:

> Mosaic rules are applied to mosaic datasets, which are a mesh of several rasters managed together as one

raster/apply-mosaic-rule-to-rasters/src/main/java/com/esri/samples/apply_mosaic_rule_to_rasters/ApplyMosaicRuleToRastersSample.java
```diff
@@ -0,0 +1,214 @@
...
+      stage.setHeight(700);
+      stage.setScene(scene);
+      stage.show();
+
+      // authentication with an API key or named user is required to access basemaps and other location services
+      String yourAPIKey = System.getProperty("apiKey");
+      ArcGISRuntimeEnvironment.setApiKey(yourAPIKey);
+
+      // setup the UI
+      setupUI();
```
Rachael-E: let's move this to after the mapview has been set to the map: keeps consistency with our other samples where we open the code with our own Runtime specific set up.

raster/apply-mosaic-rule-to-rasters/src/main/java/com/esri/samples/apply_mosaic_rule_to_rasters/ApplyMosaicRuleToRastersSample.java
Rachael-E: yeh we could add that in here since this dialog box would pop up: if there is a meaningful message. Sometimes there isn't in which case we don't put that extra detail in.

raster/apply-mosaic-rule-to-rasters/src/main/java/com/esri/samples/apply_mosaic_rule_to_rasters/ApplyMosaicRuleToRastersSample.java
```diff
@@ -0,0 +1,214 @@
...
+            mosaicRule.setMosaicOperation(MosaicOperation.BLEND);
+            break;
+          case "By attribute":
+            mosaicRule.setMosaicMethod(MosaicMethod.ATTRIBUTE);
+            mosaicRule.setSortField("OBJECTID");
+            break;
+          case "Lock raster":
+            mosaicRule.setMosaicMethod(MosaicMethod.LOCK_RASTER);
+            mosaicRule.getLockRasterIds().clear();
+            mosaicRule.getLockRasterIds().addAll(Arrays.asList(1L, 7L, 12L));
```
Rachael-E, line 148: I'm not too sure as a user why I have to add these lock raster IDs: can you add a comment to explain what this is? Is this selecting certain rasters from the mosaic dataset to display?

raster/apply-mosaic-rule-to-rasters/src/main/java/com/esri/samples/apply_mosaic_rule_to_rasters/ApplyMosaicRuleToRastersSample.java
```diff
@@ -0,0 +1,214 @@
...
+      e.printStackTrace();
+    }
+  }
+
+  /**
+   * Creates a UI with a drop down box.
+   */
+  private void setupUI() {
+    // create a progress indicator
+    progressIndicator = new ProgressIndicator();
```
Rachael-E: I'd move the progressIndicator out of here and put it in place where it is first called in the start method. You could then also remove it as a member variable.

raster/apply-mosaic-rule-to-rasters/src/main/java/com/esri/samples/apply_mosaic_rule_to_rasters/ApplyMosaicRuleToRastersSample.java
```diff
@@ -0,0 +1,214 @@
...
+      setupUI();
+
+      // create a map with the light gray basemap style
+      ArcGISMap map = new ArcGISMap(BasemapStyle.ARCGIS_LIGHT_GRAY);
+
+      // create a map view
+      mapView = new MapView();
+
+      // show progress indicator while map view is drawing
+      mapView.addDrawStatusChangedListener(drawStatusChangedEvent ->
```
Rachael-E:
```suggestion
      // show progress indicator while map view is drawing
      var progressIndicator = new ProgressIndicator();
      mapView.addDrawStatusChangedListener(drawStatusChangedEvent ->
```

Rachael-E (commented):
thanks @sclaridge neat small sample! just a few minor comments and reordering of code to consider, as well as an update to the readme thanks to the new checker!

sclaridge:
@Rachael-E I've added back the original feature service URL and rearranged some of the UI code.

Rachael-E (approved):

PR #625, opened 2021-03-03
Show labels on layer in 3D sample

Rachael-E:
### Description

This is for the "Show labels on layer in 3D" sample, targeted at the 100.11.0 release so that we can demonstrate the new Arcade expression script functionality for label definitions. It loads a portal item in a scene, looks through the layers to find the layer which contains labels, and defines a label expression to display it in the application. 

I'm also experimenting with more modern features in this PR: namely the use of type inference with `var` where I think is appropriate/readable. Keen to hear thoughts on this approach! 

@jenmerritt could you take a first look please? I'll keep an eye out for the auto checks too. Thanks!


scene/show-labels-on-layer-in-3d/src/main/java/com/esri/samples/show_labels_on_layer_in_3d/ShowLabelsOnLayerIn3dSample.java
```diff
@@ -0,0 +1,138 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+
+package com.esri.samples.show_labels_on_layer_in_3d;
+
+import com.esri.arcgisruntime.arcgisservices.LabelingPlacement;
```
jenmerritt: Just the runtime imports here to shift below javafx

scene/show-labels-on-layer-in-3d/src/main/java/com/esri/samples/show_labels_on_layer_in_3d/ShowLabelsOnLayerIn3dSample.java
```diff
@@ -0,0 +1,138 @@
...
+
+      // set title, size, and add the JavaFX scene to stage
+      stage.setTitle("Show Labels on Layer in 3D");
+      stage.setWidth(800);
+      stage.setHeight(700);
+      stage.setScene(fxScene);
+      stage.show();
+
+      // create a scene from an ArcGISOnline portal item
+      var portalItem = new PortalItem(
```
jenmerritt, line 61: I'm onboard with the uses of `var`  in the Sample. Think they make sense and it remains very clear what Class is being used.

scene/show-labels-on-layer-in-3d/src/main/java/com/esri/samples/show_labels_on_layer_in_3d/ShowLabelsOnLayerIn3dSample.java
```diff
@@ -0,0 +1,138 @@
...
+      Scene fxScene = new Scene(stackPane);
+
+      // set title, size, and add the JavaFX scene to stage
+      stage.setTitle("Show Labels on Layer in 3D");
+      stage.setWidth(800);
+      stage.setHeight(700);
+      stage.setScene(fxScene);
+      stage.show();
+
+      // create a scene from an ArcGISOnline portal item
```
jenmerritt:
```suggestion
      // create a scene from an ArcGIS Online portal item
```

scene/show-labels-on-layer-in-3d/src/main/java/com/esri/samples/show_labels_on_layer_in_3d/ShowLabelsOnLayerIn3dSample.java
```diff
@@ -0,0 +1,138 @@
...
+
+          // filter through the scene's operational layers to find the layer named "Gas",
+          // and find the "Gas Main" sublayer to apply the label definition to
+          scene.getOperationalLayers().stream()
+            .filter(layer -> layer.getName().equals("Gas"))
+            .flatMap(gasLayer -> gasLayer.getSubLayerContents().stream())
+            .forEach(subLayer -> {
+              if (subLayer.getName().equals("Gas Main")) {
+                FeatureLayer featureLayer = (FeatureLayer) subLayer;
+                featureLayer.setLabelsEnabled(true);
```
jenmerritt, line 88: Is this an essential step in the workflow to showing labels? i.e. is it worth highlighting with a comment?

scene/show-labels-on-layer-in-3d/src/main/java/com/esri/samples/show_labels_on_layer_in_3d/ShowLabelsOnLayerIn3dSample.java
```diff
@@ -0,0 +1,138 @@
...
+                labelDefinition.setPlacement(LabelingPlacement.LINE_ABOVE_ALONG);
+                labelDefinition.setUseCodedValues(true);
+                labelDefinition.setTextSymbol(textSymbol);
+
+                // add the label definition to the feature layer
+                featureLayer.getLabelDefinitions().add(labelDefinition);
+              }
+            });
+        } else {
+          System.out.println("Scene failed to load " + scene.getLoadError().getCause());
```
jenmerritt: I'm just curious here, as how we handle errors is something that came up a lot in review discussions previously and I'm not sure a consistent pattern was ever really reached or agreed upon. What are you thinking going forward now you have the reins longer term - to have things being printed out rather than in an Alert? Or depending on the situation? I still don't really know what I prefer, I guess for the target developer users printing out is sufficient. I just find it hard to escape being a user looking at the app with the progress indicator spinning but not getting a message as to why. Thinking aloud!

jenmerritt (approved):
@Rachael-E Looks good! Just a few thoughts to throw in the mix but not essential. So will approve.

I also looked at the readme script output - it's the same problem as before I think, it wants `URL` to be higher up the alphabet than `arcade` in the `keywords` section. Was this going to be discussed by the team as to whether it stays like this or could/should be adjusted in the script?

Rachael-E:
thanks Jen! I'll get another PR in to fix up the Github actions: it's missing the caseless matching in this instance again as you say. @colinanderson please could you do a code review? thanks!

scene/show-labels-on-layer-in-3d/build.gradle
```diff
@@ -0,0 +1,75 @@
+plugins {
+    id 'application'
+    id 'org.openjfx.javafxplugin' version '0.0.8'
```
colinanderson, line 3: There is a 0.0.9 version now.

scene/show-labels-on-layer-in-3d/src/main/java/com/esri/samples/show_labels_on_layer_in_3d/ShowLabelsOnLayerIn3dSample.java
```diff
@@ -0,0 +1,140 @@
...
+      stage.setTitle("Show Labels on Layer in 3D");
+      stage.setWidth(800);
+      stage.setHeight(700);
+      stage.setScene(fxScene);
+      stage.show();
+
+      // create a scene from an ArcGIS Online portal item
+      var portalItem = new PortalItem(
+        new Portal("https://www.arcgis.com", false), "850dfee7d30f4d9da0ebca34a533c169");
+      ArcGISScene scene = new ArcGISScene(portalItem);
```
colinanderson, line 63: You could also use `var` here.

scene/show-labels-on-layer-in-3d/src/main/java/com/esri/samples/show_labels_on_layer_in_3d/ShowLabelsOnLayerIn3dSample.java
```diff
@@ -0,0 +1,140 @@
...
+          // remove the progress indicator when the scene loads
+          progressIndicator.setVisible(false);
+
+          // filter through the scene's operational layers to find the layer named "Gas",
+          // and find the "Gas Main" sublayer to apply the label definition to
+          scene.getOperationalLayers().stream()
+            .filter(layer -> layer.getName().equals("Gas"))
+            .flatMap(gasLayer -> gasLayer.getSubLayerContents().stream())
+            .forEach(subLayer -> {
+              if (subLayer.getName().equals("Gas Main")) {
```
colinanderson: Can't you use another `filter` or `findFirst` or similar here?

colinanderson (commented):
@Rachael-E looks ok - a few minor comments.

scene/show-labels-on-layer-in-3d/src/main/java/com/esri/samples/show_labels_on_layer_in_3d/ShowLabelsOnLayerIn3dSample.java
Rachael-E: I've avoided using `var` here since we don't name this variable explicitly `arcGISScene`: since we also have the JavaFX scene perhaps it could be confused. The guidelines for using `var` suggest only to use it when it's obvious from the local variable name what the type is, so I think sticking here with declaring the type works. 

We could consider renaming `map` and `scene` local variables to `arcGISMap` and `arcGISScene` and use `var` that way, but I think that makes it less readable.

PR #626, opened 2021-03-10
Update blurry basemap imagery

Rachael-E:
### Description

This PR updates or removes the blurry basemap imagery imposed over a part of the city of Brest, France, aerial imagery. It replaces the Brest data for Viewshed camera with an integrated mesh layer from Girona, and updates the basemap type from aerial to topographic in Scene Layer, Distance Measurement and Scene Layer Selection. It also updates the images and a gif following the basemap updates. @jenmerritt could you take a look please? Thanks!


jenmerritt:
Changes look good @Rachael-E  

Only thought - on Viewshed Camera for the image/gif, I like the viewpoint you used with the Cathedral building, but initially I thought there was a glitch in the image with the line going across it. I now realise this is a camera marking, but wonder if there's any way of getting more of the camera lines potentially rather than a straight line through it? Probably picky so feel free to ignore if it's not easy to get a different shot.

Not from your changes but one thing I also noticed when looking at the new images/gifs on the Sample Viewer was the Viewshed Location Sample's image seems to be 600x400 rather than 800x400. Not sure if it's worth tidying that in this PR or making a note for another time?

Rachael-E:
thanks @jenmerritt , I've updated both images as suggested above! could you have another look please?

jenmerritt (approved):
LGTM @Rachael-E 👍

PR #633, opened 2021-04-22
Update IdentifyKMLFeaturesSample.java

Rachael-E:
### Description

In recent updates, core introduced a condition to return an empty string for the balloons/KML nodes that doesn't have description or extended data. The data source in the Identify KML features doesn't provide a description for each significant weather condition, thus empty balloons are displayed in the callout.

If description is absent for a placemark, this PR adds a placeholder so that the content of the balloon can be displayed.

Before: the popup for any feature in the map is empty
In this PR: the popup should show HTML content

@jenmerritt could you take a look please?


jenmerritt (approved):
Your changes look fine @Rachael-E , I couldn't check it specifically due to what data was being served but code looks good to me!

Readme Checker failures fixes needed:
- Line 26 of Readme has unnecessary whitespace to be removed
- Metadata inconsistency: move "Keyhole" to the top of the keywords list before "KML"

Also, I suspect it's been like this a while and from a cursory look I think we're constrained by the HTML being sent back from the service so not sure how easy it would be to fix / if it's worth spending the time, but some of the text overruns so it's not readable:
image

Rachael-E:
thanks @jenmerritt : good spot with the webview, as you say it looks like we are limited by the html. I experimented with styling of the Webview component but without spending too much more time on it I couldn't fix anything up obviously. Something to come back to!

PR #634, opened 2021-04-22
Add an additional condition to handle null pointer exception

Rachael-E:
### Description

This PR adds an additional check that handles the throwing of a null pointer exception in the event that no errors are returned. This has been blocking the sample and leading to a blank user interaction on the sample viewer. @sclaridge could you take a look please?


sclaridge (approved):
Looks good in the sample viewer :+1:

PR #636, opened 2021-05-03
Sample updates

Rachael-E:
### Description

This PR fixes up some issues that were raised during testing/sample viewer verification (common samples issues 2473 and 1845). The changes are small:

- Service area task: increase UI width so that text isn't truncated on Linux
- Manage operational layers: zoom out to show more of the USA for the range of data shown (is variable and the further zoom out should capture more of the data as it varies)
- Download preplanned map controller: reduce the time it takes for the viewpoint to transition and add a listener to prevent a blank map appearing if a user clicks rapidly whilst the mapview is navigating
- Create and save map: update to use new basemapstyle constructor workflow by creating a new map everytime and setting a basemap style to it. This is also in sync with the Change Basemap sample now, and there is a minor improvement to that sample in this PR also (setting the viewpoint once rather than creating a new viewpoint everytime in the same place). 


jenmerritt (approved):
LGTM @Rachael-E

PR #637, opened 2021-05-06
More sample updates

Rachael-E:
### Description

This PR contains 3 small updates:

- Show callout: update to display lat/long rather than x and y co-ordinates
- Display subtype feature layer: increase vbox width to account for truncated text on linux
- Change basemap: use the `setBasemap` method to set the basemap rather than creating a new map everytime and cleaned up the viewpoint code slightly so that the viewpoint is only set when the app launches. That way if a user navigates the map and then changes the basemap, they'll remain in the same place to compare the new basemap.

### DIFF
--- a/map/change-basemap/src/main/java/com/esri/samples/change_basemap/ChangeBasemapSample.java
+++ b/map/change-basemap/src/main/java/com/esri/samples/change_basemap/ChangeBasemapSample.java
@@ -16,6 +16,7 @@
 
 package com.esri.samples.change_basemap;
 
+import com.esri.arcgisruntime.mapping.Basemap;
 import javafx.application.Application;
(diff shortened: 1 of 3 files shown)


map/change-basemap/src/main/java/com/esri/samples/change_basemap/ChangeBasemapSample.java
```diff
@@ -16,6 +16,7 @@
 
 package com.esri.samples.change_basemap;
 
+import com.esri.arcgisruntime.mapping.Basemap;
```
jenmerritt: Imports just need re-alphabetizing

jenmerritt (approved):
Changes look good @Rachael-E , just a stray import to tidy up.

PR #639, opened 2021-05-12
Support API keys on basemap for Create and Save map

Rachael-E:
### Description

Users should be able to use API Keys to access Basemaps to start with in an app and then continue to work with a named-user account. This PR updates the Create and Save map sample to show this workflow. @TADraeseke could you review please?


map/create-and-save-map/src/main/java/com/esri/samples/create_and_save_map/CreateAndSaveMapController.java
```diff
@@ -78,7 +69,6 @@ private void initialize() {
 
     // authentication with an API key or named user is required to access basemaps and other location services
```
TADraeseke: I just wonder if this comment is a bit confusing here as this line is technically just defining a string. The actual authentication is now happening below

TADraeseke (commented):
LGTM! Just one thought about the original comment now that API key code has moved

Rachael-E:
thanks @TADraeseke , I've updated that comment! @sclaridge could you take a second look please?

TADraeseke (approved):
LGTM! Thanks @Rachael-E!

Rachael-E:
@TADraeseke @jenmerritt I have updated this PR to follow the slight change in scope we had originally for this sample. This PR now:
- removes the Application Credentials workflow from Create and Save map (i.e. removes the option to add a client ID, url and redirect window), and the setting of the API Key. This sample will now authenticate basemaps via the ArcGIS Identity (named user login) following successful loading of the portal.
- updates integrated windows authentication sample to authenticate and display a portal map via ArcGIS Identity, and removes setting of the API key from the sample. Setting the api key would trigger an additional authentication window to authenticate the basemap, but since we are displaying a portal map we don't need that initial basemap.

Please could you take a look? Thanks!

map/create-and-save-map/README.md
```diff
@@ -1,24 +1,24 @@
...
 
 ## How to use the sample
 
-This sample requires you to setup your own app on arcgis.com. See https://github.com/Esri/arcgis-runtime-samples-java/blob/master/OAuth.md for instructions.
-
-Fill in your portal and registered app credentials in the starting dialog to authenticate. Then, choose the basemap and layers for your new map. To save the map, choose a title, tags and description (optional), and a folder on your portal (you will need to create one in your portal's My Content). Click the Save button to save the map to the chosen folder.
+When you run the sample, you will be challenged for an ArcGIS Online login. Enter a user name and password for an ArcGIS Online named user account (such as your ArcGIS for Developers account). Then, choose the basemap and layers for your new map. To save the map, choose a title, tags and description (optional), and a folder on your portal (you will need to create one in your portal's My Content section if you don't already have one). Click the Save button to save the map to the chosen folder.
```
TADraeseke:
```suggestion
When you run the sample, you will be challenged for an ArcGIS Online login. Enter a user name and password for an ArcGIS Online named user account (such as your ArcGIS for Developers account). Then, choose the basemap and layers for your new map. To save the map, add a title, tags and description (optional), and a folder on your portal (you will need to create one in your portal's My Content section if you don't already have one). Click the Save button to save the map to the chosen folder.
```

Choose makes me think there might be a list to choose from

.github/workflows/gradle.yml
```diff
@@ -5,9 +5,9 @@ name: Java CI with Gradle
 
 on:
   push:
-    branches: [ master ]
+    branches: [ main ]
   pull_request:
-    branches: [ master, dev ]
+    branches: [ main, v.next ]
```
jenmerritt, line 10: nice spot!

jenmerritt (commented):
Changes look good and ran both samples with my credentials. I think there's an additional step that can be done to clean up the `build.gradle` file since the API Key is not longer being set. Want to arrange to look at that together?

jenmerritt (approved):
Looks good @Rachael-E

TADraeseke (approved):
LGTM! Thanks @Rachael-E - Just a small comment on wording in the README

PR #649, opened 2021-07-12
Replace deprecated utility network property

Rachael-E:
This PR replaces the `.getTraceConfiguration` method with `.getDefaultTraceConfiguration` method. The former will be deprecated in 100.12.0 on the `UtilityTier` and `UtilityNamedTraceConfiguration` classes. Reason being, even though these were intended to be read-only properties, they have the possibility of being accidentally edited. The new method `getDefaultTraceConfiguration()` will return the value from the source service.

@jenmerritt please could you review? thanks!


jenmerritt (approved):
LGTM @Rachael-E

PR #650, opened 2021-07-16
Generate offline map with local basemap update

Rachael-E:
This PR updates the data for this sample to tpkx and also calls an additional method that is required for Runtime to find the data. @jenmerritt could you take a look please? thanks!


jenmerritt (approved):
@Rachael-E the changes look good. Note the authentication issues we discussed, but not related to these changes.

PR #651, opened 2021-07-19
Update build.gradle files and remove gradle-wrapper.jars

Rachael-E:
This PR:

- removes instances of `gradle-wrapper.jar` : these aren't required to run the sample and not all samples have them, so brings the repo to a good level of consistency across samples
- updates the gradle version from 6.5.1 to 7.0.2
- excludes signing jars when building a jar
- updates the jar task to include dependencies


mbcoder (approved):
Looks good to me

Rachael-E:
thanks @mbcoder , @jenmerritt could you take a look please?

jenmerritt:
@Rachael-E 
2 comments:
- Does the root `gradle` folder also need its `gradle-wrapper.jar` deleting or is that one being kept deliberately?
- It looks like configure subnetwork trace `gradle-wrapper.properties` needs the version updating

Rachael-E:
cool nice spot @jenmerritt thanks! The gradle-wrapper.jar should be ok to keep at this high level, it was more at the sample level a developer may have issues with whilst running with gradle from the command line.

PR #654, opened 2021-08-06
[Sample Update] Show curve in add graphics renderer

Rachael-E:
This PR updates the Add Graphics with Renderer sample to include a heart shaped graphic, built using our new Curves API in 100.12.0. 

It also updates the samples readme, metadata and image, as well as making use of our ColorUtil class to set Colors rather than using argb value. @jenmerritt could you take a look please?


display_information/add-graphics-with-renderer/README.metadata.json
```diff
@@ -1,20 +1,30 @@
 {
     "category": "Display information",
-    "description": "A renderer allows you to change the style of all graphics in a graphics overlay by referencing a single symbol style. A renderer will only affect graphics that do not specify their own symbol style.",
+    "description": "A renderer allows you to change the style of all graphics in a graphics overlay by referencing a single symbol style. A renderer will only effect graphics that do not specify their own symbol style.",
```
jenmerritt:
```suggestion
    "description": "A renderer allows you to change the style of all graphics in a graphics overlay by referencing a single symbol style. A renderer will only affect graphics that do not specify their own symbol style.",
```

I think affect is correct? The renderer would apply an effect which would cause an affect.

display_information/add-graphics-with-renderer/src/main/java/com/esri/samples/add_graphics_with_renderer/AddGraphicsWithRendererSample.java
```diff
@@ -108,18 +118,83 @@ public void start(Stage stage) {
...
+   *
+   * @param centerOfHeart the center of the square that contains the heart shape
+   * @return a heart-shape geometry
+   */
+  private Geometry makeHeartGeometry(Point centerOfHeart) {
+
+    // define the side length of the square that contains the heart shape
+    double sideLength = 10e5;
+
+    SpatialReference spatialReference = centerOfHeart.getSpatialReference();
```
jenmerritt, line 158: I presume the spatial reference being set here is pretty essential to the curve process? Maybe a comment above that would be useful.

display_information/add-graphics-with-renderer/src/main/java/com/esri/samples/add_graphics_with_renderer/AddGraphicsWithRendererSample.java
```diff
@@ -108,18 +118,83 @@ public void start(Stage stage) {
...
       // add the map view to stack pane
       stackPane.getChildren().add(mapView);
+
     } catch (Exception e) {
       e.printStackTrace();
     }
   }
 
+  /**
+   * Create a heart-shape geometry with bezier and elliptic arc segments.
```
jenmerritt:
```suggestion
   * Create a heart-shaped geometry with bezier and elliptic arc segments.
```

display_information/add-graphics-with-renderer/src/main/java/com/esri/samples/add_graphics_with_renderer/AddGraphicsWithRendererSample.java
```diff
@@ -108,18 +118,83 @@ public void start(Stage stage) {
...
+   *
+   * @param centerOfHeart the center of the square that contains the heart shape
+   * @return a heart-shape geometry
```
jenmerritt:
```suggestion
   * @return a heart-shaped geometry
```

display_information/add-graphics-with-renderer/src/main/java/com/esri/samples/add_graphics_with_renderer/AddGraphicsWithRendererSample.java
```diff
@@ -108,18 +118,83 @@ public void start(Stage stage) {
...
+   */
+  private Geometry makeHeartGeometry(Point centerOfHeart) {
```
jenmerritt, line 153: Does this need to be a separate method? It does remove some lines from the main method but maybe isn't necessary if the comments separate out each of the graphics quite clearly?

display_information/add-graphics-with-renderer/src/main/java/com/esri/samples/add_graphics_with_renderer/AddGraphicsWithRendererSample.java
```diff
@@ -108,18 +118,83 @@ public void start(Stage stage) {
...
+
+    // create a part and add a collection of segments to it to define the heart shape
+    Part part = new Part(spatialReference);
+    part.addAll(Arrays.asList(leftCurve, leftArc, rightArc, rightCurve));
+
+    // use a polygon builder to construct a heart shape from the parts above
+    PolygonBuilder heartShape = new PolygonBuilder(spatialReference);
+    heartShape.getParts().add(part);
+
+    // return the geometry of the heart shape polygon
```
jenmerritt:
```suggestion
    // return the geometry of the heart-shaped polygon
```

display_information/add-graphics-with-renderer/src/main/java/com/esri/samples/add_graphics_with_renderer/AddGraphicsWithRendererSample.java
```diff
@@ -108,18 +118,83 @@ public void start(Stage stage) {
...
       mapView.getGraphicsOverlays().add(polygonGraphicOverlay);
 
+      // polygon with curve segments (red heart) graphic
+      GraphicsOverlay curvedGraphicOverlay = new GraphicsOverlay();
+      // create a simple fill symbol with outline
+      SimpleLineSymbol curvedLineSymbol = new SimpleLineSymbol(SimpleLineSymbol.Style.SOLID, ColorUtil.colorToArgb(Color.BLACK), 1);
+      SimpleFillSymbol curvedFillSymbol = new SimpleFillSymbol(SimpleFillSymbol.Style.SOLID, ColorUtil.colorToArgb(Color.RED), curvedLineSymbol);
+      SimpleRenderer curvedRenderer = new SimpleRenderer(curvedFillSymbol);
+      curvedGraphicOverlay.setRenderer(curvedRenderer);
+      // create a heart shaped graphic and add it to the map view
```
jenmerritt:
```suggestion
      // create a heart-shaped graphic and add it to the map view
```

jenmerritt (approved):
Looks good @Rachael-E ! Cool to see the true curves after hearing so much about it.

Unless it's a requirement I'm not sure the separate method is necessary but up to you. Also just make sure the consistency is there for "heart-shaped" - I'm hoping I've suggested them all but might be worth a double check!

display_information/add-graphics-with-renderer/src/main/java/com/esri/samples/add_graphics_with_renderer/AddGraphicsWithRendererSample.java
Rachael-E: I was back and forth with this too. I think in this case, even though it's single use, the method approach works well as it keeps the logic for setting the renderers across each different geometry standard and bite sized, e.g. create the graphics overlay, create the symbol and fill, create the renderer from that and apply to the graphics overlay, then add the graphic to the overlay. it's a good repeated code "story" for each geometry, and the user can dig down into the method to find more details for the heart shape. 

we will have more functionality coming for the curves API, I think it might be nice to have a separate sample for this so that the curve creation code is right at the centre (or the heart? :p ) of the sample.

PR #661, opened 2021-09-06
Update export vector tiles sample to use vector basemaps

Rachael-E:
This PR updates the existing Export Vector Tiles sample to use the `STREETS_NIGHT` vector tiled basemap, allowing authentication via API key rather than named user. This PR updates the sample code along with the build.gradle script and readme. It also includes a few updates to the similar Export Tiles sample.  @jenmerritt could you take a look please? thanks!


jenmerritt (approved):
LGTM @Rachael-E

PR #663, opened 2021-09-09
update open street map layer sample

Rachael-E:
Our "OpenStreetMap layer" sample uses an OSM basemap. The original intention of this sample was to demonstrate use of the OpenStreetMapLayer class, this sample is updated to reflect that. @jenmerritt @mbcoder could you take a look please? thanks!


jenmerritt (approved):
LGTM

PR #674, opened 2022-01-13
Add dispose method to preview map view

Rachael-E:
The sample viewer was occasionally crashing due to the second map view in the export tiles and export vector tiles samples not being disposed of. This PR addresses this. @colinanderson could you have a look please?


colinanderson (approved):
@Rachael-E looks ok.

PR #676, opened 2022-02-17
Update scene based samples to use basemap styles

Rachael-E:
With 100.13 we have full support for vector tiled layers in scenes which means all samples should now be using the new BasemapStyles wherever an Esri basemap is relevantly used.

This PR updates around 34 samples to use these API key metered BasemapStyles, and each sample has therefore had its gradle script updated to allow use of the API key functionality along with an update to their main class to remove reference to existing Basemap.create classes and to add the API key setter.

@jenmerritt could you take a look please? Thanks


jenmerritt (approved):
LGTM @Rachael-E . As mentioned, you could update the image for Line of Sight but it's not a huge deviation and the others all seem fine.

Rachael-E:
thanks Jen, I've updated the image and gif for the Line of Sight Location sample.

PR #677, opened 2022-02-21
Add redirect urls to Toggle between feature modes sample

Rachael-E:
This PR adds redirects to the newer Toggle between feature modes sample. This newer sample combined the original three individual samples, and as such requires a redirect url so that if a user has bookmarked an earlier version of the sample, they will be redirected to this one. @jenmerritt please could you take a look? thanks!


jenmerritt (approved):
LGTM

PR #678, opened 2022-03-02
Sketch Editor sample: add rectangle

Rachael-E:
This PR updates the Sketch on Map sample to include the draw a rectangle functionality of the Sketch Editor. 

@colinanderson could you take a look please? 

I have an observation about what happens when we edit the rectangle after saving it that I'd be keen for your thoughts on.  It looks like we can't access the graphics geometry as a rectangle, only as a polygon. Which means if the sketch is already "saved", the rectangle geometry will be edited as polygon and not be retained as a rectangle. This might be an API enhancement we wish to consider. 

tagging @kathleen1 for QA headsup.


Rachael-E:
closing this in favour of cleaner merge...

PR #679, opened 2022-03-02
Sketch Editor sample: add rectangle

Rachael-E:
This PR updates the Sketch on Map sample to include the draw a rectangle functionality of the Sketch Editor. 

@colinanderson could you take a look please? 

I have an observation about what happens when we edit the rectangle after saving it that I'd be keen for your thoughts on.  It looks like we can't access the graphics geometry as a rectangle, only as a polygon. Which means if the sketch is already "saved", the rectangle geometry will be edited as polygon and not be retained as a rectangle. This might be an API enhancement we wish to consider. 

tagging @kathleen1 for QA headsup.


jenmerritt (approved):
Changes look good to me in terms of the sample @Rachael-E  

Agree about the rectangle vs polygon observation - I'd say its not just after you save and edit, it's after you've initially drawn you can just click around and it's no longer a rectangle. Unsure of how it's meant to work though to be more helpful.

colinanderson:
For the polygon behaviour if you start adding extra points this is as designed. 
```
* A tool to sketch a rectangle polygon on the sketch overlay. A rectangle is drawn by three click points - the first
* two establish a base line and the third defines the rectangle on one side of the base line. Once a rectangle has been
* created it can be edited like a normal polygon. If editing reduces the polygon to two points or fewer it will then
* behave like creating a rectangle.
```

colinanderson (approved):
@Rachael-E looks ok.

PR #681, opened 2022-03-18
Update Nearest Vertex sample

Rachael-E:
This PR addresses a planar vs geodetic distance bug in this sample. It updates the sample to use a distance preserving projected coordinate system, to work with the current planar nearest parts of our API code (so that both geometry and calculations are carried out in planar SR).

@jenmerritt could you take a look please?

tagging @kathleen1 and @jasongao1230 as there are GUI changes in this PR.


geometry/nearest-vertex/src/main/java/com/esri/samples/nearest_vertex/NearestVertexSample.java
```diff
@@ -35,24 +35,28 @@
...
+import com.esri.arcgisruntime.portal.Portal;
+import com.esri.arcgisruntime.portal.PortalItem;
 import com.esri.arcgisruntime.symbology.SimpleFillSymbol;
 import com.esri.arcgisruntime.symbology.SimpleLineSymbol;
 import com.esri.arcgisruntime.symbology.SimpleMarkerSymbol;
 
 public class NearestVertexSample extends Application {
 
   private MapView mapView;
+  // California zone 5 (ftUS) state plane coordinate system.
```
jenmerritt:
```suggestion
  // California zone 5 (ftUS) state plane coordinate system
```

geometry/nearest-vertex/src/main/java/com/esri/samples/nearest_vertex/NearestVertexSample.java
```diff
@@ -85,14 +95,16 @@ public void start(Stage stage) {
       GraphicsOverlay graphicsOverlay = new GraphicsOverlay();
       mapView.getGraphicsOverlays().add(graphicsOverlay);
 
-      // create a graphic for the polygon
-      PointCollection polygonPoints = new PointCollection(SpatialReferences.getWebMercator());
+      // create a point collection to construct a polygon with, using the California zone 5 (ftUS) state plane coordinate system
```
jenmerritt:
```suggestion
      // construct a polygon from a point collection that uses the California zone 5 (ftUS) state plane coordinate system
```

jenmerritt (approved):
Looks good to me @Rachael-E . Couple of minor comment suggestions. Also one observation is the orange colour on the cross is slightly less obvious given the new feature layer also being orange, passable I think but could consider another colour, not sure if other platforms are the same?

Rachael-E:
thanks @jenmerritt , I've updated the colour to make it a more obvious orange and also changed the green: have made use of our ColorUtil api code to make this sample a bit more modern.

PR #684, opened 2022-04-19
Update Apply Scheduled Updates To Preplanned Map Area sample

Rachael-E:
This PR fixes a bug in the apply scheduled updates to preplanned map area sample, where the updated mobile map package wasn't displaying on the map. This has been addressed by instantiating a new instance of the updated mobile map package and now works as expected. 

@jenmerritt please could you review? thanks!


jenmerritt (approved):
LGTM

PR #688, opened 2022-06-24
Remove deprecated code

Rachael-E:
This PR removes and replaces deprecated code calls. @mbcoder could you review please?


mbcoder (approved):
More than I expected, but they look good changes.  I've made before and after runs of the apps and they still work.

PR #691, opened 2022-07-08
[New Sample] Set Max Extent

01smito01:
### Description

Implementation of the set max extent sample.

Would appreciate advice on the structure of the code and the complexity of blocks below a single comment -  there were some places (EG initialisation of the on/off buttons) where I'm not sure if there's too much going on within the one block of code.

Would also be good to get advice on use of the `var` keyword - I've used it quite liberally when initialising variables but I'm not sure if this is standard practice or if it impacts legibility.

Checklist:
- [x] Set target branch to main (current release) or dev (next release)
- [x] Request a reviewer
- [x] Assign a reviewer
- [x] Tag reviewer in comment


map/set-max-extent/README.metadata.json
```diff
@@ -0,0 +1,29 @@
...
+        "map",
+        "mapview",
+        "max extent",
+        "zoom",
+        "ArcGISMap",
+        "Envelope"
+    ],
+    "redirect_from": [
+        "/java/latest/sample-code/set-max-extent.htm"
+    ],
```
Rachael-E: the `redirect_from` key in this instance can be left clear, since this is a new sample. This key is for samples for which a url may have existed already: e.g. we are going through a samples process of merging similar samples in to one. We would therefore copy those original sample URLS into redirect_from, so that if a user went to that original link, it wouldn't result in a 404 and would instead take them to the new sample. 

In this instance, the correct notation for there being no redirect is: 

```suggestion
    "redirect_from": "",
```

map/set-max-extent/README.metadata.json
```diff
@@ -0,0 +1,29 @@
...
+    "relevant_apis": [
+        "ArcGISMap",
+        "Envelope"
+    ],
+    "snippets": [
+        "src/main/java/com.esri.samples.set_max_extent/SetMaxExtentSample.java"
```
Rachael-E: this should be forward slash for the snippets: the website config won't recognise the package syntax

```suggestion
        "src/main/java/com/esri/samples/set_max_extent/SetMaxExtentSample.java"
```

map/set-max-extent/src/main/java/com.esri.samples.set_max_extent/SetMaxExtentLauncher.java
```diff
@@ -0,0 +1,9 @@
+package com.esri.samples.set_max_extent;
+
+/**
+ * Wrapper required for launching a JavaFX 11 app through Gradle or from a jar.
+ */
+public class SetMaxExtentLauncher {
+
+    public static void main(String[] args) { SetMaxExtentSample.main(args); }
```
Rachael-E: there's no reason why this couldn't be a one liner, but for consistency and better readability, we put this on a new line. Also looks like the IDE indents are out: I'll send a separate message to you about this

```suggestion
  public static void main(String[] args) {
    SetMaxExtentSample.main(args);
  }
```

map/set-max-extent/src/main/java/com.esri.samples.set_max_extent/SetMaxExtentSample.java
```diff
@@ -0,0 +1,142 @@
...
+import com.esri.arcgisruntime.geometry.Envelope;
+import com.esri.arcgisruntime.geometry.Point;
+import com.esri.arcgisruntime.geometry.SpatialReferences;
+import com.esri.arcgisruntime.mapping.view.Graphic;
+import com.esri.arcgisruntime.mapping.view.GraphicsOverlay;
+import com.esri.arcgisruntime.symbology.*;
+import com.esri.arcgisruntime.ArcGISRuntimeEnvironment;
+import com.esri.arcgisruntime.mapping.ArcGISMap;
+import com.esri.arcgisruntime.mapping.BasemapStyle;
+import com.esri.arcgisruntime.mapping.view.MapView;
```
Rachael-E: good ordering with these statements: we always separate the Runtime Import statements out from the Java or JavaFX ones for readability.

For imports which auto collapse with a *, we prefer to show all the imports. If you hover over the star in IntelliJ, you'll see an orange lightbulb tooltip. Open that up and it will give you the option to replace with single class imports. This would be the case for both JavaFX, and Runtime import statements here.

image

map/set-max-extent/src/main/java/com.esri.samples.set_max_extent/SetMaxExtentSample.java
```diff
@@ -0,0 +1,142 @@
...
+
+public class SetMaxExtentSample extends Application {
+
+    private MapView mapView;
+
+
```
Rachael-E: remove extra white space here

map/set-max-extent/src/main/java/com.esri.samples.set_max_extent/SetMaxExtentSample.java
```diff
@@ -0,0 +1,142 @@
...
+    @Override
+    public void start(Stage stage) {
+
+        try {
+            // create stack pane and application scene
+            var stackPane = new StackPane();
+            var scene = new Scene(stackPane);
```
Rachael-E: these are good examples to use `var` for. It's very clear from the class instantiation that it is a `StackPane`, and the variable name is `stackPane`, so a user is unlikely to need to scroll back up to the top of the code to see the full instantiated class here. it's clear from the variable name.

map/set-max-extent/src/main/java/com.esri.samples.set_max_extent/SetMaxExtentSample.java
```diff
@@ -0,0 +1,142 @@
...
+            stage.setHeight(700);
+            stage.setScene(scene);
+            stage.show();
+
+            // authentication with an API key or named user is required to access basemaps and other location services
+            String yourAPIKey = System.getProperty("apiKey");
+            ArcGISRuntimeEnvironment.setApiKey(yourAPIKey);
+
+            // create a map with the streets basemap style
+            var map = new ArcGISMap(BasemapStyle.ARCGIS_STREETS);
```
Rachael-E: I tend to declare the explicit type with `ArcGISMap`. 

`map` can have many meanings in Java, and for Java developers, a geographical map may not necessarily be the first thing they think of when seeing a variable name later down the code! So for the avoidance of all doubt, we keep `ArcGISMap` declared and avoid the use of var here.

Another option would be to name the variable `arcGISMap` but it's not quite as readable as map later down the code. 

```suggestion
            ArcGISMap map = new ArcGISMap(BasemapStyle.ARCGIS_STREETS);
```

map/set-max-extent/src/main/java/com.esri.samples.set_max_extent/SetMaxExtentSample.java
```diff
@@ -0,0 +1,142 @@
...
+
+            // create a map view and set the map to it
+            mapView = new MapView();
+            mapView.setMap(map);
+
+            // create a max extent envelope with points on Colorado's northwest and southeast corners
+            var coloradoNW = new Point(-12139393.2109, 5012444.0468, SpatialReferences.getWebMercator());
+            var coloradoSE = new Point(-11359277.5124, 4438148.7816, SpatialReferences.getWebMercator());
```
Rachael-E: this is a good example of where `var` is best avoided. Imagine you are coming into this code fresh, and you see `coloradoNW` coded. It wouldn't be instantly obvious that this was a point: it could be graphic, a geometry, or a symbol. 

Our customers can be anywhere from complete beginners with coding or GIS, through to seasoned developers/GIS users. To ensure we cater for all users, we always try to go with the most readable variable names possible to save users scrolling around to find the instantiation.

If this were a straight forward `Point point = new Point` then that would be a case for var. But because the variable name has to be a bit more detailed for readability, better to use the explicit declaration here.

So here we could have instead:

```suggestion
            Point coloradoNorthWestPoint = new Point(-12139393.2109, 5012444.0468, SpatialReferences.getWebMercator());
            Point coloradoSouthEastPoint = new Point(-11359277.5124, 4438148.7816, SpatialReferences.getWebMercator());
```

map/set-max-extent/src/main/java/com.esri.samples.set_max_extent/SetMaxExtentSample.java
```diff
@@ -0,0 +1,142 @@
...
+            var extentEnvelope = new Envelope(coloradoNW, coloradoSE);
```
Rachael-E: here it's ok to just go with `envelope` as a variable name: this is the only Envelope in the sample so it doesn't need a unique name to define it. 
```suggestion
            var envelope = new Envelope(coloradoNorthWestPoint, coloradoSouthEastPoint);
```

map/set-max-extent/src/main/java/com.esri.samples.set_max_extent/SetMaxExtentSample.java
```diff
@@ -0,0 +1,142 @@
...
+
+            // set the map's max extent to the max extent envelope
```
Rachael-E: let's add an additional comment in here to make it clear to a new user it's this bit of code that is doing the constraining 
```suggestion
            // constrain the display of the map to the state borders of Colorado by setting the map's max extent to the envelope
```

map/set-max-extent/src/main/java/com.esri.samples.set_max_extent/SetMaxExtentSample.java
```diff
@@ -0,0 +1,142 @@
...
+            ArcGISRuntimeEnvironment.setApiKey(yourAPIKey);
+
+            // create a map with the streets basemap style
+            var map = new ArcGISMap(BasemapStyle.ARCGIS_STREETS);
+
+            // create a map view and set the map to it
+            mapView = new MapView();
+            mapView.setMap(map);
+
+            // create a max extent envelope with points on Colorado's northwest and southeast corners
```
Rachael-E:
```suggestion
            // create an envelope with an extent covering the state of Colorado
```

map/set-max-extent/src/main/java/com.esri.samples.set_max_extent/SetMaxExtentSample.java
```diff
@@ -0,0 +1,142 @@
...
+            var coloradoNW = new Point(-12139393.2109, 5012444.0468, SpatialReferences.getWebMercator());
+            var coloradoSE = new Point(-11359277.5124, 4438148.7816, SpatialReferences.getWebMercator());
+            var extentEnvelope = new Envelope(coloradoNW, coloradoSE);
+
+            // set the map's max extent to the max extent envelope
+            map.setMaxExtent(extentEnvelope);
+
+            // create a graphics overlay of the map's max extent and add it to the map view
+            var coloradoGraphicsOverlay = new GraphicsOverlay();
```
Rachael-E: as this is the only graphics overlay in this sample, we can keep var and simply name it

```suggestion
            var graphicsOverlay = new GraphicsOverlay();
```

map/set-max-extent/src/main/java/com.esri.samples.set_max_extent/SetMaxExtentSample.java
```diff
@@ -0,0 +1,142 @@
...
+
+            // create a max extent envelope with points on Colorado's northwest and southeast corners
+            var coloradoNW = new Point(-12139393.2109, 5012444.0468, SpatialReferences.getWebMercator());
+            var coloradoSE = new Point(-11359277.5124, 4438148.7816, SpatialReferences.getWebMercator());
+            var extentEnvelope = new Envelope(coloradoNW, coloradoSE);
+
+            // set the map's max extent to the max extent envelope
+            map.setMaxExtent(extentEnvelope);
+
+            // create a graphics overlay of the map's max extent and add it to the map view
```
Rachael-E: this isn't strictly what this code is doing: we have to break it up a bit more
```suggestion
            // create a new graphics overlay and add a new graphic to it that shows the Colorado border as a red dashed line
```

map/set-max-extent/src/main/java/com.esri.samples.set_max_extent/SetMaxExtentSample.java
```diff
@@ -0,0 +1,142 @@
...
+            var coloradoGraphicsOverlay = new GraphicsOverlay();
+            coloradoGraphicsOverlay.getGraphics().add(new Graphic(map.getMaxExtent()));
+            mapView.getGraphicsOverlays().add(coloradoGraphicsOverlay);
+
+            // create a simple renderer with a red dashed line and set it to be the renderer for the graphics overlay
+            var simpleRenderer = new SimpleRenderer(new SimpleLineSymbol(SimpleLineSymbol.Style.DASH, ColorUtil.colorToArgb(Color.RED), 5));
+            coloradoGraphicsOverlay.setRenderer(simpleRenderer);
```
Rachael-E: Let's move this section between the adding of the new graphic and the adding the graphics overlay to the map view. That way it keeps all the code related to the graphics overlay in one neat chunk

map/set-max-extent/src/main/java/com.esri.samples.set_max_extent/SetMaxExtentSample.java
```diff
@@ -0,0 +1,142 @@
...
+
+            // set the map's max extent to the max extent envelope
+            map.setMaxExtent(extentEnvelope);
+
+            // create a graphics overlay of the map's max extent and add it to the map view
+            var coloradoGraphicsOverlay = new GraphicsOverlay();
+            coloradoGraphicsOverlay.getGraphics().add(new Graphic(map.getMaxExtent()));
+            mapView.getGraphicsOverlays().add(coloradoGraphicsOverlay);
+
+            // create a simple renderer with a red dashed line and set it to be the renderer for the graphics overlay
```
Rachael-E:
```suggestion
            // create a simple renderer with a red dashed styled symbol, and set it as the renderer for the graphics overlay
```

map/set-max-extent/src/main/java/com.esri.samples.set_max_extent/SetMaxExtentSample.java
```diff
@@ -0,0 +1,142 @@
...
+            mapView = new MapView();
+            mapView.setMap(map);
+
+            // create a max extent envelope with points on Colorado's northwest and southeast corners
+            var coloradoNW = new Point(-12139393.2109, 5012444.0468, SpatialReferences.getWebMercator());
+            var coloradoSE = new Point(-11359277.5124, 4438148.7816, SpatialReferences.getWebMercator());
+            var extentEnvelope = new Envelope(coloradoNW, coloradoSE);
+
+            // set the map's max extent to the max extent envelope
+            map.setMaxExtent(extentEnvelope);
```
Rachael-E: with the Java Runtime API, it's best practice to listen for the map loading to finish, and if it loaded successfully, to then carry out map related tasks. We can move this into an addDoneLoadingListener block as follows (with a bonus setting the viewpoint of the map view: another common workflow). It also allows us to display a message if the map fails to load: if the map fails to load then there's no need to set its max extent! 

```suggestion
      // listen for the map to finish loading, and check it has loaded
      map.addDoneLoadingListener(() -> {
        if (map.getLoadStatus() == LoadStatus.LOADED) {
          // constrain the display of the map to the state borders of Colorado by setting the map's max extent to the envelope
          map.setMaxExtent(envelope);
          // set the map view's viewpoint with the envelope
          mapView.setViewpoint(new Viewpoint(envelope));
        } else if (map.getLoadStatus() == LoadStatus.FAILED_TO_LOAD) {
          new Alert(Alert.AlertType.ERROR, "Map failed to load").show();
        }
      });
```

map/set-max-extent/src/main/java/com.esri.samples.set_max_extent/SetMaxExtentSample.java
```diff
@@ -0,0 +1,142 @@
...
+
+            // create a graphics overlay of the map's max extent and add it to the map view
+            var coloradoGraphicsOverlay = new GraphicsOverlay();
+            coloradoGraphicsOverlay.getGraphics().add(new Graphic(map.getMaxExtent()));
```
Rachael-E: it would be preferable if the new graphic was created from the envelope here: using map.getMaxExtent in a simple sample like this one is unlikely to cause any async issues, but a bigger app might (the setting of the map's max extent might not have completed before the new graphic is created). Let's use the envelope instead which has been instantiated on the same thread above and is safer:

```suggestion
            coloradoGraphicsOverlay.getGraphics().add(new Graphic(envelope));
```

map/set-max-extent/src/main/java/com.esri.samples.set_max_extent/SetMaxExtentSample.java
Rachael-E: we can move all of this graphics related code below the envelope instantiation, and before the addDoneLoadingListener. That way the set up is all taken care of first, and then finally once the map has loaded, we can set the max extent.

map/set-max-extent/src/main/java/com.esri.samples.set_max_extent/SetMaxExtentSample.java
```diff
@@ -0,0 +1,142 @@
...
+            mapView.getGraphicsOverlays().add(coloradoGraphicsOverlay);
+
+            // create a simple renderer with a red dashed line and set it to be the renderer for the graphics overlay
+            var simpleRenderer = new SimpleRenderer(new SimpleLineSymbol(SimpleLineSymbol.Style.DASH, ColorUtil.colorToArgb(Color.RED), 5));
+            coloradoGraphicsOverlay.setRenderer(simpleRenderer);
+
+            // create a control panel for the on/off toggles
```
Rachael-E:
```suggestion
            // create a control panel for toggling the max extent setting
```

map/set-max-extent/src/main/java/com.esri.samples.set_max_extent/SetMaxExtentSample.java
```diff
@@ -0,0 +1,142 @@
...
+                    Insets.EMPTY)));
+            controlsVBox.setPadding(new Insets(10.0));
+            controlsVBox.setMaxSize(120, 75);
+
+            // create a toggle group for the enabled/disable buttons
+            var toggleGroup = new ToggleGroup();
+
+            // create the enable/disable buttons and add them to the toggle group
+            var toggleOn = new ToggleButton("Max Extent ON ");
+            var toggleOff = new ToggleButton("Max Extent OFF");
```
Rachael-E: There are a couple of gotchas with this toggle button approach that we would have to address. Given the code complexity we already have here, I'm loathe to suggest we add more code. On reflection I think going for a check box is going to be the better route here. Nonetheless this has been good experience for you to get to grips with a range of JavaFX UI components!

Rachael-E (changes requested):
@01smito01 really great job on your first sample. Great attention to detail, and neat coding throughout, you are already following our coding standards naturally! 

There are a few things to address here: I'll leave these with you to have a look over, address, and then we'll meet to go over these together. You've come across a number of good workflow examples in this sample that will be good to share on a call with yourself and @Luiy0 .

map/set-max-extent/README.metadata.json
01smito01: Originally this used forward slash but that caused the metadata checker to fail because it wanted the version with dots. Assume it's fine to ignore it in this case?

Rachael-E: that's interesting you saw a message for that: we'll want to update that in the script. This looks like this is passing, and matches the other json format.

01smito01:
(Hopefully) fixed the issue with the metadata/directory structure, unfortunately it's made the diff a bit messy.

map/set-max-extent/README.md
```diff
@@ -0,0 +1,33 @@
...
+that you won't be able to pan far from the Colorado border or zoom out beyond the minimum scale set
+by the max extent. Use the "Max Extent OFF" button to disable the max extent to freely pan/zoom around the map.
+
+## How it works
+
+1. Create an `ArcGISMap` object.
+2. Create an envelope of the extent using `Envelope(Point(x, y), Point(x, y))`
+3. Set the maximum extent of the map with `map.setMaxExtent(envelope)`.
+4. Set the map to a `MapView` object.
+5. Set `map.setMaxExtent(null)` to disable the maximum extent of the map
```
Rachael-E: for java let's avoid directly referring to object in the readme, this is the approach the doc team take for web content. A couple of other edits here too:

```suggestion
1. Create a new `ArcGISMap`.
2. Create a new `Envelope` with the required max extent.
3. Wait for the map to load then set the maximum extent of the map with `map.setMaxExtent(envelope)`.
4. Create a new `MapView` and set the map to it.
5. Set `map.setMaxExtent(null)` to disable the maximum extent of the map.
```

map/set-max-extent/src/main/java/com/esri/samples/set_max_extent/SetMaxExtentSample.java
```diff
@@ -0,0 +1,162 @@
...
+      // create a map with the streets basemap style
+      ArcGISMap map = new ArcGISMap(BasemapStyle.ARCGIS_STREETS);
+
+      // create a map view and set the map to it
+      mapView = new MapView();
+      mapView.setMap(map);
+
+      // create an envelope with an extent covering the state of Colorado
+      Point coloradoNorthWestPoint = new Point(-12139393.2109, 5012444.0468, SpatialReferences.getWebMercator());
+      Point coloradoSouthEastPoint = new Point(-11359277.5124, 4438148.7816, SpatialReferences.getWebMercator());
```
Rachael-E: we can remove the spatial references constructor here since the basemap is already in Web Mercator: no need to set it for this instance. 

```suggestion
      Point coloradoNorthWestPoint = new Point(-12139393.2109, 5012444.0468);
      Point coloradoSouthEastPoint = new Point(-11359277.5124, 4438148.7816);
```

map/set-max-extent/src/main/java/com/esri/samples/set_max_extent/SetMaxExtentSample.java
```diff
@@ -0,0 +1,162 @@
...
+      var envelope = new Envelope(coloradoNorthWestPoint, coloradoSouthEastPoint);
+
+      // create a new graphics overlay and add a new graphic to it that shows the Colorado border as a red dashed line
+      var graphicsOverlay = new GraphicsOverlay();
+      graphicsOverlay.getGraphics().add(new Graphic(envelope));
+      var simpleRenderer = new SimpleRenderer(new SimpleLineSymbol(SimpleLineSymbol.Style.DASH,
+        ColorUtil.colorToArgb(Color.RED), 5));
```
Rachael-E: good to break up this line so it's not extending too long. For further readability, let's break it up where we're setting up the new SimpleLineSymbol

 ```suggestion
      var simpleRenderer = new SimpleRenderer(
        new SimpleLineSymbol(SimpleLineSymbol.Style.DASH, ColorUtil.colorToArgb(Color.RED), 5));
```

File: map/set-max-extent/src/main/java/com/esri/samples/set_max_extent/SetMaxExtentSample.java [hunk 23]
```diff
@@ -0,0 +1,162 @@
...
+
+      // listen for the map to finish loading, and check it has loaded
+      map.addDoneLoadingListener(() -> {
+        if (map.getLoadStatus() == LoadStatus.LOADED) {
+          // constrain the display of the map to the state borders of Colorado by setting the map's max extent to the
+          // envelope
+          map.setMaxExtent(envelope);
+          // set the map view's viewpoint with the envelope
+          mapView.setViewpoint(new Viewpoint(envelope));
+          //enable the checkbox now that the map has loaded
```
Rachael-E:
```suggestion
          // enable the checkbox now that the map has loaded
```

map/set-max-extent/src/main/java/com/esri/samples/set_max_extent/SetMaxExtentSample.java
```diff
@@ -0,0 +1,162 @@
...
+      });
+
+      //add the checkbox to the control panel
+      controlsVBox.getChildren().add(checkbox);
+
+      // listen for the map to finish loading, and check it has loaded
+      map.addDoneLoadingListener(() -> {
+        if (map.getLoadStatus() == LoadStatus.LOADED) {
+          // constrain the display of the map to the state borders of Colorado by setting the map's max extent to the
+          // envelope
```
Rachael-E:
```suggestion
          // constrain the display of the map to the borders of Colorado by setting the map's max extent to the envelope
```

map/set-max-extent/src/main/java/com/esri/samples/set_max_extent/SetMaxExtentSample.java
```diff
@@ -0,0 +1,162 @@
...
+      // add the graphics overlay to the map view
+      mapView.getGraphicsOverlays().add(graphicsOverlay);
+
+      // create a control panel for toggling the max extent setting
+      var controlsVBox = new VBox(6);
+      controlsVBox.setBackground(new Background(new BackgroundFill(Paint.valueOf("rgba(255,255,255,0.5)"),
+        CornerRadii.EMPTY, Insets.EMPTY)));
+      controlsVBox.setPadding(new Insets(10.0));
+      controlsVBox.setMaxSize(145, 40);
+      controlsVBox.setDisable(true);
```
Rachael-E: let's move this up before the setting API code is called, and after the stage set up. That keeps this out the way of the main code, and set up ready for use with the checkbox later.

map/set-max-extent/src/main/java/com/esri/samples/set_max_extent/SetMaxExtentSample.java
```diff
@@ -0,0 +1,162 @@
...
+      var simpleRenderer = new SimpleRenderer(new SimpleLineSymbol(SimpleLineSymbol.Style.DASH,
+        ColorUtil.colorToArgb(Color.RED), 5));
+      graphicsOverlay.setRenderer(simpleRenderer);
+
+      // add the graphics overlay to the map view
+      mapView.getGraphicsOverlays().add(graphicsOverlay);
+
+      // create a control panel for toggling the max extent setting
+      var controlsVBox = new VBox(6);
+      controlsVBox.setBackground(new Background(new BackgroundFill(Paint.valueOf("rgba(255,255,255,0.5)"),
```
Rachael-E: I do like this white background with black label: it's more modern. We will be modernising samples but for now for keeping samples consistent, let's stick with the dark grey background and white label text

`checkbox.setTextFill(Color.WHITE);`

```suggestion
      controlsVBox.setBackground(new Background(new BackgroundFill(Paint.valueOf("rgba(0,0,0,0.5)"),
```

map/set-max-extent/src/main/java/com/esri/samples/set_max_extent/SetMaxExtentSample.java
```diff
@@ -0,0 +1,162 @@
...
+        CornerRadii.EMPTY, Insets.EMPTY)));
+      controlsVBox.setPadding(new Insets(10.0));
+      controlsVBox.setMaxSize(145, 40);
+      controlsVBox.setDisable(true);
+
+      // create a checkbox for toggling max extent
+      var checkbox = new CheckBox("Enable Max Extent");
+      checkbox.setSelected(true);
```
Rachael-E:
```suggestion
      var checkbox = new CheckBox("Enable Max Extent");
      checkbox.setTextFill(Color.WHITE);
      checkbox.setSelected(true);
```

map/set-max-extent/src/main/java/com/esri/samples/set_max_extent/SetMaxExtentSample.java
```diff
@@ -0,0 +1,162 @@
...
+      checkbox.setOnMouseClicked(e -> {
+        if (checkbox.isSelected()) {
+          map.setMaxExtent(envelope);
+        } else {
+          map.setMaxExtent(null);
+        }
+      });
```
Rachael-E, line 103: this is fine as it is, it's clear what event is triggering the max extent. There is another alternative way to handle this that tidies up the code even further but will leave it up to you which syntax you prefer! 

```java
      // if the checkbox is checked, set the envelope as the max extent. if it is unchecked, set the max extent as null
      checkbox.setOnMouseClicked(e -> 
        map.setMaxExtent(checkbox.isSelected() ? envelope : null)
      );
```

map/set-max-extent/src/main/java/com/esri/samples/set_max_extent/SetMaxExtentSample.java
```diff
@@ -0,0 +1,162 @@
...
+
+      //add the checkbox to the control panel
+      controlsVBox.getChildren().add(checkbox);
```
Rachael-E: let's move this to the top of the code block where the adding of the mapview and vbox to the stackpane happens

```suggestion
      // add the checkbox to the control panel
      controlsVBox.getChildren().add(checkbox);
```

Rachael-E (commented):
thanks @01smito01, just a couple more tidy up suggestions here and you're ready for a second review. The new checkbox looks good, and the code is reading more clearly. Nice work.

map/set-max-extent/src/main/java/com/esri/samples/set_max_extent/SetMaxExtentSample.java
01smito01: I think in terms of legibility I prefer it as is - appreciate the conciseness of your suggestion but the syntax is IMO fairly opaque - I get what it's doing but I think it would be a bit unfriendly for beginners.

01smito01:
@Rachael-E that should be everything changed and committed now.

Rachael-E (approved):
thanks @01smito01 , these changes look great! this is ready for a second review now from @jenmerritt .

map/set-max-extent/src/main/java/com/esri/samples/set_max_extent/SetMaxExtentLauncher.java
```diff
@@ -0,0 +1,11 @@
...
+
+/**
+ * Wrapper required for launching a JavaFX 11 app through Gradle or from a jar.
+ */
+public class SetMaxExtentLauncher {
+
+  public static void main(String[] args) {
+    SetMaxExtentSample.main(args);
+  }
+}
```
jenmerritt: Best to have a blank line at the end of files

map/set-max-extent/src/main/java/com/esri/samples/set_max_extent/SetMaxExtentSample.java
```diff
@@ -0,0 +1,161 @@
...
+      var controlsVBox = new VBox(6);
+      controlsVBox.setBackground(new Background(new BackgroundFill(Paint.valueOf("rgba(0,0,0,0.5)"),
+        CornerRadii.EMPTY, Insets.EMPTY)));
+      controlsVBox.setPadding(new Insets(10.0));
+      controlsVBox.setMaxSize(145, 40);
+      controlsVBox.setDisable(true);
+
+      // authentication with an API key or named user is required to access basemaps and other location services
+      String yourAPIKey = System.getProperty("apiKey");
+      ArcGISRuntimeEnvironment.setApiKey(yourAPIKey);
```
jenmerritt:
```suggestion
      // authentication with an API key or named user is required to access basemaps and other location services
      String yourAPIKey = System.getProperty("apiKey");
      ArcGISRuntimeEnvironment.setApiKey(yourAPIKey);
      
      // create a control panel for toggling the max extent setting
      var controlsVBox = new VBox(6);
      controlsVBox.setBackground(new Background(new BackgroundFill(Paint.valueOf("rgba(0,0,0,0.5)"),
        CornerRadii.EMPTY, Insets.EMPTY)));
      controlsVBox.setPadding(new Insets(10.0));
      controlsVBox.setMaxSize(145, 40);
      controlsVBox.setDisable(true);
```

I'd move the API key setting to be the next thing after setting up the stage, keeps it consistent with the pattern on most other samples.

map/set-max-extent/src/main/java/com/esri/samples/set_max_extent/SetMaxExtentSample.java
jenmerritt: I actually went through the same process here and had this very conversation with myself! I agree it's tidier code but also agree with your decision to leave as is @01smito01 in terms of legibility. This is one of those where it's a balance and there's no right answer depending on which perspective you take.

map/set-max-extent/src/main/java/com/esri/samples/set_max_extent/SetMaxExtentSample.java
```diff
@@ -0,0 +1,161 @@
...
+      checkbox.setTextFill(Color.WHITE);
+      checkbox.setSelected(true);
+      checkbox.setOnMouseClicked(e -> {
+        if (checkbox.isSelected()) {
+          map.setMaxExtent(envelope);
+        } else {
+          map.setMaxExtent(null);
+        }
+      });
+
```
jenmerritt, line 104: I personally would put the block that instantiates the VBox here. So:

```
      // create a control panel for toggling the max extent setting
      var controlsVBox = new VBox(6);
      controlsVBox.setBackground(new Background(new BackgroundFill(Paint.valueOf("rgba(0,0,0,0.5)"),
      controlsVBox.setMaxSize(145, 40);
      controlsVBox.setDisable(true);
```

followed by:
```
// add the checkbox to the control panel
      controlsVBox.getChildren().add(checkbox);
```

map/set-max-extent/src/main/java/com/esri/samples/set_max_extent/SetMaxExtentSample.java
```diff
@@ -0,0 +1,161 @@
...
+      var scene = new Scene(stackPane);
+
+      // set title, size, and add scene to stage
+      stage.setTitle("Set Max Extent Sample");
+      stage.setWidth(800);
+      stage.setHeight(700);
+      stage.setScene(scene);
+      stage.show();
+
+      // create a control panel for toggling the max extent setting
```
jenmerritt:
```suggestion
      // create a control panel
```

Would argue the checkbox does the toggling

map/set-max-extent/src/main/java/com/esri/samples/set_max_extent/SetMaxExtentSample.java
```diff
@@ -0,0 +1,161 @@
...
+      controlsVBox.getChildren().add(checkbox);
+
+      // listen for the map to finish loading, and check it has loaded
+      map.addDoneLoadingListener(() -> {
+        if (map.getLoadStatus() == LoadStatus.LOADED) {
+          // constrain the display of the map to the borders of Colorado by setting the map's max extent to the envelope
+          map.setMaxExtent(envelope);
+          // set the map view's viewpoint with the envelope
+          mapView.setViewpoint(new Viewpoint(envelope));
+          // enable the checkbox now that the map has loaded
```
jenmerritt:
```suggestion
          // enable the UI now that the map has loaded
```

jenmerritt (approved):
Great work @01smito01 and congrats on the first sample! I've made a few suggestions but I'm approving the PR as the suggestions are more stylistic. From my experience it also helps to just show you some thought processes as well and stimulate any discussion. Feel free to reach out or re-request a review if you want me to recheck anything.

01smito01:
Thanks for the review @jenmerritt, I've implemented the changes now. Assume it's fine to merge now?

jenmerritt:
Good to merge @01smito01 - use the "squash and merge" option, which is the default.

PR #692, opened 2022-07-12
[New Sample] Apply Unique Values With Alternate Symbols

Luiy0:
### Description

This sample showcases how to apply a unique value with alternate symbols depending on different scale range values. 

The feature layer variable is not local to avoid out of scope issues.  Is the use of `List.of` encouraged for creating unique values or should the Iterable container be created in a formal way?  

Additionally, to improve code readability, should a method be created to move the creation of symbols from the start method. Please feel free to comment on variable naming conventions and other general coding practices.

Checklist:
- [x] Request a reviewer
- [x] Assign a reviewer
- [x] Tag reviewer in comment


symbology/apply-unique-values-with-alternate-symbols/README.md
```diff
@@ -0,0 +1,50 @@
+# Apply unique values with alternate symbols
+
+Apply a unique value with alternate symbols at different scales.
+
+![ApplyUniqueValuesWithAlternateSymbols](ApplyUniqueValuesWithAlternateSymbols.gif)
```
Rachael-E: Let's stick with the .png for this sample. It'll keep the file size down (9mb is hefty for a gif), and I don't think we need it for what this sample is showing.

The sample design will mostly have a gif so that other platforms can see how the sample should run without having have the sample live in front of you: it'll be rare we do include a gif in a sample. For example I tend to only add a gif if there is a very neat visualisation, e.g. the Animate 3D Graphic sample or Animate Images with Image Overlay.

symbology/apply-unique-values-with-alternate-symbols/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Use case
+
+When a layer is symbolized with unique value symbology, you can specify the visible scale range for each unique value. This is an effective strategy to limit the amount of detailed data at smaller scales without having to make multiple versions of the layer, each with a unique definition query.
+
+Once scale ranges are applied to unique values, you can further refine the appearance of features within those scale ranges by establishing alternate symbols to different parts of the symbol class scale range.
+
+## How to use the sample
+
+Zoom in and out of the map to see alternate symbols at each scale. The symbology changes according to the following scale ranges: 0-5000, 5000-10000, 10000-20000. To go back to the initial viewpoint, tap "Reset Viewpoint".
```
Rachael-E: This is totally a nitpick on the design text, but we should use the following notation when referring to scales in plain english (as supposed to using the actual value in code): see https://developers.arcgis.com/documentation/mapping-apis-and-services/reference/zoom-levels-and-scale/

```suggestion
Zoom in and out of the map to see alternate symbols at each scale. The symbology changes according to the following scale ranges: 1:0-1:5000, 1:5000-1:10000, 1:10000-1:20000. To go back to the initial viewpoint, tap "Reset Viewpoint".
```

Rachael-E: this happens quite frequently: with samples created for both mobile and desktop versions, the design will often be either mobile or desktop detailed. in this case it was mobile, so we can update this from tap to click for our desktop Java based samples

```suggestion
Zoom in and out of the map to see alternate symbols at each scale. The symbology changes according to the following scale ranges: 0-5000, 5000-10000, 10000-20000. To go back to the initial viewpoint, click "Reset Viewpoint".
```

symbology/apply-unique-values-with-alternate-symbols/README.md
```diff
@@ -0,0 +1,50 @@
...
+## How it works
+
+1. Create a `FeatureLayer` using the service url and add it to the map's list of operational layers.
+2. Create two alternate symbols (a blue square and a yellow diamond) to be used as alternate symbols. To create an alternate symbol:
+
+   a. Create a symbol using `SimpleMarkerSymbol`.
+
+   b. Convert the simple marker symbol to a `MultilayerPointSymbol` using `SimpleMarkerSymbol.toMultilayerSymbol()`.
+
+   c. Set the valid scale range through reference properties on the multilayer point symbols blue square and yellow diamond by calling `multilayerPointSymbol.referenceProperties = SymbolReferenceProperties(double minScale, double maxScale)`.
```
Rachael-E: can remove excess white space here, and let's keep the code references Java specific. We don't need specific details in the How it Works (the idea being a user can use this in their own application with their own data/code) so I've also removed the mention to the square and diamond here

```suggestion
   a. Create a symbol using `SimpleMarkerSymbol`.
   b. Convert the simple marker symbol to a `MultilayerPointSymbol` using `SimpleMarkerSymbol.toMultilayerSymbol()`.
   c. Set the valid scale range through reference properties on the multilayer point symbols by calling `multilayerPointSymbol.setReferenceProperties(new SymbolReferenceProperties(double minScale, double maxScale))`.
```

symbology/apply-unique-values-with-alternate-symbols/README.md
```diff
@@ -0,0 +1,50 @@
...
+Once scale ranges are applied to unique values, you can further refine the appearance of features within those scale ranges by establishing alternate symbols to different parts of the symbol class scale range.
+
+## How to use the sample
+
+Zoom in and out of the map to see alternate symbols at each scale. The symbology changes according to the following scale ranges: 0-5000, 5000-10000, 10000-20000. To go back to the initial viewpoint, tap "Reset Viewpoint".
+
+## How it works
+
+1. Create a `FeatureLayer` using the service url and add it to the map's list of operational layers.
+2. Create two alternate symbols (a blue square and a yellow diamond) to be used as alternate symbols. To create an alternate symbol:
```
Rachael-E:
```suggestion
2. Create two alternate symbols:
```

symbology/apply-unique-values-with-alternate-symbols/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+   a. Create a symbol using `SimpleMarkerSymbol`.
+
+   b. Convert the simple marker symbol to a `MultilayerPointSymbol` using `SimpleMarkerSymbol.toMultilayerSymbol()`.
+
+   c. Set the valid scale range through reference properties on the multilayer point symbols blue square and yellow diamond by calling `multilayerPointSymbol.referenceProperties = SymbolReferenceProperties(double minScale, double maxScale)`.
+
+3. Create a third multilayer symbol to be used to create a `UniqueValue` class.
+4. Create a unique value using the red triangle and the list of alternate symbols.
+5. Create a `UniqueValueRenderer` and add the unique value.
```
Rachael-E: the design has a little too much specific detail here: for example, reading this, I'm not clear what the 3rd multilayer symbol is or why it's referenced here. We want to capture the main workflow here, not the fine details. 

We can cut that down and simplify it a bit with:

```suggestion
3. Create a new `UniqueValue`, and add it to a new `UniqueValueRenderer` with uniqueValueRenderer.getUniqueValues.add(uniqueValue)` 
```

symbology/apply-unique-values-with-alternate-symbols/README.md
```diff
@@ -0,0 +1,50 @@
...
+6. Create a purple diamond simple marker and convert it to a multilayer symbol to be used as the default symbol.
```
Rachael-E: this step has already been covered in point 2 as to how to convert a multilayer symbol. let's focus instead on how to set the default symbol as follows in the step below. 

```suggestion
```

symbology/apply-unique-values-with-alternate-symbols/README.md
```diff
@@ -0,0 +1,50 @@
...
+7. Set the default symbol on the unique value renderer to the purple diamond using `uniqueValueRenderer.defaultSymbol`.
```
Rachael-E:
```suggestion
7. Set the default symbol on the unique value renderer to a multilayer point symbol with `uniqueValueRenderer.setDefaultSymbol(multilayerPointSymbol)`.
```

symbology/apply-unique-values-with-alternate-symbols/README.md
```diff
@@ -0,0 +1,50 @@
...
+8. Set the `uniqueValueRenderer.fieldNames` on the unique value renderer to "req_type".
```
Rachael-E:
```suggestion
8. Set the field name on the unique value renderer with `uniqueValueRenderer.getFieldNames().add(("req_type"))`.
```

symbology/apply-unique-values-with-alternate-symbols/README.md
```diff
@@ -0,0 +1,50 @@
...
+9. Apply this unique value renderer to the `featureLayer.renderer`.
```
Rachael-E:
```suggestion
9. Apply this unique value renderer to the feature layer with `featureLayer.setRenderer(uniqueValueRenderer)`.
```

symbology/apply-unique-values-with-alternate-symbols/README.metadata.json
```diff
@@ -0,0 +1,36 @@
...
+        "unique value renderer",
+        "MultilayerPointSymbol",
+        "SimpleMarkerSymbol",
+        "SymbolReferenceProperties",
+        "UniqueValue",
+        "UniqueValueRenderer"
+    ],
+    "redirect_from": [
+        ""
+    ],
```
Rachael-E: this is one area with the metadata json that we run into issues with, so it's good you've come across this on your first sample! Basically, if there is no redirect, we must use the following notation i.e. an empty string within an array will not be accepted and the webpage will crash on loading.

```suggestion
    "redirect_from": "",
```

this has happened a couple of times, so when we updated our Github actions, we put in a check for this. If you have a look at the Readme check, and scroll down to the bottom, you'll see there a warning and a message as to why this check on the metadata json has failed.

https://github.com/Esri/arcgis-runtime-samples-java/pull/692/checks

symbology/apply-unique-values-with-alternate-symbols/src/main/java/com/esri/samples/apply_unique_values_with_alternate_symbols/ApplyUniqueValuesWithAlternateSymbolsSample.java
```diff
@@ -0,0 +1,231 @@
...
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package com.esri.samples.apply_unique_values_with_alternate_symbols;
+
+import java.util.List;
+
+import com.esri.arcgisruntime.loadable.LoadStatus;
```
Rachael-E: let's move this import down so that all of the runtime import statements are together. 

IntelliJ does have some configurable settings to handle auto import organising/ordering, but I've found it tends to stick them back up on top and not in a group so I usually order them manually otherwise a couple of imports in the wrong place slip through!

symbology/apply-unique-values-with-alternate-symbols/src/main/java/com/esri/samples/apply_unique_values_with_alternate_symbols/ApplyUniqueValuesWithAlternateSymbolsSample.java
```diff
@@ -0,0 +1,231 @@
...
+public class ApplyUniqueValuesWithAlternateSymbolsSample extends Application{
+
+  private MapView mapView;
+
+  private FeatureLayer featureLayer;
+
+  private final static String FEATURE_SERVICE_URL = "https://sampleserver6.arcgisonline.com/arcgis/rest/services/SF311/FeatureServer/0";
+
+  private Button btnReset;
+
```
Rachael-E: we can remove the white space between these member variables here
```suggestion

  private FeatureLayer featureLayer;
  private final static String FEATURE_SERVICE_URL = "https://sampleserver6.arcgisonline.com/arcgis/rest/services/SF311/FeatureServer/0";
  private Button btnReset;
```

symbology/apply-unique-values-with-alternate-symbols/src/main/java/com/esri/samples/apply_unique_values_with_alternate_symbols/ApplyUniqueValuesWithAlternateSymbolsSample.java
```diff
@@ -0,0 +1,231 @@
...
+      stage.setHeight(700);
+      stage.setScene(scene);
+      stage.show();
+
+      // authentication with an API key or named user is required to access base maps and other location services
+      String yourAPIKey = System.getProperty("apiKey");
+      ArcGISRuntimeEnvironment.setApiKey(yourAPIKey);
+
+      // create a point located at San Francisco, CA to be used as the viewpoint for the map
+      Point vPoint = new Point(-13631205.660131, 4546829.846004, SpatialReferences.getWebMercator());
```
Rachael-E: let's move this point instantiation to just above where we first use it: above the setting of the map view view point.

Also, since this is the only Point instantiation in the sample, we can use the following:

```suggestion
      // create a point located at San Francisco, CA to be used as the viewpoint for the map
      var point= new Point(-13631205.660131, 4546829.846004, SpatialReferences.getWebMercator());
```

symbology/apply-unique-values-with-alternate-symbols/src/main/java/com/esri/samples/apply_unique_values_with_alternate_symbols/ApplyUniqueValuesWithAlternateSymbolsSample.java
```diff
@@ -0,0 +1,231 @@
...
+      ArcGISMap map = new ArcGISMap(BasemapStyle.ARCGIS_TOPOGRAPHIC);
+
+      // create a map view and set the map to it
+      mapView = new MapView();
+      mapView.setMap(map);
+
+      // set a viewpoint to the map view centered on a point
+      mapView.setViewpointCenterAsync(vPoint, 25000.0);
+
+      // creating box, button, and label UI components
```
Rachael-E: this applies to the following comments too, let's keep the tense the same within the code in comments and go with e.g. create, reset, update instead. 

```suggestion
      // create a controls vbox, button, and label UI components
```

symbology/apply-unique-values-with-alternate-symbols/src/main/java/com/esri/samples/apply_unique_values_with_alternate_symbols/ApplyUniqueValuesWithAlternateSymbolsSample.java
```diff
@@ -0,0 +1,231 @@
...
+      VBox vBox = controlsVBox();
+
+      // resetting map viewpoint on button press
```
Rachael-E: actually we don't need a comment up here since the code is self explanatory that on action, the map viewpoint is reset. and there is a useful comment in the event.

symbology/apply-unique-values-with-alternate-symbols/src/main/java/com/esri/samples/apply_unique_values_with_alternate_symbols/ApplyUniqueValuesWithAlternateSymbolsSample.java
```diff
@@ -0,0 +1,231 @@
...
+      btnReset.setOnAction(event -> {
+        // reset the map viewpoint
+        mapView.setViewpointAsync(new Viewpoint(vPoint, 25000.0), 1.5f, AnimationCurve.EASE_IN_OUT_SINE);
```
Rachael-E: since you are reusing the same viewpoint as you did on line 96, we can make this more readable by setting that viewpoint as a variable above the first line of code it will be used (e.g. after you instantiate the point) and use it in both places the viewpoint is set.

```java
var viewpoint = new Viewpoint(point, 25000.0), 1.5f, AnimationCurve.EASE_IN_OUT_SINE
```

symbology/apply-unique-values-with-alternate-symbols/src/main/java/com/esri/samples/apply_unique_values_with_alternate_symbols/ApplyUniqueValuesWithAlternateSymbolsSample.java
```diff
@@ -0,0 +1,231 @@
...
+      });
+
+      // updating label scale values
```
Rachael-E: same for this comment: since you already have a helpful comment in the listener block, we don't need that information repeated above

symbology/apply-unique-values-with-alternate-symbols/src/main/java/com/esri/samples/apply_unique_values_with_alternate_symbols/ApplyUniqueValuesWithAlternateSymbolsSample.java
```diff
@@ -0,0 +1,231 @@
...
+      mapView.addViewpointChangedListener(event->
+        // update label text
+        lblScale.setText("Current scale: 1:" + String.format("%.2f", mapView.getMapScale())));
+
+      // creating simple markers where the blue square and yellow diamond are alternate symbols
```
Rachael-E: just another tense update here, and some others to address following this one
```suggestion
      // create simple marker symboles where the blue square and yellow diamond are alternate symbols
```

symbology/apply-unique-values-with-alternate-symbols/src/main/java/com/esri/samples/apply_unique_values_with_alternate_symbols/ApplyUniqueValuesWithAlternateSymbolsSample.java
```diff
@@ -0,0 +1,231 @@
...
+      // the red triangle is used to create a unique symbol and the purple diamond is the default symbol for the unique value renderer
+      SimpleMarkerSymbol sBlueSquareSymbol = new SimpleMarkerSymbol(SimpleMarkerSymbol.Style.SQUARE, 0xFF0000FF, 15);
+      SimpleMarkerSymbol sYellowDiamondSymbol = new SimpleMarkerSymbol(SimpleMarkerSymbol.Style.DIAMOND, 0xFFFFFF00, 15);
+      SimpleMarkerSymbol sRedTriangleSymbol = new SimpleMarkerSymbol(SimpleMarkerSymbol.Style.TRIANGLE, 0xFFFF0000, 15);
+      SimpleMarkerSymbol sPurpleDiamondSymbol = new SimpleMarkerSymbol(SimpleMarkerSymbol.Style.DIAMOND, 0xFF800080, 12);
```
Rachael-E: ah cool, I see what you're doing here with `s` and `m` pre-pended to the variable name. However since we cater to all range of developers from brand new to experienced, we keep our code as readable and spelled out as possible (within reason). So for this I would go for: 

```suggestion
      SimpleMarkerSymbol simpleBlueSquareSymbol = new SimpleMarkerSymbol(SimpleMarkerSymbol.Style.SQUARE, 0xFF0000FF, 15);
      SimpleMarkerSymbol simpleYellowDiamondSymbol = new SimpleMarkerSymbol(SimpleMarkerSymbol.Style.DIAMOND, 0xFFFFFF00, 15);
      SimpleMarkerSymbol simpleRedTriangleSymbol = new SimpleMarkerSymbol(SimpleMarkerSymbol.Style.TRIANGLE, 0xFFFF0000, 15);
      SimpleMarkerSymbol simplePurpleDiamondSymbol = new SimpleMarkerSymbol(SimpleMarkerSymbol.Style.DIAMOND, 0xFF800080, 12);
```

symbology/apply-unique-values-with-alternate-symbols/src/main/java/com/esri/samples/apply_unique_values_with_alternate_symbols/ApplyUniqueValuesWithAlternateSymbolsSample.java
```diff
@@ -0,0 +1,231 @@
...
+
+      // converting simpler marker symbols to multilayer point symbols
+      MultilayerPointSymbol mBlueSquareSymbol = sBlueSquareSymbol.toMultilayerSymbol();
+      MultilayerPointSymbol mYellowDiamondSymbol = sYellowDiamondSymbol.toMultilayerSymbol();
+      MultilayerPointSymbol mRedTriangleSymbol = sRedTriangleSymbol.toMultilayerSymbol();
+      MultilayerPointSymbol mPurpleDiamondSymbol = sPurpleDiamondSymbol.toMultilayerSymbol();
```
Rachael-E: same here: let's spell it out

```suggestion
      // convert simple marker symbols to multilayer point symbols
      MultilayerPointSymbol multilayerBlueSquareSymbol = sBlueSquareSymbol.toMultilayerSymbol();
      MultilayerPointSymbol multilayerYellowDiamondSymbol = sYellowDiamondSymbol.toMultilayerSymbol();
      MultilayerPointSymbol multilayerRedTriangleSymbol = sRedTriangleSymbol.toMultilayerSymbol();
      MultilayerPointSymbol multilayerPurpleDiamondSymbol = sPurpleDiamondSymbol.toMultilayerSymbol();
```

symbology/apply-unique-values-with-alternate-symbols/src/main/java/com/esri/samples/apply_unique_values_with_alternate_symbols/ApplyUniqueValuesWithAlternateSymbolsSample.java
```diff
@@ -0,0 +1,231 @@
...
+
+      // setting scale range values through reference properties for the multilayer symbols
+      SymbolReferenceProperties blueSquareRef = new SymbolReferenceProperties(10000.0, 5000.0);
+      SymbolReferenceProperties yellowDiamondRef = new SymbolReferenceProperties(20000.0, 10000.0);
+      SymbolReferenceProperties redTriangleRef = new SymbolReferenceProperties(5000.0, 0.0);
+
+      mBlueSquareSymbol.setReferenceProperties(blueSquareRef);
+      mYellowDiamondSymbol.setReferenceProperties(yellowDiamondRef);
+      mRedTriangleSymbol.setReferenceProperties(redTriangleRef);
```
Rachael-E: since we just use these symbolreference properties in one place, to save on lines of code we can instantiate them in place here e.g.

```suggestion
      mBlueSquareSymbol.setReferenceProperties(new SymbolReferenceProperties(10000.0, 5000.0));
      mYellowDiamondSymbol.setReferenceProperties(new SymbolReferenceProperties(20000.0, 10000.0));
      mRedTriangleSymbol.setReferenceProperties(new SymbolReferenceProperties(5000.0, 0.0));
```

symbology/apply-unique-values-with-alternate-symbols/src/main/java/com/esri/samples/apply_unique_values_with_alternate_symbols/ApplyUniqueValuesWithAlternateSymbolsSample.java
```diff
@@ -0,0 +1,231 @@
...
+
+      // Creating unique value for the red triangle and list of alternate symbols
+      UniqueValueRenderer.UniqueValue uniqueValue =
```
Rachael-E: let's use the import statement for this above so that we can just keep the readability. This then makes this a good candidate for the use of `var` to save the repeat of UniqueValue x 3! 

```
import com.esri.arcgisruntime.symbology.UniqueValueRenderer.UniqueValue;
```

```suggestion
      var uniqueValue =
```

symbology/apply-unique-values-with-alternate-symbols/src/main/java/com/esri/samples/apply_unique_values_with_alternate_symbols/ApplyUniqueValuesWithAlternateSymbolsSample.java
```diff
@@ -0,0 +1,231 @@
...
+        new UniqueValueRenderer.UniqueValue("unique values based on request type", "unique value",
+          mRedTriangleSymbol, List.of("Damaged Property"), List.of(mBlueSquareSymbol, mYellowDiamondSymbol));
```
Rachael-E: good question here on `List.of`: I like this approach! There is an alternative approach to use but I'm just putting that here more of an example of how else you could do it, I don't think we have to change this. This is quite a good stack overflow post showing the differences between `List.of` and `Arrays.asList`

https://stackoverflow.com/questions/46579074/what-is-the-difference-between-list-of-and-arrays-aslist

and some thoughts with `List.of` and `Collections.singletonList`
https://dzone.com/articles/singleton-list-showdown-collectionssingletonlist-v

```java
// alternative
        new UniqueValue("unique values based on request type", "unique value",
          mRedTriangleSymbol, Collections.singletonList("Damaged Property"), Arrays.asList(mBlueSquareSymbol, mYellowDiamondSymbol));
```

symbology/apply-unique-values-with-alternate-symbols/src/main/java/com/esri/samples/apply_unique_values_with_alternate_symbols/ApplyUniqueValuesWithAlternateSymbolsSample.java
```diff
@@ -0,0 +1,231 @@
...
+
+      // Creating unique value renderer
+      UniqueValueRenderer uniqueValRenderer = new UniqueValueRenderer();
```
Rachael-E: we can use var here with updated variable name

```suggestion
      var uniqueValueRenderer = new UniqueValueRenderer();
```

We could also make use of the second constructor here for UniqueValueRenderer (https://developers.arcgis.com/java/api-reference/reference/com/esri/arcgisruntime/symbology/UniqueValueRenderer.html): the empty one above creates default empty lists/strings which we then add to later in the current code, but we can pass those properties directly in for the second constructur (using an empty string as the label). This would cut down on a couple lines of code without affecting readability

```java
      UniqueValueRenderer uniqueValRenderer = new UniqueValueRenderer(List.of("req_type"), List.of(uniqueValue), "", mPurpleDiamondSymbol);
```

symbology/apply-unique-values-with-alternate-symbols/src/main/java/com/esri/samples/apply_unique_values_with_alternate_symbols/ApplyUniqueValuesWithAlternateSymbolsSample.java
```diff
@@ -0,0 +1,231 @@
...
+      uniqueValRenderer.getUniqueValues().add(uniqueValue);
+      uniqueValRenderer.getFieldNames().add(("req_type"));
+
+      // Applying default symbol for the unique value renderer.
+      uniqueValRenderer.setDefaultSymbol(mPurpleDiamondSymbol);
+
+      // creating a service feature table using the url
+      final ServiceFeatureTable featureTable = new ServiceFeatureTable(FEATURE_SERVICE_URL);
```
Rachael-E: var opportunity here too

```suggestion
      final var serviceFeatureTable = new ServiceFeatureTable(FEATURE_SERVICE_URL);
```

symbology/apply-unique-values-with-alternate-symbols/src/main/java/com/esri/samples/apply_unique_values_with_alternate_symbols/ApplyUniqueValuesWithAlternateSymbolsSample.java
```diff
@@ -0,0 +1,231 @@
...
+
+      featureLayer.addDoneLoadingListener(()->{
+        if (featureLayer.getLoadStatus() == LoadStatus.LOADED) {
+          // adding feature layer to ArcGISMap
+          map.getOperationalLayers().add(featureLayer);
+        } else {
+          new Alert(Alert.AlertType.ERROR, "Error loading Feature Table from service" + featureLayer.getLoadError().getMessage()).show();
+        }
+      });
+
```
Rachael-E: This is another classic example of where we have multiple ways to achieve the same thing, and a really good thing to ask about!

This approach will work just fine since you are loading the feature layer, waiting for it to load, before adding it to the map. However you have two loads in here, both of which aren't strictly necessary: the loadAsync call being the obvious one. The second one is less obvious: calling `map.getOperationalLayers.add(featureLayer)` automatically loads the layer being added to the map. So we could just simplify this to the following:

```java
          map.getOperationalLayers().add(featureLayer); // loads the feature layer when added to the operational layers
```

However this could result in the feature layer failing silently, say for example the URL service was down. The map would load but not the feature layer and we wouldn't see why. So what we can do is ask the service feature table to load instead, and the load of the service feature table will bubble up to the feature layer. That way if the service is down we'll get a meaningful message from the service feature table, plus if a user wanted to put in their own service which may have more than one feature layer, they'd be able to lift this code out as is and it would load all the feature layers within the table. 

I've also added in a call to enable the vBox at this stage: if you set it's disable property as true when you create the box, it will mean the vbox will only become active when the feature layer has loaded. if the feature layer hasn't loaded, no need to have an active vbox :D 

It's also generally best practice to put the loading call _after_ the addDoneLoadingListener. That way you're telling the code what to expect to do when it has finished loading, before giving it the instruction to load. 

```java
      // check the service feature table has loaded before adding the feature layer to the map
      serviceFeatureTable.addDoneLoadingListener( () -> {
        if (serviceFeatureTable.getLoadStatus() == LoadStatus.LOADED) {
          map.getOperationalLayers().add(featureLayer);
          vBox.setDisable(false);
        } else {
          new Alert(Alert.AlertType.ERROR,
            "Error loading Feature Table from service" + serviceFeatureTable.getLoadError().getMessage()).show();
        }
      });
      // load the service feature table
      serviceFeatureTable.loadAsync();
```

symbology/apply-unique-values-with-alternate-symbols/src/main/java/com/esri/samples/apply_unique_values_with_alternate_symbols/ApplyUniqueValuesWithAlternateSymbolsSample.java
```diff
@@ -0,0 +1,231 @@
...
+   */
+  private VBox controlsVBox() {
+
+    // creating label to show scale values and button to reset viewpoint
+    lblScale = new Label("Current scale: 1:25000.00");
+    lblScale.setTextFill(Color.WHITE);
+    btnReset = new Button("Reset Viewpoint");
+
+    // creating and setting a box to hold the scale label and reset button
+    VBox vBox = new VBox(5);
```
Rachael-E: can add that vBox disable property in here at start

```
vBox.setDisable(true)
```

symbology/apply-unique-values-with-alternate-symbols/src/main/java/com/esri/samples/apply_unique_values_with_alternate_symbols/ApplyUniqueValuesWithAlternateSymbolsSample.java
```diff
@@ -0,0 +1,231 @@
...
+      // on any error, display the stack trace.
+      e.printStackTrace();
+    }
+  }
+
+  /**
+   * Creates a UI with three buttons and a label.
+   * @return a vBox populated with buttons and labels
+   */
+  private VBox controlsVBox() {
```
Rachael-E: cool: I like this keeping the UI separate! We can make this a little more readable with the following

- rename the method `setUpControlsVBox` and have it set as a void method (don't return as VBox). Suggesting this as when the method `controlsVBox` is called up front, it's not clear it contains all of the additional good UI-ness. So for a new user, naming the method in a way that is clear this is a UI set up helps them understand where it's coming from better. It also saves you creating two new VBoxes: the one in this method and the one in the `start` method.
- set `vBox` as a member variable instead and add that to the stackpane in the `start` method.

symbology/apply-unique-values-with-alternate-symbols/src/main/java/com/esri/samples/apply_unique_values_with_alternate_symbols/ApplyUniqueValuesWithAlternateSymbolsSample.java
```diff
@@ -0,0 +1,231 @@
...
+
+    // creating label to show scale values and button to reset viewpoint
+    lblScale = new Label("Current scale: 1:25000.00");
```
Rachael-E: since we just have one label in this sample let's just call this 

```suggestion
    label = new Label("Current scale: 1:25000.00");
```

symbology/apply-unique-values-with-alternate-symbols/src/main/java/com/esri/samples/apply_unique_values_with_alternate_symbols/ApplyUniqueValuesWithAlternateSymbolsSample.java
```diff
@@ -0,0 +1,231 @@
...
+    lblScale.setTextFill(Color.WHITE);
+    btnReset = new Button("Reset Viewpoint");
```
Rachael-E: since this is also the only button in the sample, let's go with 

```suggestion
    button = new Button("Reset Viewpoint");
```

symbology/apply-unique-values-with-alternate-symbols/src/main/java/com/esri/samples/apply_unique_values_with_alternate_symbols/ApplyUniqueValuesWithAlternateSymbolsSample.java
Rachael-E: when the member variable is added this can become

```suggestion
    vBox = new VBox(5);
```

symbology/apply-unique-values-with-alternate-symbols/src/main/java/com/esri/samples/apply_unique_values_with_alternate_symbols/ApplyUniqueValuesWithAlternateSymbolsSample.java
```diff
@@ -0,0 +1,231 @@
...
+
+    // creating and setting a box to hold the scale label and reset button
+    VBox vBox = new VBox(5);
+    vBox.setMinWidth(150);
+    vBox.getChildren().addAll(btnReset);
+    btnReset.prefWidthProperty().bind(vBox.widthProperty());
+
+    VBox controlsVBox = new VBox(10);
```
Rachael-E: we can simplify this set up: as you currently have it set up, you have a VBox which contains a HBox with a label, and then a VBox with a button. Each Box contains its own element: this would be important in a complex UI layout where you need to tightly control placement, but in this sample where we have two elements, one placed about the over, a single VBox (so that the elements are stacked vertically) is fine and saves on a bunch of unnecessary UI set up code :D 

```java
    // creating label to show scale values and button to reset viewpoint
    label = new Label("Current scale: 1:25000");
    label .setTextFill(Color.WHITE);
    button = new Button("Reset Viewpoint");

    controlsVBox = new VBox(10); // this can be set as the member variable vBox
    controlsVBox.setBackground(new Background(new BackgroundFill(Paint.valueOf("rgba(0,0,0,0.5)"), // this can be a bit paler
    controlsVBox.setMaxSize(180, 75); // have updated the maxWidth property here to fit the label

    controlsVBox.getChildren().addAll(label, button);

```

symbology/apply-unique-values-with-alternate-symbols/src/main/java/com/esri/samples/apply_unique_values_with_alternate_symbols/ApplyUniqueValuesWithAlternateSymbolsSample.java
```diff
@@ -0,0 +1,231 @@
...
+   * Opens and runs application.
+   *
+   * @param args arguments passed to this application
+   */
+  public static void main(String[] args) {
+
+    Application.launch(args);
+  }
+
+}
```
Rachael-E: add a new line at the end of the class

Rachael-E (changes requested):
great job on your first sample @Luiy0 , very smooth UI and well structured code! Don't be alarmed by the number of comments, these are all normal/regular things that come up when starting on the team, and the more you are on samples, the more second nature they will be become!

General summary:
- redirect_from key in the metadata json file has been flagged up as a failing check: you can check the Github checks in the PR to see if there are any issues with the metadata.json or readme files, they should provide helpful (sometimes unhelpful!) messages if something is up
- let's keep constant tense across samples comments, e.g. "create" instead of "creating"
- good thoughts with List.of, this is fine and a good approach to take
- loadables: good examples in this code of a couple of different approaches I have commented on
- some opportunities in here for using `var` in place of explicit class declaration
- let's keep variable names as plain english readable as possible: keep in mind the view of a brand new user who may not have much development experience
- good use of separate method to handle UI set up: couple of tidy ups suggested in there.

symbology/apply-unique-values-with-alternate-symbols/src/main/java/com/esri/samples/apply_unique_values_with_alternate_symbols/ApplyUniqueValuesWithAlternateSymbolsLauncher.java
```diff
@@ -0,0 +1,9 @@
+package com.esri.samples.apply_unique_values_with_alternate_symbols;
+
+/**
+ * Wrapper required for launching a JavaFX 11 app through Gradle or from a jar.
+ */
+public class ApplyUniqueValuesWithAlternateSymbolsLauncher {
+
+  public static void main(String[] args) { ApplyUniqueValuesWithAlternateSymbolsSample.main(args); }
```
Rachael-E: this was a nit pick for Oliver too: it doesn't affect how the code runs but just for readability let's separate this out a bit

```suggestion
  public static void main(String[] args) { 
    ApplyUniqueValuesWithAlternateSymbolsSample.main(args); 
  }
```

Luiy0:
@Rachael-E Thanks for the in-depth review. I have implemented the necessary changes and pushed the branch again for you to revise. I will be on the lookout for any other suggestions that may come up.

symbology/apply-unique-values-with-alternate-symbols/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+Zoom in and out of the map to see alternate symbols at each scale. The symbology changes according to the following scale ranges: 1:0-1:5000, 1:5000-1:10000, 1:10000-1:20000. To go back to the initial viewpoint, click "Reset Viewpoint".
+
+## How it works
+
+1. Create a `FeatureLayer` using the service url and add it to the map's list of operational layers.
+2. Create two alternate symbols:
+   * Create a symbol using `SimpleMarkerSymbol`.
+   * Convert the simple marker symbol to a `MultilayerPointSymbol` using `SimpleMarkerSymbol.toMultilayerSymbol()`.
+   * Set the valid scale range through reference properties on the multilayer point symbols by calling `multilayerPointSymbol.referenceProperties = SymbolReferenceProperties(double minScale, double maxScale)`.
```
Rachael-E:
```suggestion
   * Set the valid scale range through reference properties on the multilayer point symbols by calling `multilayerPointSymbol.setReferenceProperties(new SymbolReferenceProperties(double minScale, double maxScale))`.
```

symbology/apply-unique-values-with-alternate-symbols/src/main/java/com/esri/samples/apply_unique_values_with_alternate_symbols/ApplyUniqueValuesWithAlternateSymbolsSample.java
```diff
@@ -0,0 +1,212 @@
...
+import com.esri.arcgisruntime.layers.FeatureLayer;
+import com.esri.arcgisruntime.symbology.MultilayerPointSymbol;
+import com.esri.arcgisruntime.symbology.SimpleMarkerSymbol;
+import com.esri.arcgisruntime.symbology.SymbolReferenceProperties;
+import com.esri.arcgisruntime.symbology.UniqueValueRenderer;
+
+public class ApplyUniqueValuesWithAlternateSymbolsSample extends Application{
+
+  private MapView mapView;
+  private FeatureLayer featureLayer;
```
Rachael-E: since it's not always obvious in samples why we use a member variable for a loadable, we've taken the approach in samples to add a comment here on loadables to let the user know why this is the case:

```suggestion
  private FeatureLayer featureLayer; // keep loadable in scope to avoid garbage collection
```

symbology/apply-unique-values-with-alternate-symbols/src/main/java/com/esri/samples/apply_unique_values_with_alternate_symbols/ApplyUniqueValuesWithAlternateSymbolsSample.java
```diff
@@ -0,0 +1,212 @@
...
+      stage.setWidth(800);
+      stage.setHeight(700);
+      stage.setScene(scene);
+      stage.show();
+
+      // authentication with an API key or named user is required to access base maps and other location services
+      String yourAPIKey = System.getProperty("apiKey");
+      ArcGISRuntimeEnvironment.setApiKey(yourAPIKey);
+
+      // create a map with the standard imagery basemap style
```
Rachael-E:
```suggestion
      // create a map with the topographic basemap style
```

symbology/apply-unique-values-with-alternate-symbols/src/main/java/com/esri/samples/apply_unique_values_with_alternate_symbols/ApplyUniqueValuesWithAlternateSymbolsSample.java
```diff
@@ -0,0 +1,212 @@
...
+      mapView.setMap(map);
+
+      // create a point located at San Francisco, CA to be used as the viewpoint for the map
+      var point = new Point(-13631205.660131, 4546829.846004, SpatialReferences.getWebMercator());
+
+      // set a viewpoint to the map view centered on a point
+      mapView.setViewpointCenterAsync(point, 25000.0);
+
+      // create a controls vbox, button, and label UI components
+      setUpControlsVBox();
```
Rachael-E: let's put this up straight after the apiKey code is called, just helps keep the flow of Runtime specific code together

symbology/apply-unique-values-with-alternate-symbols/src/main/java/com/esri/samples/apply_unique_values_with_alternate_symbols/ApplyUniqueValuesWithAlternateSymbolsSample.java
```diff
@@ -0,0 +1,212 @@
...
+
+      var viewpoint = new Viewpoint(point, 25000);
```
Rachael-E: this can be called after the point is instantiated, and then the mapview viewpoint can be set with

```mapView.setViewpointCenterAsync(viewpoint)```

File: symbology/apply-unique-values-with-alternate-symbols/src/main/java/com/esri/samples/apply_unique_values_with_alternate_symbols/ApplyUniqueValuesWithAlternateSymbolsSample.java [hunk 39]
```diff
@@ -0,0 +1,212 @@
...
+      button.setOnAction(event -> {
+        // reset the map viewpoint
+        mapView.setViewpointAsync(viewpoint, 1.5f, AnimationCurve.EASE_IN_OUT_SINE);
+      });
+
+      DecimalFormat decFormat = new DecimalFormat("#");
+      mapView.addViewpointChangedListener(event->
+        // update label text and formatting scale values
+        label.setText("Current scale: 1:" + decFormat.format(mapView.getMapScale())));
```
Rachael-E: to save a line of code, decimalFormat can be removed and replaced with

```suggestion
               label.setText("Current scale: 1:" + Math.round(mapView.getMapScale())));
```

Rachael-E (commented):
thanks @Luiy0 , just a couple more small changes needed here then it's good for 2nd review!

Rachael-E (approved):
looks great, thanks @Luiy0 ! This is ready for second review now. @jenmerritt could you take a second look please?

symbology/apply-unique-values-with-alternate-symbols/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+* MultilayerPointSymbol
+* SimpleMarkerSymbol
+* SymbolReferenceProperties
+* UniqueValue
+* UniqueValueRenderer
+
+## About the data
+
+The [San Francisco 311 incidents layer](https://sampleserver6.arcgisonline.com/arcgis/rest/services/SF311/FeatureServer/0) in this sample displays point features related to crime incidents such as graffiti and tree damage that have been reported by city residents.
```
jenmerritt:
```suggestion
The [San Francisco 311 incidents layer](https://sampleserver6.arcgisonline.com/arcgis/rest/services/SF311/FeatureServer/0) used in this sample displays point features related to crime incidents such as graffiti and tree damage that have been reported by city residents.
```

symbology/apply-unique-values-with-alternate-symbols/src/main/java/com/esri/samples/apply_unique_values_with_alternate_symbols/ApplyUniqueValuesWithAlternateSymbolsSample.java
```diff
@@ -0,0 +1,210 @@
...
+import com.esri.arcgisruntime.data.ServiceFeatureTable;
+import com.esri.arcgisruntime.geometry.Point;
+import com.esri.arcgisruntime.geometry.SpatialReferences;
+import com.esri.arcgisruntime.layers.FeatureLayer;
+import com.esri.arcgisruntime.symbology.MultilayerPointSymbol;
+import com.esri.arcgisruntime.symbology.SimpleMarkerSymbol;
+import com.esri.arcgisruntime.symbology.SymbolReferenceProperties;
+import com.esri.arcgisruntime.symbology.UniqueValueRenderer;
+
+public class ApplyUniqueValuesWithAlternateSymbolsSample extends Application{
```
jenmerritt:
```suggestion
public class ApplyUniqueValuesWithAlternateSymbolsSample extends Application {
```

symbology/apply-unique-values-with-alternate-symbols/src/main/java/com/esri/samples/apply_unique_values_with_alternate_symbols/ApplyUniqueValuesWithAlternateSymbolsSample.java
```diff
@@ -0,0 +1,210 @@
...
+  @Override
+  public void start(Stage stage) {
+
+    try {
+      // create stack pane and application scene
+      StackPane stackPane = new StackPane();
+      Scene scene = new Scene(stackPane);
+
+      // set title, size, and add scene to stage
+      stage.setTitle("Apply unique values with alternate symbols");
```
jenmerritt:
```suggestion
      stage.setTitle("Apply Unique Values With Alternate Symbols");
```
Camel Case

symbology/apply-unique-values-with-alternate-symbols/src/main/java/com/esri/samples/apply_unique_values_with_alternate_symbols/ApplyUniqueValuesWithAlternateSymbolsSample.java
```diff
@@ -0,0 +1,210 @@
...
+
+      // create a map view and set the map to it
+      mapView = new MapView();
+      mapView.setMap(map);
+
+      // create a point located at San Francisco, CA to be used as the viewpoint for the map
+      var point = new Point(-13631205.660131, 4546829.846004, SpatialReferences.getWebMercator());
+      var viewpoint = new Viewpoint(point, 25000);
+
+      // set a viewpoint to the map view centered on a point
```
jenmerritt:
```suggestion
      // set the viewpoint to the map view
```

symbology/apply-unique-values-with-alternate-symbols/src/main/java/com/esri/samples/apply_unique_values_with_alternate_symbols/ApplyUniqueValuesWithAlternateSymbolsSample.java
```diff
@@ -0,0 +1,210 @@
...
+      mapView.setViewpointCenterAsync(point, 25000.0);
+
+      button.setOnAction(event -> {
+        // reset the map viewpoint
+        mapView.setViewpointAsync(viewpoint, 1.5f, AnimationCurve.EASE_IN_OUT_SINE);
+      });
+
+      mapView.addViewpointChangedListener(event->
+        // update label text and formatting scale values
+        label.setText("Current scale: 1:" + Math.round(mapView.getMapScale())));
```
jenmerritt:
```suggestion
      mapView.addViewpointChangedListener(event -> {
        // update label text and formatting scale values
        label.setText("Current scale: 1:" + Math.round(mapView.getMapScale()));
      });
```

symbology/apply-unique-values-with-alternate-symbols/src/main/java/com/esri/samples/apply_unique_values_with_alternate_symbols/ApplyUniqueValuesWithAlternateSymbolsSample.java
```diff
@@ -0,0 +1,210 @@
...
+      multilayerRedTriangleSymbol.setReferenceProperties(new SymbolReferenceProperties(5000.0, 0.0));
+
+      // create unique value for the red triangle and list of alternate symbols
+      var uniqueValue = new UniqueValueRenderer.UniqueValue("unique values based on request type", "unique value",
+          multilayerRedTriangleSymbol, List.of("Damaged Property"), List.of(multilayerBlueSquareSymbol, multilayerYellowDiamondSymbol));
+
+      // create unique value renderer
+      var uniqueValRenderer = new UniqueValueRenderer(List.of("req_type"), List.of(uniqueValue), "", multilayerPurpleDiamondSymbol);
+
+      // apply default symbol for the unique value renderer.
```
jenmerritt:
```suggestion
      // apply default symbol for the unique value renderer
```

symbology/apply-unique-values-with-alternate-symbols/src/main/java/com/esri/samples/apply_unique_values_with_alternate_symbols/ApplyUniqueValuesWithAlternateSymbolsSample.java
```diff
@@ -0,0 +1,210 @@
...
+      uniqueValRenderer.setDefaultSymbol(multilayerPurpleDiamondSymbol);
+
+      // create a service feature table using the url
```
jenmerritt:
```suggestion
      // create a service feature table using the feature service url
```

symbology/apply-unique-values-with-alternate-symbols/src/main/java/com/esri/samples/apply_unique_values_with_alternate_symbols/ApplyUniqueValuesWithAlternateSymbolsSample.java
```diff
@@ -0,0 +1,210 @@
...
+      featureLayer.setRenderer(uniqueValRenderer);
+
+      // check the service feature table has loaded before adding the feature layer to the map
+      serviceFeatureTable.addDoneLoadingListener( () -> {
+        if (serviceFeatureTable.getLoadStatus() == LoadStatus.LOADED) {
+          map.getOperationalLayers().add(featureLayer);
+          vBox.setDisable(false);
+        } else {
+          new Alert(Alert.AlertType.ERROR,
+            "Error loading Feature Table from service" + serviceFeatureTable.getLoadError().getMessage()).show();
```
jenmerritt:
```suggestion
            "Error loading Feature Table from Service: " + serviceFeatureTable.getLoadError().getMessage()).show();
```
Add a space so that the message string isn't just appended directly to your first string.

symbology/apply-unique-values-with-alternate-symbols/src/main/java/com/esri/samples/apply_unique_values_with_alternate_symbols/ApplyUniqueValuesWithAlternateSymbolsSample.java
```diff
@@ -0,0 +1,210 @@
...
+      e.printStackTrace();
+    }
+  }
+
+  /**
+   * Creates a UI with three buttons and a label.
+   */
+  private void setUpControlsVBox() {
+
+    // creating label to show scale values and button to reset viewpoint
```
jenmerritt:
```suggestion
    // create label to show scale values and button to reset viewpoint
```

symbology/apply-unique-values-with-alternate-symbols/src/main/java/com/esri/samples/apply_unique_values_with_alternate_symbols/ApplyUniqueValuesWithAlternateSymbolsSample.java
```diff
@@ -0,0 +1,210 @@
...
+    label = new Label("Current scale: 1:25000");
+    label .setTextFill(Color.WHITE);
+    button = new Button("Reset Viewpoint");
+
+    vBox = new VBox(10);
```
jenmerritt, line 180:
```suggestion
    // create and configure a VBox
    vBox = new VBox(10);
```

symbology/apply-unique-values-with-alternate-symbols/src/main/java/com/esri/samples/apply_unique_values_with_alternate_symbols/ApplyUniqueValuesWithAlternateSymbolsSample.java
```diff
@@ -0,0 +1,210 @@
...
+    vBox.setBackground(new Background(new BackgroundFill(Paint.valueOf("rgba(0,0,0,0.6)"),
+      CornerRadii.EMPTY, Insets.EMPTY)));
+    vBox.setPadding(new Insets(10.0));
+    vBox.setMaxSize(180, 75);
+    vBox.setDisable(true);
+
+    vBox.getChildren().addAll(label, button);
+
```
jenmerritt, line 189:
```suggestion
    // add the label and button to the VBox
    vBox.getChildren().addAll(label, button);
```

jenmerritt (commented):
Nice work @Luiy0 ! Congrats on the first sample. Some very minor suggestions to consider below.

Luiy0:
Hello @jenmerritt, I've implemented the suggestions listed above. Thanks for the feedback.

Luiy0:
Hello @kathleen1, this is ready for verification in the next sample viewer build.

PR #696, opened 2022-07-21
[Update] Set Max Extent

01smito01:
### Description

A couple minor updates to this sample to fix issues found in verification:
- Increased the width of the control panel to prevent text clipping on Linux.
- Updated the README instructions so they are consistent with the name of the checkbox in the sample.

I'd quite like the control panel to crop nicely if the extra width for the text wasn't needed. I tried fiddling around with `setMinSize` and `setPrefSize` to achieve this but wasn't successful. Would appreciate being shown how to do this, assuming it's possible.

@Rachael-E I'm not really sure what the proper workflow is for resolving issues found in verification so feel free to delete if this isn't it.

Checklist:
- [x] Request a reviewer
- [ ] Assign a reviewer
- [x] Tag reviewer in comment

### DIFF
--- a/map/set-max-extent/README.md
+++ b/map/set-max-extent/README.md
@@ -13,7 +13,7 @@ ability to pan or zoom away.
 
 The application loads with a map whose maximum extent has been set to the borders of Colorado. Note
 that you won't be able to pan far from the Colorado border or zoom out beyond the minimum scale set
-by the max extent. Use the "Max Extent OFF" button to disable the max extent to freely pan/zoom around the map.
+by the max extent. Click the "Enable Max Extent" checkbox to disable the max extent to freely pan/zoom around the map.
 
 ## How it works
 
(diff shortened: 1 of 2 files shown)


Rachael-E (approved):
thanks for making these changes @01smito01 . JavaFX is a fiddly one: I've not yet found a satisfactory solution to this and it's not unusual that Linux or Mac will clip the text in a VBox which looks otherwise perfectly fine on Windows. 

JavaFX has bindings, which can bind the properties of other elements to that of a UI component. For example, we could try the following but as you can see you'd still have to hard code the numerical values somewhere. @jenmerritt might have more thoughts here with her toolkit UI experiments!

```java
      // create a checkbox for toggling max extent
      var checkbox = new CheckBox("Enable Max Extent");
      checkbox.setPrefSize(160, 40);
      checkbox.setTextFill(Color.WHITE);
      checkbox.setSelected(true);
      checkbox.setOnMouseClicked(e -> {
        if (checkbox.isSelected()) {
          map.setMaxExtent(envelope);
        } else {
          map.setMaxExtent(null);
        }
      });

      // create a control panel
      var controlsVBox = new VBox(6);
      controlsVBox.setBackground(new Background(new BackgroundFill(Paint.valueOf("rgba(0,0,0,0.5)"),
      controlsVBox.maxWidthProperty().bind(checkbox.prefWidthProperty());
      controlsVBox.maxHeightProperty().bind(checkbox.prefHeightProperty());
      controlsVBox.setDisable(true);
```

PR #698, opened 2022-07-26
[New Sample] Query Features With Arcade Expression

01smito01:
### Description

Implementation of the "Query Features With Arcade Expression" sample. Main points I'd like feedback on:

- Not sure what category this should be in! I put it in `search` because search & query feel kind of similar. But I'm not sure it's the right place. Other platforms place it in `search and query`, `data`, and `display information`, so there isn't any consensus here.
- Layout/behaviour of the callout. Because of the Java callout implementation, I can't display the information all on one line like the Kotlin sample. I just want to check that the current layout is ok. Additionally, the callout will appear immediately if a layer is clicked, and then populate with data once the Arcade query has evaluated. I'm not sure if this is preferable to only showing the callout once the query has evaluated, which typically takes a second or two between clicking and the callout appearing.
- Adding on to what Colin said in his review of the Multilayer Symbols sample - at present this implementation calls a method that makes two separate async calls without making this clear at the top level, which isn't great. If there's an elegant way of refactoring this that would be good to know about!
- Style checker will fail because it doesn't recognise Arcade as a proper noun. Is what it is.

@Rachael-E over to you!

Checklist:
- [x] Set target branch to main (current release) or dev (next release)
- [x] Request a reviewer
- [x] Assign a reviewer
- [x] Tag reviewer in comment


search/query-features-with-arcade-expression/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+![QueryFeaturesWithArcadeExpression](QueryFeaturesWithArcadeExpression.png)
+
+## Use case
+
+Arcade is a portable, lightweight, and secure expression language used to create custom content in ArcGIS applications. Like other expression languages, it can perform mathematical calculations, manipulate text, and evaluate logical statements. It also supports multi-statement expressions, variables, and flow control statements. What makes Arcade particularly unique when compared to other expression and scripting languages is its inclusion of feature and geometry data types. This sample uses an Arcade expression to query the number of crimes in a neighborhood in the last 60 days.
+
+## How to use the sample
+
+Tap on any neighborhood to see the number of crimes in the last 60 days in a callout.
```
Rachael-E:
```suggestion
Click on any neighborhood to see the number of crimes in the last 60 days in a callout.
```

must resist autocorrect to neighbourhood... :D

search/query-features-with-arcade-expression/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+## How it works
+
+1. Create a `Portal` using the URL
```
Rachael-E:
```suggestion
1. Create a `Portal` using the URL.
```

search/query-features-with-arcade-expression/README.md
```diff
@@ -0,0 +1,62 @@
...
+2. Create a `PortalItem` using the `Portal` and ID.
+3. Create a `ArcGISMap` using the `PortalItem`.
```
Rachael-E:
```suggestion
3. Create an `ArcGISMap` using the `PortalItem`.
```

search/query-features-with-arcade-expression/README.md
```diff
@@ -0,0 +1,62 @@
...
+4. Set up a listener for taps on the map.
```
Rachael-E: can we elaborate here with a code snippet example? should also be `clicks` rather than `taps`, same for below

search/query-features-with-arcade-expression/README.md
```diff
@@ -0,0 +1,62 @@
...
+5. Identify the visible layer where it is tapped using `mapView.identifyLayerAsync()` and get the feature.
+6. Create the following `ArcadeExpression`:
+```java
+var arcadeExpression = new ArcadeExpression("var crimes = FeatureSetByName(\$map, 'Crime in the last 60 days');\n" + 
+  "return Count(Intersects(\$feature, crimes));");
+```
```
Rachael-E: ah I like this, but I don't know if our website will render this. best sticking with it in back ticks as we do our other code:

Maybe it could be something like:
```suggestion
`new ArcadeExpression("var crimes = FeatureSetByName(\$map, 'Crime in the last 60 days'); return Count(Intersects(\$feature, crimes));");`
```

search/query-features-with-arcade-expression/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+6. Create an `ArcadeEvaluator` using the Arcade expression and `ArcadeProfile.FORM_CALCULATION`.
+7. Create a map of profile variables with the following key-value pairs:
+```java
+var hashMap = new HashMap<String, Object>();
+hashMap.put("$feature", feature);
+hashMap.put("$map", mapView.getMap());
+```
```
Rachael-E: same comment as above. We don't need to go into too much detail in the readme since the user will look at the code. this could be summarised instead as something like below. Are $feature and $map arcade variables? 

```suggestion
Create a map of profile variables with $feature and $map as keys
```

search/query-features-with-arcade-expression/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+8. Call `ArcadeEvaluator.evaluateAsync()` on the Arcade evaluator object and pass it the profile variables map.
```
Rachael-E: since `ArcadeEvaluator` has already been "instantiated" above, it can be referred to here as below 

```suggestion
8. Call `arcadeEvaluator.evaluateAsync()` and pass it the profile variables map as parameters.
```

search/query-features-with-arcade-expression/README.md
```diff
@@ -0,0 +1,62 @@
...
+Assign the resulting `ListenableFuture<ArcadeEvaluationResult>` to its own variable.
+9. Add an `addDoneListener` to the `ListenableFuture`.
+10. Once the ListenableFuture is done, get the `ArcadeEvaluationResult` with `listenableFuture.get()`.
```
Rachael-E: These can be shortened into one line: the instructions don't need to be so precise here

```suggestion
9. Once the `evaluateAsync()` has finished evaluating, get the `ArcadeEvaluationResult` with `listenableFuture.get()`
```

search/query-features-with-arcade-expression/README.md
```diff
@@ -0,0 +1,62 @@
...
+11. Get the result from the `ArcadeEvaluationResult` with `arcadeEvaluationResult.getResult()`.
```
Rachael-E:
```suggestion
11. Get the result from the arcade evaluation result with `arcadeEvaluationResult.getResult()`.
```

search/query-features-with-arcade-expression/README.metadata.json
```diff
@@ -0,0 +1,35 @@
+{
+    "category": "Search",
```
Rachael-E: as with many of our samples, this is one that could live in a couple of places. I don't think Search is right as we use that category mainly for routing/geocoding. I think this sample is probably best placed in the `Map` category.

search/query-features-with-arcade-expression/src/main/java/com/esri/samples/query_features_with_arcade_expression/QueryFeaturesWithArcadeExpressionSample.java
```diff
@@ -0,0 +1,198 @@
...
+      var portalItem = new PortalItem(portal, "539d93de54c7422f88f69bfac2aebf7d");
+      map = new ArcGISMap(portalItem);
+
+      // create a map view and set the map to it
+      mapView = new MapView();
+      mapView.setMap(map);
+
+      // set callout properties that never change
+      mapView.getCallout().setLeaderPosition(Callout.LeaderPosition.BOTTOM);
+      mapView.getCallout().setTitle("RPD Beats");
```
Rachael-E: I don't see this title appear in the callout

search/query-features-with-arcade-expression/src/main/java/com/esri/samples/query_features_with_arcade_expression/QueryFeaturesWithArcadeExpressionSample.java
```diff
@@ -0,0 +1,198 @@
...
+
+      // wait for the map to finish loading data from portal before accessing its layers
+      map.addDoneLoadingListener(() -> {
+        if (map.getLoadStatus() == LoadStatus.LOADED) {
+          // hide all layers other than the RPD Beats Layer to reduce clutter
+          map.getOperationalLayers().forEach(layer -> layer.setVisible(layer.getName().equals(rpdBeatsLayerName)));
+
+          // get the RPD beats layer.
```
Rachael-E:
```suggestion
          // get the RPD beats layer
```

search/query-features-with-arcade-expression/src/main/java/com/esri/samples/query_features_with_arcade_expression/QueryFeaturesWithArcadeExpressionSample.java
```diff
@@ -0,0 +1,198 @@
...
+          policeBeatsLayer = map.getOperationalLayers().stream().filter(
+            layer -> layer.getName().equals(rpdBeatsLayerName)).toList().get(0);
```
Rachael-E: this is lovely use of a Java 16 feature here with `.toList()`. However, even though we work with, build, and test with Java 17, we don't include Java 17 code in our samples yet. This is because a large number of Java developers are still using Java 11, and wouldn't be able to run our samples if we sprinkle Java 17 language through them. Nice opportunity here though to see a new feature! tagging @Luiy0 for info.

Java 11 safe one here. Since streams can be a bit unreadable when chained together, we tend to split them up like so

```suggestion
          policeBeatsLayer = map.getOperationalLayers()
            .stream()
            .filter(layer -> layer.getName().equals(rpdBeatsLayerName))
            .collect(Collectors.toList()).get(0);
          
```

search/query-features-with-arcade-expression/src/main/java/com/esri/samples/query_features_with_arcade_expression/QueryFeaturesWithArcadeExpressionSample.java
```diff
@@ -0,0 +1,198 @@
...
+
+          // if map clicked, evaluate an arcade expression at that point
+          mapView.addEventHandler(MouseEvent.MOUSE_CLICKED, event -> {
```
Rachael-E: you should be able to shorten this a bit with

```  
        mapView.setOnMouseClicked(event -> {});
```

search/query-features-with-arcade-expression/src/main/java/com/esri/samples/query_features_with_arcade_expression/QueryFeaturesWithArcadeExpressionSample.java
```diff
@@ -0,0 +1,198 @@
...
+            if (event.getButton() == MouseButton.PRIMARY && event.isStillSincePress()) {
+              try {
+                evaluateArcadeExpression(new Point2D(event.getX(), event.getY()));
+              } catch (Exception e) {
+                throw new RuntimeException(e);
+              }
+            }});
```
Rachael-E: new line here to break up the curly braces cluster

search/query-features-with-arcade-expression/src/main/java/com/esri/samples/query_features_with_arcade_expression/QueryFeaturesWithArcadeExpressionSample.java
```diff
@@ -0,0 +1,198 @@
...
+      // add the map view to the stack pane
+      stackPane.getChildren().addAll(mapView);
+    } catch (Exception e) {
+      // on any error, display the stack trace.
+      e.printStackTrace();
+    }
+  }
+
+  /**
+   * Evaluates an arcade expression at the point on the map that was clicked.
```
Rachael-E:
```suggestion
   * Evaluates an arcade expression at the point where the map that was clicked.
```

search/query-features-with-arcade-expression/src/main/java/com/esri/samples/query_features_with_arcade_expression/QueryFeaturesWithArcadeExpressionSample.java
```diff
@@ -0,0 +1,198 @@
...
+   *
+   * @param point The point on the map which was clicked, in screen (rather than map) coordinates.
+   * @throws RuntimeException If unable to get results from futures due to interruption or concurrent execution.
+   */
+  private void evaluateArcadeExpression(Point2D point) throws RuntimeException {
```
Rachael-E: the contents of this whole method can just go in place where it's called above. We don't have any repeated code here and it's more of a story flow, so it's ok to have it in place.

search/query-features-with-arcade-expression/src/main/java/com/esri/samples/query_features_with_arcade_expression/QueryFeaturesWithArcadeExpressionSample.java
```diff
@@ -0,0 +1,198 @@
...
+        try {
+          var identifyLayerResult = identifyLayerResultFuture.get();
+
+          // only execute query if the police beats layer has data at the selected point
+          if (identifyLayerResult.getElements().size() != 0) {
+            ArcGISFeature feature = (ArcGISFeature) identifyLayerResult.getElements().get(0);
+
+            // show callout and text without data
+            mapView.getCallout().setTitle(calloutText);
+            mapView.getCallout().setDetail("");
```
Rachael-E: can we set the calloutText as the detail, and keep the Title as it was set at the beginning of the sample? Whilst there is a slight delay, I suggest something like the following to indicate some populating is happening. I originally tried with "Calculating" but the time lag is so minimal that it just looked weird seeing the callout go from a long to a shorter width as it adapted to the longer "Calculating" word. Maybe you could set a min size to the callout to counter that if you preferred that to the "..." approach.

```java
            mapView.getCallout().setDetail("..." + calloutText);

```

search/query-features-with-arcade-expression/src/main/java/com/esri/samples/query_features_with_arcade_expression/QueryFeaturesWithArcadeExpressionSample.java
```diff
@@ -0,0 +1,198 @@
...
+            mapView.getCallout().showCalloutAt(mapPoint);
+            mapView.getCallout().setVisible(true);
+
+            //setup arcade expression and evaluator
```
Rachael-E:
```suggestion
            // setup arcade expression and evaluator
```

search/query-features-with-arcade-expression/src/main/java/com/esri/samples/query_features_with_arcade_expression/QueryFeaturesWithArcadeExpressionSample.java
```diff
@@ -0,0 +1,198 @@
...
+            String expressionString = "var crimes = FeatureSetByName($map, 'Crime in the last 60 days');\n" +
```
Rachael-E: let's make this variable name super explicit that it's an arcade expression

```suggestion
            String arcadeExpressionString = "var crimes = FeatureSetByName($map, 'Crime in the last 60 days');\n" +
```

search/query-features-with-arcade-expression/src/main/java/com/esri/samples/query_features_with_arcade_expression/QueryFeaturesWithArcadeExpressionSample.java
```diff
@@ -0,0 +1,198 @@
...
+            var arcadeExpression = new ArcadeExpression(expressionString);
+            var arcadeEvaluator = new ArcadeEvaluator(arcadeExpression, ArcadeProfile.FORM_CALCULATION);
+
+            // instantiate key/value pairs
+            var hashMap = new HashMap<String, Object>();
+            hashMap.put("$feature", feature);
+            hashMap.put("$map", mapView.getMap());
+
+            // evaluate the arcade expression asynchronously
+            var resultFuture = arcadeEvaluator.evaluateAsync(hashMap);
```
Rachael-E: this isn't super clear what Object Type is being returned here, so let's use the type here instead of `var`

search/query-features-with-arcade-expression/src/main/java/com/esri/samples/query_features_with_arcade_expression/QueryFeaturesWithArcadeExpressionSample.java
```diff
@@ -0,0 +1,198 @@
...
+
+      // set callout properties that never change
+      mapView.getCallout().setLeaderPosition(Callout.LeaderPosition.BOTTOM);
+      mapView.getCallout().setTitle("RPD Beats");
+
+      // wait for the map to finish loading data from portal before accessing its layers
+      map.addDoneLoadingListener(() -> {
+        if (map.getLoadStatus() == LoadStatus.LOADED) {
+          // hide all layers other than the RPD Beats Layer to reduce clutter
+          map.getOperationalLayers().forEach(layer -> layer.setVisible(layer.getName().equals(rpdBeatsLayerName)));
```
Rachael-E: let's split this up over two lines for readability

```suggestion
          map.getOperationalLayers().forEach(layer -> 
          layer.setVisible(layer.getName().equals(rpdBeatsLayerName)));
```

search/query-features-with-arcade-expression/src/main/java/com/esri/samples/query_features_with_arcade_expression/QueryFeaturesWithArcadeExpressionSample.java
```diff
@@ -0,0 +1,198 @@
...
+import com.esri.arcgisruntime.portal.Portal;
+import com.esri.arcgisruntime.portal.PortalItem;
+
+public class QueryFeaturesWithArcadeExpressionSample extends Application {
+
+  private MapView mapView;
+  private ArcGISMap map;
+  private Layer policeBeatsLayer;
+  private final String rpdBeatsLayerName = "RPD Beats  - City_Beats_Border_1128-4500";
+  private final String calloutText = "Crimes in the last 60 days: ";
```
Rachael-E: I think we can have the text all in the one line here, like the following format:
```suggestion
  private final String calloutText = " crimes in the last 60 days";
```

image

search/query-features-with-arcade-expression/src/main/java/com/esri/samples/query_features_with_arcade_expression/QueryFeaturesWithArcadeExpressionSample.java
```diff
@@ -0,0 +1,198 @@
...
+            // wait for the expression to finish evaluating before attempting to access it
+            resultFuture.addDoneListener(() -> {
+              if (resultFuture.isDone()) {
+                try {
+                  // get result from async method and cast to int
+                  var arcadeEvaluationResult = resultFuture.get();
+                  var crimesCount = (int) ((double) arcadeEvaluationResult.getResult());
+
+                  // add data from arcade evaluation to callout
+                  mapView.getCallout().setDetail(String.valueOf(crimesCount));
```
Rachael-E: suggestions above would mean this could be

```suggestion
                  mapView.getCallout().setDetail(Math.round(crimesCount) + calloutText);
```

search/query-features-with-arcade-expression/src/main/java/com/esri/samples/query_features_with_arcade_expression/QueryFeaturesWithArcadeExpressionSample.java
```diff
@@ -0,0 +1,198 @@
...
+            // evaluate the arcade expression asynchronously
+            var resultFuture = arcadeEvaluator.evaluateAsync(hashMap);
+
+            // wait for the expression to finish evaluating before attempting to access it
+            resultFuture.addDoneListener(() -> {
+              if (resultFuture.isDone()) {
+                try {
+                  // get result from async method and cast to int
+                  var arcadeEvaluationResult = resultFuture.get();
+                  var crimesCount = (int) ((double) arcadeEvaluationResult.getResult());
```
Rachael-E: to save this double cast, use Math.round when you are later calling the crime count.

or use it here

```suggestion
                  var crimesCount = Math.round((double) arcadeEvaluationResult.getResult());
```

search/query-features-with-arcade-expression/src/main/java/com/esri/samples/query_features_with_arcade_expression/QueryFeaturesWithArcadeExpressionSample.java
```diff
@@ -0,0 +1,198 @@
...
+                }
+              }});
+          } else {
+            mapView.getCallout().setVisible(false);
+          }
+        } catch (Exception e) {
+          // resend any exceptions from the method body to be dealt with further up the call stack
+          throw new RuntimeException(e);
+        }
+      }});
```
Rachael-E: break up curly brace cluster here too

search/query-features-with-arcade-expression/src/main/java/com/esri/samples/query_features_with_arcade_expression/QueryFeaturesWithArcadeExpressionSample.java
```diff
@@ -0,0 +1,198 @@
...
+      // create a map view and set the map to it
+      mapView = new MapView();
+      mapView.setMap(map);
+
+      // set callout properties that never change
+      mapView.getCallout().setLeaderPosition(Callout.LeaderPosition.BOTTOM);
+      mapView.getCallout().setTitle("RPD Beats");
+
+      // wait for the map to finish loading data from portal before accessing its layers
+      map.addDoneLoadingListener(() -> {
```
Rachael-E: after going through this, there is a neater way to handle this that saves a bit of repeated code here too, see below suggestion.

I think the addition of a `ProgressIndicator` would be good here, as I've noticed the web map can take a while to load sometimes. You can instantiate the progress indicator after the api key set up (you shouldn't need to set it's visibility to true as that is its default value). Then once the map has loaded and you have the policeBeatsLayer, set it to invisible. Then add it to the stackpane before the mapView. 

```java
      // wait for the map to finish loading data from portal before accessing its layers
        if (map.getLoadStatus() == LoadStatus.LOADED) {
          
          for (Layer operationalLayer : map.getOperationalLayers()) {
            // if the operational layer is the RPD Beats layer, show it on the map and hide the progress indicator
            if (operationalLayer.getName().equals(rpdBeatsLayerName)) {
              policeBeatsLayer = operationalLayer;
              operationalLayer.setVisible(true);
            } else {
              // if the operational layer is any layer other than the RPD Beats layer, don't show it.
              operationalLayer.setVisible(false);
            }
          }

```

Rachael-E (changes requested):
thanks @01smito01 , very nice sample here, it's great to see another sample with real data sitting behind. Also cool to see an Arcade expression used outside of ArcGIS Pro and ArcGIS Online! 

I've suggested an alternative layout for the call out, and a couple of other code refactoring suggestions that will help with some of the feedback requests you had for this sample. There are also some tiny typos sprinkled throughout for good measure :D

search/query-features-with-arcade-expression/src/main/java/com/esri/samples/query_features_with_arcade_expression/QueryFeaturesWithArcadeExpressionSample.java
01smito01: I should have removed this - had it in when I was experimenting with callout designs because the Qt has it, but I don't think the other implementations use it and IMO is fine to omit.

search/query-features-with-arcade-expression/src/main/java/com/esri/samples/query_features_with_arcade_expression/QueryFeaturesWithArcadeExpressionSample.java
01smito01: Have added this if/else block although the progress indicator would disappear without it - presumably the map was being drawn over it once it loaded?

01smito01:
Thanks for the review @Rachael-E, have implemented most of the changes now. I've more or less left the callout as it was before, although I've added the `...` to the detail while waiting for the number to load. I didn't think the title of "RPD Beats" really added much to the callout so I've dropped that for now - can add it back in if you feel strongly about it.

At any rate, the readme passes the checker now!

.github/scripts/ci/readme_metadata_style_check/README_style_checker.py
```diff
@@ -17,6 +17,7 @@
     'ArcGIS Online',
     'OAuth',
     'Web Mercator',
+    'Arcade',
```
Rachael-E, line 20: nice: glad this worked!

map/query-features-with-arcade-expression/src/main/java/com/esri/samples/query_features_with_arcade_expression/QueryFeaturesWithArcadeExpressionSample.java
```diff
@@ -0,0 +1,197 @@
...
+            if (operationalLayer.getName().equals(rpdBeatsLayerName)) {
+              policeBeatsLayer = operationalLayer;
+              operationalLayer.setVisible(true);
+              progressIndicator.setVisible(false);
+            } else {
+              // if the operational layer is any layer other than the RPD Beats layer, don't show it.
+              operationalLayer.setVisible(false);
+            }
+          }
+          // if map clicked, evaluate an arcade expression at that point
```
Rachael-E: new line here for readability
```suggestion
          }
          
          // if map clicked, evaluate an arcade expression at that point
```

map/query-features-with-arcade-expression/src/main/java/com/esri/samples/query_features_with_arcade_expression/QueryFeaturesWithArcadeExpressionSample.java
```diff
@@ -0,0 +1,197 @@
...
+                              var crimesCount = Math.round((double) arcadeEvaluationResult.getResult());
+
+                              // add data from arcade evaluation to callout
+                              mapView.getCallout().setDetail(String.valueOf(crimesCount));
+                            } catch (Exception e) {
+                              throw new RuntimeException(e);
+                            }
+                          }
+                        });
+                      } else {
```
Rachael-E: new line here too 

```suggestion
                        });
                        
                      } else {
```

map/query-features-with-arcade-expression/src/main/java/com/esri/samples/query_features_with_arcade_expression/QueryFeaturesWithArcadeExpressionSample.java
```diff
@@ -0,0 +1,197 @@
...
+                        mapView.getCallout().setVisible(false);
+                      }
+                    } catch (Exception e) {
+                      throw new RuntimeException(e);
+                    }
+                  }
+                });
+            }
```
Rachael-E: something's off with the formatting here: if you do `ctrl-alt-l` (lower case L) IntelliJ will auto correct format for you

map/query-features-with-arcade-expression/src/main/java/com/esri/samples/query_features_with_arcade_expression/QueryFeaturesWithArcadeExpressionSample.java
```diff
@@ -0,0 +1,197 @@
...
+          });
+        } else if (map.getLoadStatus() == LoadStatus.FAILED_TO_LOAD) {
```
Rachael-E: new line here too
```suggestion
          });
          
        } else if (map.getLoadStatus() == LoadStatus.FAILED_TO_LOAD) {
```

map/query-features-with-arcade-expression/src/main/java/com/esri/samples/query_features_with_arcade_expression/QueryFeaturesWithArcadeExpressionSample.java
```diff
@@ -0,0 +1,197 @@
...
+                              mapView.getCallout().setDetail(String.valueOf(crimesCount));
+                            } catch (Exception e) {
+                              throw new RuntimeException(e);
+                            }
+                          }
+                        });
+                      } else {
+                        mapView.getCallout().setVisible(false);
+                      }
+                    } catch (Exception e) {
```
Rachael-E: new line here too

```suggestion
                      }
                      
                    } catch (Exception e) {
```

map/query-features-with-arcade-expression/src/main/java/com/esri/samples/query_features_with_arcade_expression/QueryFeaturesWithArcadeExpressionSample.java
Rachael-E: I think we could add an additional comment in here to help the user keep track of what condition this is handling

```suggestion
                      } else {
                      // if no data is found hide the callout

```

map/query-features-with-arcade-expression/src/main/java/com/esri/samples/query_features_with_arcade_expression/QueryFeaturesWithArcadeExpressionSample.java
```diff
@@ -0,0 +1,197 @@
...
+      var portal = new Portal("https://www.arcgis.com/");
+      var portalItem = new PortalItem(portal, "539d93de54c7422f88f69bfac2aebf7d");
+      map = new ArcGISMap(portalItem);
+
+      // create a map view and set the map to it
+      mapView = new MapView();
+      mapView.setMap(map);
+
+      // set callout leader to appear at the bottom of the callout
+      mapView.getCallout().setLeaderPosition(Callout.LeaderPosition.BOTTOM);
```
Rachael-E: we can instantiate the `Callout` here to save repeat usages of `mapView.getCallout()`, and apply that to the code where the detail and title are being set

```suggestion
      // set callout leader to appear at the bottom of the callout
      Callout callout = mapView.getCallout();
      callout.setLeaderPosition(Callout.LeaderPosition.BOTTOM)
```

map/query-features-with-arcade-expression/src/main/java/com/esri/samples/query_features_with_arcade_expression/QueryFeaturesWithArcadeExpressionSample.java
```diff
@@ -0,0 +1,197 @@
...
+                        hashMap.put("$map", mapView.getMap());
+
+                        // evaluate the arcade expression asynchronously
+                        ListenableFuture<ArcadeEvaluationResult> resultFuture = arcadeEvaluator.evaluateAsync(hashMap);
+
+                        // wait for the expression to finish evaluating before attempting to access it
+                        resultFuture.addDoneListener(() -> {
+                          if (resultFuture.isDone()) {
+                            try {
+                              // get result from async method and cast to int
```
Rachael-E: this comment can be updated

Rachael-E (commented):
thanks for making those changes @01smito01 . mostly styling comments remaining, a couple of other small things. once these are addressed this is ready for the 2nd review. Thanks!

Rachael-E (approved):
thanks @01smito01 this is ready for a second review now. @jenmerritt please could you have a look? Thanks!

map/query-features-with-arcade-expression/src/main/java/com/esri/samples/query_features_with_arcade_expression/QueryFeaturesWithArcadeExpressionSample.java
```diff
@@ -0,0 +1,200 @@
...
+          // if map clicked, evaluate an arcade expression at that point
+          mapView.setOnMouseClicked(event -> {
+            if (event.getButton() == MouseButton.PRIMARY && event.isStillSincePress()) {
+              // get the location of the click on the screen
+              Point2D screenPoint = new Point2D(event.getX(), event.getY());
+
+              // convert the click location from screen coordinates to map coordinates
+              Point mapPoint = mapView.screenToLocation(screenPoint);
+
+              var identifyLayerResultFuture = mapView.identifyLayerAsync(policeBeatsLayer,
```
jenmerritt, line 110:
```suggestion
              // identify any features on the layer at the clicked location
              var identifyLayerResultFuture = mapView.identifyLayerAsync(policeBeatsLayer,
```

map/query-features-with-arcade-expression/src/main/java/com/esri/samples/query_features_with_arcade_expression/QueryFeaturesWithArcadeExpressionSample.java
```diff
@@ -0,0 +1,200 @@
...
+              });
+            }
+          });
+
+        } else if (map.getLoadStatus() == LoadStatus.FAILED_TO_LOAD) {
+          new Alert(Alert.AlertType.ERROR, "Map failed to load").show();
+        }
+      });
+
+      // add the map view to the stack pane
```
jenmerritt:
```suggestion
      // add the map view and progress indicator to the stack pane
```

jenmerritt (approved):
Looks great @01smito01 , nice work.

PR #699, opened 2022-07-26
Apply Unique Values sample patch

Rachael-E:
This PR addresses a bug which was causing garbage collection of the service feature table before it could render on the map. The service feature table is now set as a member variable to keep it in scope, and the creation of the feature layer from the service feature table has been moved into the done loading listener of the table.

This PR also makes use of the CSS style of setting the label colour in this sample and another sample: when using COLOR.WHITE the sample viewer build will not recognise that and will default to original css behaviour (black label). Using CSS styling gets over this issue.

@Luiy0 could you review please? thanks! tagging @01smito01 for info with garbage collection practice :D


Luiy0 (approved):
@Rachael-E Thanks for the bug fixes. Sample runs successfully after implementing these changes locally.

PR #700, opened 2022-07-27
[New Sample] Create Mobile Geodatabase

Luiy0:
### Description
Implementation of the Create Mobile Geodatabase Sample. 

In general, the sample works as expected however, I'd like to make the following observations: 

- The implementation in other platforms includes an option for sharing the geodatabase. I am unsure on how to add this functionality. Consequently, I removed the references to sharing the geodatabase and constrained the sample to saving the file in the user's directory.
- I have not verified the contents of the geodatabase with ArcGISPro but the file is being constantly closed and loaded again whenever the button is pressed. The table view updates accordingly and I assume the geodatabase does as well.
- When the user closes the main window, the window containing the table remains open and while it can be closed manually, perhaps it would be better if both of them closed at the same time.
- To create the table view, a new class `GeoFeature` was created to facilitate this. I am getting a warning because the get methods of this class are not being explicitly used but are however required. I included an annotation to this class.
- To control the behavior of the window containing the table view, I created a Boolean variable that would store the status of the window and also to avoid null references when using the `isShowing()` method. There might be a more elegant way of doing this.
- There is a good amount of code within the various listeners and unsure if some of it could be moved out safely and improve the general structure or readability of the code. 

Thanks in advance for any suggestions.

Checklist:
- [x] Request a reviewer
- [x] Assign a reviewer
- [x] Tag reviewer in comment


feature_layers/create-mobile-geodatabase/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+For example, a user would like to track the location of their device at various intervals to generate a heat map of the most visited locations. The user can add each location as a feature to a table and generate a mobile geodatabase. The user can then instantly share the mobile geodatabase to ArcGIS Pro to generate a heat map using the recorded locations stored as a geodatabase feature table.
+
+## How to use the sample
+
+Tap on the map to add a feature symbolizing the user's location. Tap "View Table" to view the contents of the geodatabase feature table. Once you have added the location points to the map, click on "Create Mobile Geodatabase" to retrieve the .geodatabase file which can then be imported into ArcGIS Pro or opened in an ArcGIS Runtime application.
```
Rachael-E: Let's use "click" rather than "tap" here

Rachael-E: not sure we have to be so specific here about what the feature symbolises: it could be anything. We can keep it general here

```suggestion
Tap on the map to add a feature. Tap "View Table" to view the contents of the geodatabase feature table. Once you have added the features to the map, click on "Create Mobile Geodatabase" to retrieve the .geodatabase file which can then be imported into ArcGIS Pro or opened in an ArcGIS Runtime application.
```

feature_layers/create-mobile-geodatabase/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+![CreateMobileGeodatabase](CreateMobileGeodatabase.png)
+
+## Use case
+
+A mobile geodatabase is a collection of various types of GIS datasets contained in a single file (.geodatabase) on disk that can store, query, and manage spatial and non-spatial data. Mobile geodatabases are stored in a SQLite database and can contain up to 2 TB of portable data. Users can create, edit and share mobile geodatabases across ArcGIS Pro, ArcGIS Runtime, or any SQL software. These mobile geodatabases support both viewing and editing and enable new offline editing workflows that don’t require a feature service.
+
+For example, a user would like to track the location of their device at various intervals to generate a heat map of the most visited locations. The user can add each location as a feature to a table and generate a mobile geodatabase. The user can then instantly share the mobile geodatabase to ArcGIS Pro to generate a heat map using the recorded locations stored as a geodatabase feature table.
+
+## How to use the sample
```
Rachael-E, line 13: I think when the sample loads it would be better if the "Create Mobile Geodatabase" was greyed out. It's a bit confusing when I click to add features and I click that button, I get a notification that the geodatabase has been created successfully and saved in the directory, when it already exists and has already been created. 

I think two approaches would make this sample clearer here (it does deviate from the design but results in a more clear workflow to the user):

1. On sample start up, don't create the mobile geodatabase and don't allow interaction with the mapview. Prompt the user to click "Create Mobile Geodatabase" to create and save the mobile geodatabase. 
2. When it has been created, then allow the user to add features to the map and to view the table generated.
3. Add a new button (or change the button text of the "Create Mobile Geodatabase") that when clicked shows the Information dialog where the geodatabase is stored "Show mobile geodatabase location" or similar.

Alternatively:
1. Load the sample with the geodatabase being created and saved on start
2. Remove the button "Create Mobile Geodatabase" and just have the "View Table" button.

feature_layers/create-mobile-geodatabase/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+Tap on the map to add a feature symbolizing the user's location. Tap "View Table" to view the contents of the geodatabase feature table. Once you have added the location points to the map, click on "Create Mobile Geodatabase" to retrieve the .geodatabase file which can then be imported into ArcGIS Pro or opened in an ArcGIS Runtime application.
+
+## How it works
+
+1. Create the Geodatabase from the mobile geodatabase location on file using `Geodatabase.createAsync()`.
```
Rachael-E: I often refer back to the Java API doc to help with Java specific phrasing for the How it Works section https://developers.arcgis.com/java/api-reference/reference/com/esri/arcgisruntime/data/Geodatabase.html#createAsync(java.lang.String)

```suggestion
1. Create an empty mobile geodatabase with `Geodatabase.createAsync(String path)` and pass in a specific path as the parameter.
```

feature_layers/create-mobile-geodatabase/README.md
```diff
@@ -0,0 +1,39 @@
...
+2. Create a new TableDescription and add the list of FieldDescriptions to the table description.
```
Rachael-E: Any time we refer to our API class names, they should be in back ticks

```suggestion
2. Create a new `TableDescription` and add a list of new `FieldDescription`s to its collection of field descriptions with `tableDescription.getFieldDescriptions().addAll()`.
```

feature_layers/create-mobile-geodatabase/README.md
```diff
@@ -0,0 +1,39 @@
...
+3. Create a GeodatabaseFeatureTable in the geodatabase from the TableDescription using `Geodatabase.createTableAsync()`.
```
Rachael-E: For Java this isn't strictly true: in this step we are creating a ListenableFuture with a GeodatabaseFeatureTable. We wait for the result of that listenable future, get its result, and that result is the FeatureTable. Here I am referring back to the Java API doc to get a better sense for how we can make this clearer to the user here https://developers.arcgis.com/java/api-reference/reference/com/esri/arcgisruntime/data/Geodatabase.html#createTableAsync(com.esri.arcgisruntime.data.TableDescription)

```suggestion
3. Create a new table in the geodatabase from the table description with `geodatabase.createTableAsync()`.  Once the listenable future has completed, get the newly created geodatabase feature table with `geodatabaseFeatureTableFuture.get()`.
```

feature_layers/create-mobile-geodatabase/README.md
```diff
@@ -0,0 +1,39 @@
...
+4. Create a feature on the selected map point using `GeodatabaseFeatureTable.createFeature(featureAttributes, normalizedMapPoint)`.
```
Rachael-E: including the parameters in here can make readability a bit difficult and makes it very specific to the code rather than keeping the "How it works" section general, which is what we're aiming for: instead we can plain english those parameters a little and the user will refer to the code to see the details
```suggestion
4. Create a feature on the selected map point using `geodatabaseFeatureTable.createFeature()`, passing a map of feature attributes and a geometry as parameters.
```

feature_layers/create-mobile-geodatabase/README.md
```diff
@@ -0,0 +1,39 @@
...
+5. Add the feature to the table using `GeodatabaseFeatureTable.addFeatureAsync(feature)`.
```
Rachael-E:
```suggestion
5. Add the feature to the table using `geodatabaseFeatureTable.addFeatureAsync(feature)`.
```

feature_layers/create-mobile-geodatabase/README.md
```diff
@@ -0,0 +1,39 @@
...
+6. Each feature added to the GeodatabaseFeatureTable is committed to the mobile geodatabase file.
```
Rachael-E: since we have mentioned geodatabase feature table before now, we can plain english it here (that's a general rule to follow: if it's been "instantiated" in the How it works before, it's ok to refer to it either as plain english in the text or in camel case as a variable in code back ticks).

```suggestion
6. Each feature added to the GeodatabaseFeatureTable is committed to the mobile geodatabase file.
```
```suggestion
6. Each feature added to the geodatabase feature table is committed to the mobile geodatabase file.
```

feature_layers/create-mobile-geodatabase/README.md
```diff
@@ -0,0 +1,39 @@
...
+7. Close the mobile geodatabase to save the ".geodatabase" file using `Geodatabase.close()`.
```
Rachael-E: wondering if this is strictly true looking at the Java API doc
https://developers.arcgis.com/java/api-reference/reference/com/esri/arcgisruntime/data/Geodatabase.html#close()

When features are added, that is "saving" those features to the geodatabase. When the geodatabase is closed, it's just ensuring it's safe to delete or move the geodatabase without compromising the data you are working with. 

The design has the following which works, but since for Java we're not going into the share workflow as Android is, let's keep it at an overview level

"Close the mobile geodatabase to safely share the ".geodatabase" file using Geodatabase.close()"


```suggestion
7. Close the mobile geodatabase with `geodatabase.close()`.
```

feature_layers/create-mobile-geodatabase/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Relevant API
+
+* ArcGISFeature
+* FeatureLayer
+* FeatureTable
```
Rachael-E: we don't use FeatureTable in the sample, let's remove this (we're just using geodatabase feature table). feature table is used as a tag below so will come up in search results anyway

feature_layers/create-mobile-geodatabase/README.md
```diff
@@ -0,0 +1,39 @@
...
+* FieldDescription
+* Geodatabase
+* GeodatabaseFeatureTable
+* TableDescription
+
+## Tags
+
+arcgis feature, arcgis pro, database, feature, feature layer, feature table, field description, geodatabase, geodatabase feature table, mobile geodatabase, sqlite, table description
```
Rachael-E: there are a number of additional tags in here compared with the design, and a lot are plain english version of the relevant APIs so aren't serving an additional search functionality purpose: I don't think we need all of these, let's stick with the design here

```suggestion
arcgis pro, database, feature, feature table, geodatabase, mobile geodatabase, sqlite
```

feature_layers/create-mobile-geodatabase/README.md
```diff
@@ -0,0 +1,39 @@
...
+## Relevant API
+
+* ArcGISFeature
+* FeatureLayer
+* FeatureTable
+* FieldDescription
+* Geodatabase
+* GeodatabaseFeatureTable
+* TableDescription
+
```
Rachael-E, line 35: we're missing the additional information section here from the design

feature_layers/create-mobile-geodatabase/README.metadata.json
```diff
@@ -0,0 +1,43 @@
...
+    "images": [
+        "CreateMobileGeodatabase.png"
+    ],
+    "keywords": [
+        "arcgis feature",
+        "arcgis pro",
+        "database",
+        "feature",
+        "feature layer",
+        "feature table",
```
Rachael-E: this can be updated once the tags are refined as above

feature_layers/create-mobile-geodatabase/src/main/java/com/esri/samples/create_mobile_geodatabase/CreateMobileGeodatabaseSample.java
```diff
@@ -0,0 +1,423 @@
...
+
+      // create a point located at Redlands, CA to be used as the viewpoint for the map
+      var point = new Point(-117.195800, 34.056295, SpatialReferences.getWgs84());
+      mapView.setViewpointCenterAsync(point, 10000);
+
+      // create vbox, label, and button UI components
+      setUpControlsVBox();
+
+      // create geodatabase and feature layer from geodatabase feature table descriptions
+      createGeodatabase();
```
Rachael-E: this method is doing more than just creating the geodatabase, so we'd normally think about an alternative name here but this method can just be extracted and the code used in place of its call. We don't need to abstract it into a method here, it's still a clear, "story flow" if we include it in place. The main emphasis of this sample is how to create the geodatabase, so having that up front rather than hidden in a method would be the preferred structure.

feature_layers/create-mobile-geodatabase/src/main/java/com/esri/samples/create_mobile_geodatabase/CreateMobileGeodatabaseSample.java
```diff
@@ -0,0 +1,423 @@
...
+
+  /**
+   * Create and load a new geodatabase file and feature layer from table description fields
+   */
+  private void createGeodatabase() throws IOException {
+
+    // create a directory for the geodatabase file
+    Path directoryPath = Paths.get(System.getProperty("user.dir"), "CreateMobileGeodatabase");
+    if(!Files.exists(directoryPath))
+      Files.createDirectories(directoryPath);
```
Rachael-E: for readability, best Java practice is to keep the braces.

```suggestion
    // if the directory path doesn't already exist, create it
    if(!Files.exists(directoryPath)) {
      Files.createDirectories(directoryPath);
      }
```

feature_layers/create-mobile-geodatabase/src/main/java/com/esri/samples/create_mobile_geodatabase/CreateMobileGeodatabaseSample.java
```diff
@@ -0,0 +1,423 @@
...
+    // add the label and button to the VBos
+    vBoxControls.getChildren().addAll(label, createGeodatabaseButton, viewTableButton);
+  }
+
+  /**
+   * Create and load a new geodatabase file and feature layer from table description fields
+   */
+  private void createGeodatabase() throws IOException {
+
+    // create a directory for the geodatabase file
```
Rachael-E:
```suggestion
    // get the path for the geodatabase file
```

feature_layers/create-mobile-geodatabase/src/main/java/com/esri/samples/create_mobile_geodatabase/CreateMobileGeodatabaseSample.java
```diff
@@ -0,0 +1,423 @@
...
+        geodatabase = geodatabaseFuture.get();
+      } catch (InterruptedException | ExecutionException e) {
+      throw new RuntimeException(e);
+    }
+
+      // create a table description to store features as map points and set non-required properties to false
+      var tableDescription = new TableDescription("LocationHistory", SpatialReferences.getWgs84(), GeometryType.POINT);
+      tableDescription.setHasAttachments(false);
+      tableDescription.setHasM(false);
+      tableDescription.setHasZ(false);
```
Rachael-E: the default for these is already false (see https://developers.arcgis.com/java/api-reference/reference/com/esri/arcgisruntime/data/TableDescription.html#hasM()), so we can remove these calls

feature_layers/create-mobile-geodatabase/src/main/java/com/esri/samples/create_mobile_geodatabase/CreateMobileGeodatabaseSample.java
```diff
@@ -0,0 +1,423 @@
...
+
+      // Set up the fields for the table:
+      // FieldType.OID is the primary key of the SQLite table.
+      // FieldType.Date is a date column used to store a Calendar date.
+      // FieldDescriptions can be a SHORT, INTEGER, GUID, FLOAT, DOUBLE, DATE, TEXT, OID, GLOBAL ID, BLOB, GEOMETRY, RASTER, or XML.
```
Rachael-E: this level of documentation is a bit heavy for what we usually include in samples (and in the sample we don't make use of the FieldType.Date): a lot of this will be found in the website documentation. we can keep this a little lighter and highlight the parts that might not be instantly obvious and are helpful for readability, e.g. 

```suggestion
      // set up the fields for the table, FieldType.OID is the primary key of the SQLite table
```
the user can easily see what field types are available without having to explicitly list them

feature_layers/create-mobile-geodatabase/src/main/java/com/esri/samples/create_mobile_geodatabase/CreateMobileGeodatabaseSample.java
```diff
@@ -0,0 +1,423 @@
...
+      tableDescription.getFieldDescriptions().addAll(List.of(new FieldDescription("oid", Field.Type.OID),
+        new FieldDescription("collection_timestamp", Field.Type.TEXT)));
```
Rachael-E: whilst we're only going to use these field descriptions once, they are an important part of the set up for the main purpose of this sample. We can lay them out as variables above so they're not lost in the code here (all about readability :D )

```suggestion
      var fieldDescriptionOID = new FieldDescription("oid", Field.Type.OID);
      var fieldDescriptionText = new FieldDescription("collection_timestamp", Field.Type.TEXT);
      tableDescription.getFieldDescriptions().addAll(List.of(fieldDescriptionOID, fieldDescriptionText));
```

feature_layers/create-mobile-geodatabase/src/main/java/com/esri/samples/create_mobile_geodatabase/CreateMobileGeodatabaseSample.java
```diff
@@ -0,0 +1,423 @@
...
+
+      // add a new table to the geodatabase feature table by creating one from the table description
+      var featureTableFuture = geodatabase.createTableAsync(tableDescription);
```
Rachael-E: for readability, let's call this 

```suggestion
      var geodatabaseFeatureTableFuture = geodatabase.createTableAsync(tableDescription);
```

feature_layers/create-mobile-geodatabase/src/main/java/com/esri/samples/create_mobile_geodatabase/CreateMobileGeodatabaseSample.java
```diff
@@ -0,0 +1,423 @@
...
+import com.esri.arcgisruntime.mapping.view.Graphic;
+import com.esri.arcgisruntime.mapping.view.GraphicsOverlay;
+import com.esri.arcgisruntime.symbology.SimpleMarkerSymbol;
+import com.esri.arcgisruntime.data.QueryParameters;
+
+public class CreateMobileGeodatabaseSample extends Application{
+
+  private MapView mapView;
+  private Geodatabase geodatabase;
+  private GeodatabaseFeatureTable featureTable;
```
Rachael-E: for readability, let's call this

```suggestion
  private GeodatabaseFeatureTable geodatabaseFeatureTable;
```

feature_layers/create-mobile-geodatabase/src/main/java/com/esri/samples/create_mobile_geodatabase/CreateMobileGeodatabaseSample.java
```diff
@@ -0,0 +1,423 @@
...
+      featureTableFuture.addDoneListener(() -> {
+        try {
+          // get the result of the loaded "LocationHistory" table
+          featureTable = featureTableFuture.get();
+        } catch (InterruptedException | ExecutionException e) {
+        throw new RuntimeException(e);
+      }
+        // create a feature layer for the map using the GeodatabaseFeatureTable
+        var featureLayer = new FeatureLayer(featureTable);
+        mapView.getMap().getOperationalLayers().add(featureLayer);
```
Rachael-E: this block should be moved up into the try block: if the `get` was successful, then we want to continue the workflow in that block, not outside the try/catch

feature_layers/create-mobile-geodatabase/src/main/java/com/esri/samples/create_mobile_geodatabase/CreateMobileGeodatabaseSample.java
```diff
@@ -0,0 +1,423 @@
...
+      // add a new table to the geodatabase feature table by creating one from the table description
+      var featureTableFuture = geodatabase.createTableAsync(tableDescription);
+
+      // set up the map view to display the feature layer using the loaded [tableFuture] geodatabase feature table
+      featureTableFuture.addDoneListener(() -> {
+        try {
+          // get the result of the loaded "LocationHistory" table
+          featureTable = featureTableFuture.get();
+        } catch (InterruptedException | ExecutionException e) {
+        throw new RuntimeException(e);
```
Rachael-E: same comment goes for other try/catch blocks in this PR: we usually opt for printing of the stack trace here, or even better showing an alert e.g.

`e.printStackTrace();`

` new Alert(Alert.AlertType.ERROR, "Failed to get geodatabase feature table " + e.getMessage()).show();`

feature_layers/create-mobile-geodatabase/src/main/java/com/esri/samples/create_mobile_geodatabase/CreateMobileGeodatabaseSample.java
```diff
@@ -0,0 +1,423 @@
...
+      }
+        // create a feature layer for the map using the GeodatabaseFeatureTable
+        var featureLayer = new FeatureLayer(featureTable);
+        mapView.getMap().getOperationalLayers().add(featureLayer);
+      });
```
Rachael-E: similar to last try/catch block comment, this can all live in the try block immediately following the

`geodatabase = geodatabaseFuture.get();` call

feature_layers/create-mobile-geodatabase/src/main/java/com/esri/samples/create_mobile_geodatabase/CreateMobileGeodatabaseSample.java
```diff
@@ -0,0 +1,423 @@
...
+    });
+  }
+
+  /**
+   * Create a feature with attributes on map click and add it to the feature table
+   */
+  private void addFeature(Point normalizedMapPoint) {
```
Rachael-E: let's use this `addFeature` code in place too, without abstracting it into a method

Rachael-E (changes requested):
thanks @Luiy0 , really liking the Table view you've made in this sample, it looks great! Cool to see a new JavaFX layout in our samples :) 

I've not finished the review but there are a couple of things to address in the meanwhile whilst I finish going through. There are just a couple of README tightening suggestions, and some refactoring of the code (up to the createGeodatabase method and addFeature method): I haven't gone beyond those yet.

I've also surmised a couple of ideas about the UI /buttons as the sample is loaded: I think we could make this workflow smoother for our users with just some slight tweaks.

feature_layers/create-mobile-geodatabase/src/main/java/com/esri/samples/create_mobile_geodatabase/CreateMobileGeodatabaseSample.java
```diff
@@ -0,0 +1,423 @@
...
+    }
+
+    public String getOid() {
+      return oid.get();
+    }
+
+    public String getTimestamp() {
+      return timestamp.get();
+    }
+  }
```
Rachael-E: let's move this inner class to sit at the same level as the other classes: see e.g. StatisticalQueryGroupAndSort sample. 

The name GeoFeature is a bit close to some of our API class names: looking at the statistical query group and sort sample, I wondered if `FeatureAttributeField` could be a more appropriate class name, what do you think?

I had to tweak it slightly when extracting it to higher class level:

```java

package com.esri.samples.create_mobile_geodatabase;

import javafx.beans.property.SimpleStringProperty;

/**
 * Create a new class as data model to hold feature data.
 * Instances of this class are used to populate the TableView.
 */
public class FeatureAttributeField {
  
  private final SimpleStringProperty oid;
  private final SimpleStringProperty timestamp;

  public FeatureAttributeField(String oid, String timestamp) {
    this.oid = new SimpleStringProperty(oid);
    this.timestamp = new SimpleStringProperty(timestamp);
  }

   public String getOid() {
    return oid.get();
  }

   public String getTimestamp() {
    return timestamp.get();
  }
}


```

Rachael-E: the unused methods are also showing in the StatisicalQueryGroupAndSort: I don't really know why that happens when they are legit used. I'd just follow the pattern of e.g. the OrderByField.java class in that sample, no suppression warnings and a bit of java doc to describe what's going on.

feature_layers/create-mobile-geodatabase/src/main/java/com/esri/samples/create_mobile_geodatabase/CreateMobileGeodatabaseSample.java
```diff
@@ -0,0 +1,423 @@
...
+      mapView.setViewpointCenterAsync(point, 10000);
+
+      // create vbox, label, and button UI components
+      setUpControlsVBox();
+
+      // create geodatabase and feature layer from geodatabase feature table descriptions
+      createGeodatabase();
+
+      // create a graphics overlay to display the input points
+      GraphicsOverlay graphicsOverlay = new GraphicsOverlay();
```
Rachael-E:
```suggestion
      var graphicsOverlay = new GraphicsOverlay();
```

feature_layers/create-mobile-geodatabase/src/main/java/com/esri/samples/create_mobile_geodatabase/CreateMobileGeodatabaseSample.java
```diff
@@ -0,0 +1,423 @@
...
+      mapView.getGraphicsOverlays().add(graphicsOverlay);
+
+      // create a graphic to add the simple marker symbol in the graphics overlay
+      SimpleMarkerSymbol markerSymbol = new SimpleMarkerSymbol(SimpleMarkerSymbol.Style.CIRCLE, 0xFF000000, 10);
```
Rachael-E:
```suggestion
      var simpleMarkerSymbol = new SimpleMarkerSymbol(SimpleMarkerSymbol.Style.CIRCLE, 0xFF000000, 10);
```

feature_layers/create-mobile-geodatabase/src/main/java/com/esri/samples/create_mobile_geodatabase/CreateMobileGeodatabaseSample.java
```diff
@@ -0,0 +1,423 @@
...
+      Graphic inputsGraphic = new Graphic();
```
Rachael-E:
```suggestion
      var graphic = new Graphic();
```

feature_layers/create-mobile-geodatabase/src/main/java/com/esri/samples/create_mobile_geodatabase/CreateMobileGeodatabaseSample.java
```diff
@@ -0,0 +1,423 @@
...
+        if(tableStage != null) {
+          if (!tableStage.isShowing()){
+            showTableWindow = false;
+          }
+        }
+
+        if(showTableWindow && tableStage != null) {
+          tableStage.close();
+          showTableWindow = false;
+        }
```
Rachael-E: I think this works for handling the visibility of the table view depending on mouse clicks: we could even squirrel this logic away into its own method `handleTableWindowVisibility()` since it isn't a crucial Runtime functionality we'd want upfront.

 I'm struggling a bit here with readability: some comments could go in for describing what this boolean setting is doing. I'm also not clear where the "else" is coming in here: I thought these two if statements handling the UI would be separate to any handling of mouse clicks on the map. I'm thinking you don't need that else?

```java

        // handle visibility of the pop up table view window (TODO abstract this into new method)
        if (tableStage != null) {
          if (!tableStage.isShowing()) {
            showTableWindow = false;
          }
        }
        if (showTableWindow && tableStage != null) {
          tableStage.close();
          showTableWindow = false;
        }

        if (e.isStillSincePress() && e.getButton() == MouseButton.PRIMARY) {

        

```

feature_layers/create-mobile-geodatabase/src/main/java/com/esri/samples/create_mobile_geodatabase/CreateMobileGeodatabaseSample.java
```diff
@@ -0,0 +1,423 @@
...
+        else{
+          if (e.isStillSincePress() && e.getButton() == MouseButton.PRIMARY) {
+            // create 2D point from pointer location
+            Point2D point2D = new Point2D(e.getX(), e.getY());
```
Rachael-E:
```suggestion
            var point2D = new Point2D(e.getX(), e.getY());
```

feature_layers/create-mobile-geodatabase/src/main/java/com/esri/samples/create_mobile_geodatabase/CreateMobileGeodatabaseSample.java
```diff
@@ -0,0 +1,423 @@
...
+            // the map point should be normalized to the central meridian when wrapping around a map, so its value stays within the coordinate system of the map view
+            Point normalizedMapPoint = (Point) GeometryEngine.normalizeCentralMeridian(mapPoint);
+
+            // add a point where the user clicks on the map and update the inputs graphic geometry
+            inputs.add(normalizedMapPoint);
+            Multipoint inputsGeometry = new Multipoint(new PointCollection(inputs));
+            inputsGraphic.setGeometry(inputsGeometry);
+
+            if (inputs.size() > 0) {
+              // Load geodatabase from path to avoid null references after it has been closed
```
Rachael-E: this code is creating a new geodatabase instance everytime the user clicks: we don't want to promote this pattern, we already have an instance of the geodatabase stored in memory. The reason you get a null reference is because the Geodatabase is being closed when the createGeodatabaseButton is clicked (this is why I'm not in agreement with the design: the method call and what the method are doing are at odds to one another).

feature_layers/create-mobile-geodatabase/src/main/java/com/esri/samples/create_mobile_geodatabase/CreateMobileGeodatabaseSample.java
```diff
@@ -0,0 +1,423 @@
...
+              geodatabase = new Geodatabase(geodatabasePath.toString());
```
Rachael-E:
```suggestion
```

this wasn't as straightforward as I expected: have pushed up a new branch with alternative design to combat this

feature_layers/create-mobile-geodatabase/src/main/java/com/esri/samples/create_mobile_geodatabase/CreateMobileGeodatabaseSample.java
```diff
@@ -0,0 +1,423 @@
...
+          if (!tableStage.isShowing()){
+            showTableWindow = false;
+          }
+        }
+
+        if(showTableWindow && tableStage != null) {
+          // close table window if open
+          tableStage.close();
+          showTableWindow = false;
+        }
```
Rachael-E: this is repeated code as above: even better reason to squirrel it away into a new method :) 

```java

 /**
   * Handles visibility of the table window.
   */
  private void handleTableWindowVisibility() {

    if (tableStage != null) {
      if (!tableStage.isShowing()) {
        showTableWindow = false;
      }
    }

    if (showTableWindow && tableStage != null) {
      // close table window if open
      tableStage.close();
      showTableWindow = false;
    }
  }

```

feature_layers/create-mobile-geodatabase/src/main/java/com/esri/samples/create_mobile_geodatabase/CreateMobileGeodatabaseSample.java
```diff
@@ -0,0 +1,423 @@
...
+  }
+
+  /**
+   * Stops and releases all resources used in application.
+   */
+  @Override
+  public void stop() {
+
+    if (mapView != null) {
+      mapView.dispose();
```
Rachael-E: let's also close the geodatabase here and delete the geodatabase from the users machine on exit. Alternatively consider using the temporary directories approach to let the system handle file tidying up as we do in our other samples: this would be the preferred scenario for consistency but the approach you have here works: worth checking if this works from the sample viewer too. 

```suggestion
      mapView.dispose();
      geodatabase.close();
            try {
        Files.deleteIfExists(geodatabasePath);
      } catch (IOException e) {
        e.printStackTrace();
      }
```

feature_layers/create-mobile-geodatabase/src/main/java/com/esri/samples/create_mobile_geodatabase/CreateMobileGeodatabaseSample.java
```diff
@@ -0,0 +1,423 @@
...
+
+    public String getTimestamp() {
+      return timestamp.get();
+    }
+  }
+
+  /**
+   * Creates UI with two buttons and a label component.
+   */
+  private void setUpControlsVBox() {
```
Rachael-E: let's move this as the last method before the `stop` method since it's the one of least significance to Runtime functionality

Rachael-E (commented):
@Luiy0 here's the second part of my review. After having looked at it again, there are some alternative workflows I think we should deploy here to get over some potential pitfalls this sample currently demonstrates. I've laid out some comments/code in this PR to try and explain, and I've also pushed up a new branch where I've done some refactoring and have presented an alternative layout. Give it a spin and let me know what you think https://github.com/Esri/arcgis-runtime-samples-java/tree/rachael_create_mobile_gdb_alternative/feature_layers/create-mobile-geodatabase. 

The essence of the design is there but with a few additional safeguards in place. Since the essence of this workflow is to create a mobile geodatabase, having the sample capable of being ran just once IMO is the safest workflow here to avoid creating multiple new instances after having closed the original geodatabase: we have other samples showing how to load and work with open ones.

Luiy0:
Hi @Rachael-E, I've implemented the suggested changes and also modified the sample to the MVC layout for further review. Thanks.

feature_layers/create-mobile-geodatabase/README.md
```diff
@@ -0,0 +1,44 @@
...
+For example, a user would like to track the location of their device at various intervals to generate a heat map of the most visited locations. The user can add each location as a feature to a table and generate a mobile geodatabase. The user can then instantly share the mobile geodatabase to ArcGIS Pro to generate a heat map using the recorded locations stored as a geodatabase feature table.
+
+## How to use the sample
+
+Click on the map to add a feature. Click "View Table" to view the contents of the geodatabase feature table. Once you have added the features to the map, click on "Create Geodatabase" to create, save, and retrieve the .geodatabase file which can then be imported into ArcGIS Pro or opened in an ArcGIS Runtime application.
+
```
Rachael-E: this description needs updating now we have a slightly new workflow for the sample

feature_layers/create-mobile-geodatabase/src/main/java/com/esri/samples/create_mobile_geodatabase/CreateMobileGeodatabaseController.java
```diff
@@ -0,0 +1,326 @@
...
+      var graphic = new Graphic();
+      graphic.setSymbol(simpleMarkerSymbol);
+      graphicsOverlay.getGraphics().add(graphic);
+
+      // keep track of the points added by the user
+      List<Point> inputs = new ArrayList<>();
+
+      // create a point from user input and add it to the geodatabase feature table
+      mapView.setOnMouseClicked(e -> {
+        if(createGeodatabaseButton.isDisabled() && !isTableWindowOpen) {
```
Rachael-E: Java styling best practice here with space as below
```suggestion
        if (createGeodatabaseButton.isDisabled() && !isTableWindowOpen) {
```

feature_layers/create-mobile-geodatabase/src/main/java/com/esri/samples/create_mobile_geodatabase/CreateMobileGeodatabaseController.java
```diff
@@ -0,0 +1,326 @@
...
+                  label.setText("Number of features added: " + geodatabaseFeatureTable.getTotalFeatureCount());
+
+                } catch (InterruptedException | ExecutionException ex) {
+                  throw new RuntimeException(ex);
+                }
+              });
+            }
+          }
+        }
+        else {
```
Rachael-E: java styling again, these are better readable on the same line

```suggestion
        } else {
```

feature_layers/create-mobile-geodatabase/src/main/java/com/esri/samples/create_mobile_geodatabase/CreateMobileGeodatabaseController.java
```diff
@@ -0,0 +1,326 @@
...
+      });
+
+    } catch (Exception e) {
+      // on any error, display the stack trace.
+      e.printStackTrace();
+    }
+  }
+
+  /**
+   * create geodatabase and feature layer from geodatabase feature table descriptions
```
Rachael-E: Javadoc (e.g. class or method descriptions) has slightly different styling than inline comments, Javadoc styling is always written as full sentence style. Comment applies for other Javadoc here too

```suggestion
   * Create geodatabase and feature layer from geodatabase feature table descriptions.
   * ```

File: feature_layers/create-mobile-geodatabase/src/main/java/com/esri/samples/create_mobile_geodatabase/CreateMobileGeodatabaseController.java [hunk 41]
```diff
@@ -0,0 +1,326 @@
...
+    // get the path for the geodatabase file
+    geodatabasePath = Paths.get(System.getProperty("user.dir") + "/LocationHistory.geodatabase");
+    try {
+      Files.deleteIfExists(geodatabasePath);
+    } catch (IOException ioException) {
+      ioException.printStackTrace();
+    }
+
+    // create geodatabase from the specified mobile geodatabase file path
+    var geodatabaseFuture = Geodatabase.createAsync(geodatabasePath.toString());
```
Rachael-E: following your discovery of the sample viewer, let's add in a conditional here which checks the path exists/has a directory in it

feature_layers/create-mobile-geodatabase/src/main/java/com/esri/samples/create_mobile_geodatabase/CreateMobileGeodatabaseController.java
```diff
@@ -0,0 +1,326 @@
...
+    geodatabaseFuture.addDoneListener(() -> {
+      try {
+        // get the instance of the mobile geodatabase
+        geodatabase = geodatabaseFuture.get();
+
+        // create a table description to store features as map points and set non-required properties to false
+        var tableDescription = new TableDescription("LocationHistory", SpatialReferences.getWgs84(),
+          GeometryType.POINT);
+
+        // Set up the fields for the table. FieldType.OID is the primary key of the SQLite table.
```
Rachael-E:
```suggestion
        // set up the fields for the table. FieldType.OID is the primary key of the SQLite table.
```

feature_layers/create-mobile-geodatabase/src/main/java/com/esri/samples/create_mobile_geodatabase/CreateMobileGeodatabaseController.java
```diff
@@ -0,0 +1,326 @@
...
+        var fieldDescriptionOID = new FieldDescription("oid", Field.Type.OID);
+        var fieldDescriptionText = new FieldDescription("collection_timestamp", Field.Type.TEXT);
+        tableDescription.getFieldDescriptions().addAll(List.of(fieldDescriptionOID, fieldDescriptionText));
+
+        // add a new table to the geodatabase feature table by creating one from the table description
+        var geodatabaseFeatureTableFuture = geodatabase.createTableAsync(tableDescription);
+
+        // set up the map view to display the feature layer using the loaded [tableFuture] geodatabase feature table
```
Rachael-E: why is tableFuture in brackets here in the comment? haven't seen this notation before

feature_layers/create-mobile-geodatabase/src/main/java/com/esri/samples/create_mobile_geodatabase/CreateMobileGeodatabaseController.java
```diff
@@ -0,0 +1,326 @@
...
+        geodatabaseFeatureTableFuture.addDoneListener(() -> {
+          try {
+            // get the result of the loaded "LocationHistory" table
+            geodatabaseFeatureTable = geodatabaseFeatureTableFuture.get();
+            // create a feature layer for the map using the GeodatabaseFeatureTable
+            var featureLayer = new FeatureLayer(geodatabaseFeatureTable);
+            mapView.getMap().getOperationalLayers().add(featureLayer);
```
Rachael-E: we have access to `map` here still, so can use this in favour of the map view instance
```suggestion
            map().getOperationalLayers().add(featureLayer);
```

feature_layers/create-mobile-geodatabase/src/main/java/com/esri/samples/create_mobile_geodatabase/CreateMobileGeodatabaseController.java
```diff
@@ -0,0 +1,326 @@
...
+      try {
+        var queryResults = queryResultFuture.get();
+        queryResults.forEach(feature ->
+          // add features to the observable list
+          fieldData.add(new FeatureAttributeField(feature.getAttributes().get("oid").toString(), feature.getAttributes().get("collection_timestamp").toString())));
+
+        // create and set up a new table view to display the features in a table
+        TableView<FeatureAttributeField> table = new TableView<>();
+        table.setEditable(false);
+        table.setVisible(true);
```
Rachael-E: I think it should be possible to move these TableView properties in FXML? https://docs.oracle.com/javafx/2/fxml_get_started/fxml_tutorial_intermediate.htm

It would require setting up a new FXML file though, as in the Integrated Windows Authentication sample

feature_layers/create-mobile-geodatabase/src/main/java/com/esri/samples/create_mobile_geodatabase/CreateMobileGeodatabaseController.java
```diff
@@ -0,0 +1,326 @@
...
+
+    if (isTableWindowOpen && tableStage != null) {
+      // close table window if open
+      tableStage.close();
+      isTableWindowOpen = false;
+    }
+  }
+
+  /**
+   * Stops the animation and disposes of application resources.
```
Rachael-E: this javadoc isn't right for this sample: we should also ensure the geodatabase is closed when the application is closed here

feature_layers/create-mobile-geodatabase/src/main/java/com/esri/samples/create_mobile_geodatabase/CreateMobileGeodatabaseController.java
```diff
@@ -0,0 +1,326 @@
...
+   */
+  void terminate() {
+    if (mapView != null) {
+      mapView.dispose();
```
Rachael-E, line 343: let's close the geodatabase here: it might need some additional conditional handling if it has already been closed before the application is closed.

feature_layers/create-mobile-geodatabase/src/main/java/com/esri/samples/create_mobile_geodatabase/CreateMobileGeodatabaseSample.java
```diff
@@ -0,0 +1,65 @@
+/*
+ * Copyright 2019 Esri.
```
Rachael-E:
```suggestion
 * Copyright 2022 Esri.
```

feature_layers/create-mobile-geodatabase/src/main/java/com/esri/samples/create_mobile_geodatabase/CreateMobileGeodatabaseController.java
```diff
@@ -0,0 +1,326 @@
...
+        throw new RuntimeException(e);
+      }
+    });
+  }
+
+  /**
+   * Handles visibility of the table window.
+   */
+  @FXML
+  private void handleTableWindowVisibility() {
```
Rachael-E: does this method have to be FXML? I'm not seeing it referenced in the FXML file

feature_layers/create-mobile-geodatabase/src/main/resources/create_mobile_geodatabase/main.fxml
```diff
@@ -0,0 +1,71 @@
...
+                textAlignment="CENTER"
+                prefWidth="150"
+                disable="true"
+                onAction="#closeGeodatabase"/>
+        <Button fx:id="viewTableButton"
+                text="View table"
+                textAlignment="CENTER"
+                prefWidth="150"
+                disable="true"
+                onAction="#displayTable"/>
```
Rachael-E: for the onAction named methods, we tend to go with naming convention "handle...." so it's obvious to the user in code what the method is responsible for. Could e.g. go with the following

```suggestion
                onAction="#handleCreateGeodatabaseButton"/>
        <Button fx:id="closeGeodatabaseButton"
                text="Close geodatabase"
                textAlignment="CENTER"
                prefWidth="150"
                disable="true"
                onAction="#handleCloseGeodatabaseButton"/>
        <Button fx:id="viewTableButton"
                text="View table"
                textAlignment="CENTER"
                prefWidth="150"
                disable="true"
                onAction="#handleViewTableButton"/>
```

feature_layers/create-mobile-geodatabase/src/main/resources/create_mobile_geodatabase/main.fxml
```diff
@@ -0,0 +1,71 @@
...
+            </Background>
+        </background>
+        <StackPane.margin><Insets left="10" top="10"/></StackPane.margin>
+        <Label  fx:id="label"
+                text="Create mobile geodatabase"/>
+        <Button fx:id="createGeodatabaseButton"
+                text="Create geodatabase"
+                textAlignment="CENTER"
+                prefWidth="150"
+                onAction="#createGeodatabase"/>
```
Rachael-E: I haven't seen this lay out before in FXML but I'm liking it (our existing samples all end up on the one line). To me this makes it more readable having it split up like this: will be keen to hear what @jenmerritt thinks!

feature_layers/create-mobile-geodatabase/src/main/resources/create_mobile_geodatabase/main.fxml
```diff
@@ -0,0 +1,71 @@
...
+        <Button fx:id="closeGeodatabaseButton"
+                text="Close geodatabase"
+                textAlignment="CENTER"
+                prefWidth="150"
+                disable="true"
+                onAction="#closeGeodatabase"/>
+        <Button fx:id="viewTableButton"
```
Rachael-E: let's swap the order of these two, with the work flow in mind:
1. User creates geodatabase
2. User adds features and views them in table
3. Finally user closes database

Rachael-E: Another thought here with the above workflow: should we make it so that a user is only able to close the geodatabase once they've viewed the table (i.e. when viewTableButton is clicked, set disable to false on the closeGeodatabaseButton)? I don't have strong opinions either way: Just a thought to help guide them via UI as they use the sample. Will leave this to your discretion!

Rachael-E (commented):
great job refactoring this to use FXML @Luiy0 . I like the readability of the syntax you've gone for in the FXML, and the code is very neat and structured in this new lay out.

Most of my comments here are styling related, just a couple more things to address (I haven't noted the user.dir vs user.home/temp directory approach since we discussed that separately). Thanks!

feature_layers/create-mobile-geodatabase/src/main/resources/create_mobile_geodatabase/main.fxml
jenmerritt: There's not really a standard so either is fine. Probably makes it more readable where there are a lot of properties being set but not necessary on shorter lines as potentially then makes the whole file less readable. Just a judgement call on readability IMO.

feature_layers/create-mobile-geodatabase/README.md
```diff
@@ -12,7 +12,7 @@ For example, a user would like to track the location of their device at various
 
 ## How to use the sample
 
-Click on the map to add a feature. Click "View Table" to view the contents of the geodatabase feature table. Once you have added the features to the map, click on "Create Geodatabase" to create, save, and retrieve the .geodatabase file which can then be imported into ArcGIS Pro or opened in an ArcGIS Runtime application.
+Click on "Create mobile geodatabase" to create the geodatabase and its feature table. Click on the map to add new features to the geodatabase. Click "View Table" to view the contents of the geodatabase feature table. Once you have added the features to the map, click on "Close Geodatabase" to save the .geodatabase file which can then be imported into ArcGIS Pro or opened in an ArcGIS Runtime application.
```
Rachael-E:
```suggestion
Click "Create geodatabase" to create the geodatabase and its feature table. Click on the map to add new features to the geodatabase. Click "View Table" to view the contents of the geodatabase feature table. Once you have added the features to the map, click on "Close Geodatabase" to save the .geodatabase file which can then be imported into ArcGIS Pro or opened in an ArcGIS Runtime application.
```

feature_layers/create-mobile-geodatabase/src/main/java/com/esri/samples/create_mobile_geodatabase/CreateMobileGeodatabaseController.java
```diff
@@ -234,18 +239,17 @@ private void closeGeodatabase() {
     // handle UI
     viewTableButton.setDisable(true);
     closeGeodatabaseButton.setDisable(true);
-    createGeodatabaseButton.setDisable(true);
-    label.setText("Restart sample to begin again");
+    createGeodatabaseButton.setDisable(false);
+    label.setText("Create mobile geodatabase");
```
Rachael-E: I think at this point we might be better just hiding this label since it will be clear from the UI what to click to begin again: then the "Click map to add features" prompt appears as a helpful hint.

Rachael-E (approved):
thanks again @Luiy0 for making those changes, this sample is really in great shape now. Just two very small things to address, I'm happy to approve these now ready for the next review from @jenmerritt . Thanks!

feature_layers/create-mobile-geodatabase/src/main/java/com/esri/samples/create_mobile_geodatabase/CreateMobileGeodatabaseController.java
```diff
@@ -0,0 +1,336 @@
...
+import com.esri.arcgisruntime.mapping.view.Graphic;
+import com.esri.arcgisruntime.mapping.view.GraphicsOverlay;
+import com.esri.arcgisruntime.symbology.SimpleMarkerSymbol;
+import com.esri.arcgisruntime.data.QueryParameters;
+
+public class CreateMobileGeodatabaseController {
+
+  @FXML private MapView mapView;
+  @FXML private Label label;
+  @FXML private Button viewTableButton, createGeodatabaseButton, closeGeodatabaseButton;
```
jenmerritt: For best practice you should declare these on separate lines

feature_layers/create-mobile-geodatabase/src/main/java/com/esri/samples/create_mobile_geodatabase/CreateMobileGeodatabaseController.java
```diff
@@ -0,0 +1,336 @@
...
+
+            // the map point should be normalized to the central meridian when wrapping around a map, so its value stays within the coordinate system of the map view
+            Point normalizedMapPoint = (Point) GeometryEngine.normalizeCentralMeridian(mapPoint);
+
+            // add a point where the user clicks on the map and update the inputs graphic geometry
+            inputs.add(normalizedMapPoint);
+            Multipoint inputsGeometry = new Multipoint(new PointCollection(inputs));
+            graphic.setGeometry(inputsGeometry);
+
+            if (inputs.size() > 0) {
```
jenmerritt: Here you can use `!inputs.isEmpty()` (i.e. not empty)

feature_layers/create-mobile-geodatabase/src/main/java/com/esri/samples/create_mobile_geodatabase/CreateMobileGeodatabaseController.java
```diff
@@ -0,0 +1,336 @@
...
+    final ObservableList<FeatureAttributeField> fieldData = FXCollections.observableArrayList();
+
+    // query all the features loaded to the table
+    var queryResultFuture = geodatabaseFeatureTable.queryFeaturesAsync(new QueryParameters());
+    queryResultFuture.addDoneListener(() -> {
+      try {
+        var queryResults = queryResultFuture.get();
+        queryResults.forEach(feature ->
+          // add features to the observable list
+          fieldData.add(new FeatureAttributeField(feature.getAttributes().get("oid").toString(), feature.getAttributes().get("collection_timestamp").toString())));
```
jenmerritt:
```suggestion
          fieldData.add(
          new FeatureAttributeField(feature.getAttributes().get("oid").toString(), feature.getAttributes().get("collection_timestamp").toString())));
```
Add a new line before `new` to keep the line lengths shorter

feature_layers/create-mobile-geodatabase/src/main/java/com/esri/samples/create_mobile_geodatabase/CreateMobileGeodatabaseController.java
```diff
@@ -0,0 +1,336 @@
...
+      mapView.setOnMouseClicked(e -> {
+        if (createGeodatabaseButton.isDisabled()) {
+          if (e.isStillSincePress() && e.getButton() == MouseButton.PRIMARY) {
+            // create 2D point from pointer location
+            var point2D = new Point2D(e.getX(), e.getY());
+
+            // create a map point from 2D point
+            Point mapPoint = mapView.screenToLocation(point2D);
+
+            // the map point should be normalized to the central meridian when wrapping around a map, so its value stays within the coordinate system of the map view
```
jenmerritt:
```suggestion
            // the map point should be normalized to the central meridian when wrapping around a map, so its value 
            // stays within the coordinate system of the map view
```
Add a new line at `stays` to keep the line length shorter

feature_layers/create-mobile-geodatabase/src/main/java/com/esri/samples/create_mobile_geodatabase/CreateMobileGeodatabaseController.java
```diff
@@ -0,0 +1,336 @@
...
+        var scene = new Scene(pane, 220, 230);
+
+        // set up stage properties before display
+        tableStage = new Stage();
+        tableStage.setTitle("Features");
+        tableStage.setScene(scene);
+        tableStage.show();
+
+      } catch (InterruptedException | ExecutionException e) {
+        throw new RuntimeException(e);
```
jenmerritt: If you are throwing an exception in the method, you should always add this detail to the method signature and doc i.e.:
`private void handleDisplayTable() throws Exception` and 
```
  /**
   * Displays a new window with the table of features stored in the geodatabase feature table.
   * @throws RuntimeException
   */
```

That said, I would change this for the sake of samples to handle the Exception. Other Samples generally display an alert - as you've done elsewhere in this - so I'd follow that pattern and instead do something like:
```
var alert = new Alert(Alert.AlertType.ERROR, "Failed to query the feature table.");
alert.show();
```

There is another instance of this to tidy up too. Give me a shout if you need any more info on this.

feature_layers/create-mobile-geodatabase/src/main/java/com/esri/samples/create_mobile_geodatabase/CreateMobileGeodatabaseController.java
```diff
@@ -0,0 +1,336 @@
...
+              // add the feature to the feature table
+              var addFeatureFuture = geodatabaseFeatureTable.addFeatureAsync(feature);
+              addFeatureFuture.addDoneListener(() -> {
+                try {
+                  addFeatureFuture.get();
+                  // update the total feature count on screen if feature was added successfully
+                  label.setText("Number of features added: " + geodatabaseFeatureTable.getTotalFeatureCount());
+
+                } catch (InterruptedException | ExecutionException ex) {
+                  throw new RuntimeException(ex);
```
jenmerritt: See a comment further down for suggested tidy up for these Exceptions

feature_layers/create-mobile-geodatabase/src/main/resources/create_mobile_geodatabase/main.fxml
```diff
@@ -0,0 +1,71 @@
...
+                        <fill>
+                            <Color opacity="0.7"/>
+                        </fill>
+                    </BackgroundFill>
+                </fills>
+            </Background>
+        </background>
+        <StackPane.margin><Insets left="10" top="10"/></StackPane.margin>
+        <Label  fx:id="label"
+                wrapText="true"/>
```
jenmerritt, line 53: Although you've added wrap here, because there is a maxHeight set on the VBox, the label cannot wrap - as there is not space. E.g.:
image

Rather than over complication I'd just suggest changing the VBox maxWidth to 200 to give plenty of room.
I also propose that you have an initial/default label too so there isn't just a blank space. Something like `Click the button to start.`
image

feature_layers/create-mobile-geodatabase/src/main/resources/create_mobile_geodatabase/main.fxml
```diff
@@ -0,0 +1,71 @@
...
+        <Button fx:id="createGeodatabaseButton"
+                text="Create geodatabase"
+                textAlignment="CENTER"
+                prefWidth="150"
```
jenmerritt:
```suggestion
                maxWidth="Infinity"
```
This will make the button fill available space in the VBox

feature_layers/create-mobile-geodatabase/src/main/resources/create_mobile_geodatabase/main.fxml
```diff
@@ -0,0 +1,71 @@
...
+                onAction="#handleCreateGeodatabase"/>
+        <Button fx:id="viewTableButton"
+                text="View table"
+                textAlignment="CENTER"
+                prefWidth="150"
```
jenmerritt:
```suggestion
                maxWidth="Infinity"
```
This will make the button fill available space in the VBox

feature_layers/create-mobile-geodatabase/src/main/resources/create_mobile_geodatabase/main.fxml
```diff
@@ -0,0 +1,71 @@
...
+                disable="true"
+                onAction="#handleDisplayTable"/>
+        <Button fx:id="closeGeodatabaseButton"
+                text="Close geodatabase"
+                textAlignment="CENTER"
+                prefWidth="150"
```
jenmerritt:
```suggestion
                maxWidth="Infinity"
```
This will make the button fill available space in the VBox

feature_layers/create-mobile-geodatabase/src/main/java/com/esri/samples/create_mobile_geodatabase/CreateMobileGeodatabaseController.java
```diff
@@ -0,0 +1,336 @@
...
+  /**
+   * Closes the geodatabase, displays its directory, and updates UI components.
+   */
+  @FXML
+  private void handleCloseGeodatabase() {
+
+    // close geodatabase, table, and clear input list
+    geodatabase.close();
+    inputs.clear();
+    tableStage.close();
```
jenmerritt: Found a nice little bug - if you don't view the table before closing the Geodatabase, it is not created and so a NullPointerException occurs here. So you need to add a null check:
```
    if (tableStage != null) {
      tableStage.close();
    }
```
Remember users don't necessarily click things in the right order or do what you want them to! Always worth considering when testing :)

jenmerritt (changes requested):
A real step up @Luiy0 so well done for working through this. Give me a shout if any of the below needs more explanation or isn't clear.

Luiy0:
Hi @jenmerritt, thanks for a great review and good catch with discovering the little bug! I have implemented the suggested changes.

jenmerritt (approved):
Looks great! One final suggestion though I'll go ahead with approving.

It's something I've been picked up on before - when you're doing your catch statements, try to be consistent with variable naming as it's just a bit tidier. So e.g. either go with `e` or I think `ex` is probably most commonly used across the Samples. Or `exception`. Currently there's a mix of all 3. It doesn't really matter, but does make it marginally easier to read / looks tidier.

PR #704, opened 2022-08-05
[Sample Update] Find Place: add work around for JavaFX bug

Rachael-E:
The Find Place sample has a bug where if a user types in a location made up of more than one word, upon entering the space bar, the text in the combobox editor field will reset. This appears to be related to an unresolved JavaFX bug https://bugs.openjdk.org/browse/JDK-6199690?jql=text%20~%20%22combobox%20editor%22

This PR adds a work around, where an event filter is added to the combobox and listens for if the space key is hit. The event is consumed if so. 

There is however still a remaining bug where if the user types a location, selects it from the drop down, then clicks and drags to remove the text in the editor and types again, that text entry is also removed. This sample would benefit from a redesign using an alternative to the combobox in future.


jenmerritt (approved):
Agree this change makes the Sample more usable. Also agree some time spent on a Custom ComboBox Skin or something similar would be beneficial to this sample and potentially customers/other JavaFX implementations too.

PR #705, opened 2022-08-05
[Sample Update] Query with CQL Filters update

Rachael-E:
The service used in this sample (i.e. https://demo.ldproxy.net/daraa) started to support CQL2-JSON which has a slightly different syntax than CQL-JSON. The updated CQL2 syntax will work for this service as of 100.15.0.


jenmerritt (approved):
LGTM

PR #714, opened 2022-09-15
Remove deprecated API calls from display_information

D-R-Smith:
### Description

Update display_information sample to remove use of deprecated API call.

Checklist:
- [ ] Request a reviewer
- [ ] Assign a reviewer
- [ ] Tag reviewer in comment


display_information/identify-graphics/src/main/java/com/esri/samples/identify_graphics/IdentifyGraphicsSample.java
```diff
@@ -80,7 +80,7 @@ public void start(Stage stage) {
       mapView.getGraphicsOverlays().add(graphicsOverlay);
 
       // work with the MapView after it has loaded
-      mapView.addSpatialReferenceChangedListener(src -> addGraphicsOverlay());
+      mapView.spatialReferenceProperty().addListener(e -> addGraphicsOverlay());
```
jenmerritt: I'm not convinced why this is actually being used. The `addGraphicsOverlay` method also says it's adding 4 graphics but it's adding one. I think we could just create the graphics and add them to the graphics overlay before adding the graphics overlay to the mapview. I'll look at this sample in a bit more detail.

jenmerritt:
The deprecated `addSpatialReferenceChangedListener` has been removed but as per my comment above I'm not convinced why it was there. Otherwise I've modularized and updated a Color reference. 
@mbcoder please could you review this one and check you're happy with the changes? Thanks.

mbcoder (approved):
Yes this looks okay.  I've tried turning off my network connection with testing this to make sure we don't end up with any null pointer exceptions if the basemap doesn't load.  It was fine...

jenmerritt:
Thanks! @Luiy0 second review please

Luiy0 (approved):
All good.

PR #755, opened 2022-11-01
Testing sunTime and sunLighting properties

Luiy0:
### Description

Testing `sunTime` and `sunLighting` properties in the `realistic-lighting-and-shadows` sample. As noted by @jenmerritt who provided the implementation of the `sunTime` property, a calendar is being created and set when the time slider value changes. Although creating a new calendar every time is not efficient, it provided a straightforward way to update the bound properties. 

Checklist:
- [ ] Request a reviewer
- [ ] Assign a reviewer
- [ ] Tag reviewer in comment

### DIFF
--- a/scene/realistic-lighting-and-shadows/build.gradle
+++ b/scene/realistic-lighting-and-shadows/build.gradle
@@ -6,7 +6,7 @@ plugins {
 group = 'com.esri.samples'
 
 ext {
-    arcgisVersion = '200.0.0-3661'
+    arcgisVersion = '200.0.0-3678'
 }
 
 javafx {
--- a/scene/realistic-lighting-and-shadows/src/main/java/com/esri/samples/realistic_lighting_and_shadows/RealisticLightingAndShadowsController.java
+++ b/scene/realistic-lighting-and-shadows/src/main/java/com/esri/samples/realistic_lighting_and_shadows/RealisticLightingAndShadowsController.java
@@ -21,13 +21,13 @@ import java.util.Calendar;
 import java.util.GregorianCalendar;
 import java.util.TimeZone;
 
+import javafx.beans.binding.Bindings;
 import javafx.fxml.FXML;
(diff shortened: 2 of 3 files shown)


Luiy0:
Assigning @alan-edi for a first review please.

alan-edi (commented):
@Luiy0 Works well, but the use of `calendarProperty` seems very artificial. The code could be simplified by eliminating it and just calling either `sceneView.setSunTime` or `sceneView.sunTimeProperty().set` in the 2 places it's set. The latter would at least show that we now have a sunTimeProperty.

Alternatively we could invest time in trying to find a more elegant way of using the Property. (Maybe you've already done this and we've decided to cut our losses and gone for minimal change to the app?) If we do want to invest more time, I wonder if something like the following could be made to work:
- Change the slider to not call `updateTimeOfDay`.
- Instead pass `timeSlider.valueProperty()` as an Observable to [`Bindings.createObjectBinding​(Callable<T> func, Observable... dependencies)`](https://openjfx.io/javadoc/11/javafx.base/javafx/beans/binding/Bindings.html#createObjectBinding(java.util.concurrent.Callable,javafx.beans.Observable...)).
- I'm imagining something like:
```
    sceneView.sunTimeProperty().bind(Bindings.createObjectBinding(
        () -> {
          // call timeSlider.valueProperty.get() and create a Calendar from it here
          return calendar;
        },
        timeSlider.valueProperty));
```
I must stress that I've never used `createObjectBinding` and I'm not certain the above will work, so there's a bit of risk associated with this route.

Luiy0:
Following from @alan-edi suggestions, the `timeLabel` text property is now bound to the `sunTime` property and `sunTime` bound to the `timeSlider` value property. Sample should be ready for another look.

alan-edi (approved):
@Luiy0 Looks great, well done!

jenmerritt (approved):

PR #758, opened 2022-11-06
Testing drawStatus property 2

Luiy0:
### Description
Test `drawStatus` property binding in the following samples:

- `integrated-windows-authentication`: Binding progress indicator might not be viable as it's being set multiple times
- `filter-by-definition-expression-or-display-filter`: Binding label might not be viable as it's being set multiple times
- `list-related-features`

Checklist:
- [ ] Request a reviewer
- [ ] Assign a reviewer
- [ ] Tag reviewer in comment

### DIFF
--- a/feature_layers/filter-by-definition-expression-or-display-filter/src/main/java/com/esri/samples/filter_by_definition_expression_or_display_filter/FilterByDefinitionExpressionOrDisplayFilterSample.java
+++ b/feature_layers/filter-by-definition-expression-or-display-filter/src/main/java/com/esri/samples/filter_by_definition_expression_or_display_filter/FilterByDefinitionExpressionOrDisplayFilterSample.java
@@ -53,7 +53,7 @@ import com.esri.arcgisruntime.mapping.view.DrawStatus;
 import com.esri.arcgisruntime.mapping.view.MapView;
 
 public class FilterByDefinitionExpressionOrDisplayFilterSample extends Application {
-  
+
   private Button applyExpressionButton;
   private Button applyFilterButton;
   private Button resetButton;
@@ -103,10 +103,12 @@ public class FilterByDefinitionExpressionOrDisplayFilterSample extends Applicati
       // set the viewpoint for the map view
       mapView.setViewpointCenterAsync(startPoint, 20000);
 
-      // set up user interface and add a progress indicator
+      // set up user interface
       VBox vBox = controlsVBox();
+
+      // add a progress indicator and display it when drawing is in progress
       var progressIndicator = new ProgressIndicator();
-      progressIndicator.setVisible(true);
+      progressIndicator.visibleProperty().bind(mapView.drawStatusProperty().isEqualTo(DrawStatus.IN_PROGRESS));
 
       // create a new display filter with a name and where clause
       var displayFilter = new DisplayFilter("Damaged Trees", "req_type LIKE '%Tree Maintenance%'");
@@ -122,7 +124,7 @@ public class FilterByDefinitionExpressionOrDisplayFilterSample extends Applicati
         }
       );
 
-      // reset any existing display filter definition on the feature layer, and set a definition expression to it 
+      // reset any existing display filter definition on the feature layer, and set a definition expression to it
       applyExpressionButton.setOnAction(e -> {
         // reset the display filter definition
         featureLayer.setDisplayFilterDefinition(null);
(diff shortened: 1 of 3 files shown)


Luiy0:
Assigning @jenmerritt for a first review please.

feature_layers/filter-by-definition-expression-or-display-filter/src/main/java/com/esri/samples/filter_by_definition_expression_or_display_filter/FilterByDefinitionExpressionOrDisplayFilterSample.java
```diff
@@ -138,18 +140,15 @@ public void start(Stage stage) {
...
       // the features visible in the viewpoint extent every time the map is redrawn
       featureLayer.addDoneLoadingListener(() -> {
         if (featureLayer.getLoadStatus() == LoadStatus.LOADED) {
-          mapView.addDrawStatusChangedListener(drawStatusChangedEvent -> {
-            // if the draw status is in progress, show the progress indicator and set the count label text
-            if (drawStatusChangedEvent.getDrawStatus() == DrawStatus.IN_PROGRESS) {
-              progressIndicator.setVisible(true);
+          mapView.drawStatusProperty().addListener((observable, oldValue, newValue) -> {
+            if (newValue == DrawStatus.IN_PROGRESS) {
+              // set the count label text
```
jenmerritt, line 149: Is there anything we can do with this feature count label - it is being set to "loading" at the start and then "updating" when the mapview is drawing. You could try a String Bidning - but It may not like it if you then set the text on line 166, so not to worry if it doesn't work but could be worth a try if you haven't already. Be interesting to hear any investigations.

portal/integrated-windows-authentication/src/main/java/com/esri/samples/integrated_windows_authentication/IntegratedWindowsAuthenticationController.java
```diff
@@ -78,8 +78,8 @@ public void initialize() {
       resultsListView.setCellFactory(c -> new PortalItemInfoListCell());
 
       // hide the progress indicator when the map is finished drawing
-      mapView.addDrawStatusChangedListener(drawStatusChangedEvent -> {
-        if (drawStatusChangedEvent.getDrawStatus() == DrawStatus.COMPLETED) {
+      mapView.drawStatusProperty().addListener((observable, oldValue, newValue) -> {
+        if (newValue == DrawStatus.COMPLETED) {
```
jenmerritt, line 82: The comment on line 59 says `// show progress indicator while map is drawing` - if you remove the other instances of the progress indicator being set and bind the progress indicator to when drawing is in progress, is the behaviour the same?

feature_layers/filter-by-definition-expression-or-display-filter/src/main/java/com/esri/samples/filter_by_definition_expression_or_display_filter/FilterByDefinitionExpressionOrDisplayFilterSample.java
Luiy0: It seems bound values cannot be set. There could be a workaround to update the binding when a condition is met. The same thought could be applied to the progress indicator or other properties being set multiple times.

portal/integrated-windows-authentication/src/main/java/com/esri/samples/integrated_windows_authentication/IntegratedWindowsAuthenticationController.java
Luiy0: The behavior is not the same but I might have an idea on how to work around it.

feature_layers/filter-by-definition-expression-or-display-filter/src/main/java/com/esri/samples/filter_by_definition_expression_or_display_filter/FilterByDefinitionExpressionOrDisplayFilterSample.java
jenmerritt: I think this might work better when we have more properties to bind to - as then we could bind the value to the feature count for example, which we can't yet. So perhaps this one is best how you have it.

portal/integrated-windows-authentication/src/main/java/com/esri/samples/integrated_windows_authentication/IntegratedWindowsAuthenticationController.java
Luiy0: Was not able to find a solution to replicate the indicator behavior when binding different properties using `Bindings.or`.

jenmerritt (approved):
Ready for a second review

Luiy0:
Assigning @alan-edi for a second review please.

alan-edi (approved):
@Luiy0 Looks fine. I was unable to test `IntegratedWindowsAuthenticationController` because I don't know what portal URL or credentials to use, but that's OK provided it works for you

Luiy0:
Thanks for the review @alan-edi. You can use your usual Esri credentials and I'll share the portal URL with you.

PR #760, opened 2022-11-14
Testing mapRotation property

Luiy0:
### Description

Possible use of the `mapRotation` property in the `map-rotation` sample through binding the heading property of the compass to `mapRotation` allowing the use of the default compass constructor.

Checklist:
- [ ] Request a reviewer
- [ ] Assign a reviewer
- [ ] Tag reviewer in comment


map_view/map-rotation/src/main/java/com/esri/samples/map_rotation/MapRotationSample.java
```diff
@@ -96,8 +96,9 @@ public void start(Stage stage) {
       mapView.setEnableKeyboardNavigation(false);
       mapView.setMap(map);
 
-      // create a compass to show the current heading when rotated
-      Compass compass = new Compass(mapView);
+      // create a compass and show the current heading when the map view rotates
+      Compass compass = new Compass();
```
Rachael-E: we could use `var` here: let's keep an eye out for using var as part of the samples modernising work since we're opening up samples we wouldn't usually!

Rachael-E (changes requested):
thanks @Luiy0 ! 

With the `mapView.mapRotationProperty()` being read only, as we discussed, I'm also not convinced this sample requires the use of that new binding. It doesn't serve any additional purpose to this sample other than making the implementation of the `Compass` class slightly more verbose than necessary. 

My hunch is the new property is best implemented in the Toolkit code itself for the `Compass` class, and that this sample is better off sticking with the constructor for compass which takes in a MapView, and we remove the disabling of the keyboard navigation from it. 

```java
// remove this
mapView.setEnableKeyboardNavigation(false);


// keep this
var compass = new Compass(mapView);
```

The slider at this point probably isn't a good UI option for this behaviour with the mapView `mapRotationProperty()` being read only, I'd be recommending we remove the slider and use a text instruction label instead to prompt the user to rotate the map with the keyboard and keep this sample nice and clean.

Luiy0:
Hi @Rachael-E thanks for the review. Following from the suggested changes, I have removed the slider and created useful labels for the user. Previously, the `setViewpointRotationAsync()` was used to set the mapView rotation but I saw no difference when removing this as it's likely getting done within Compass. In relation to doing bindings within Compass, I will test it and likely to create a different issue or PR.

If these changes are approved, the readme file along with the sample image will need updating.

map_view/map-rotation/src/main/java/com/esri/samples/map_rotation/MapRotationSample.java
```diff
@@ -67,41 +66,26 @@ public void start(Stage stage) {
...
-      slider.valueProperty().addListener(e -> {
-        // rotate map view based on new value in slider
-        mapView.setViewpointRotationAsync(slider.getValue());
-      });
-
       // create a map with the streets basemap style
       map = new ArcGISMap(BasemapStyle.ARCGIS_STREETS);
 
-      // enable slider when map view is done loading
-      map.addDoneLoadingListener(() -> {
```
Rachael-E, line 84: since we've removed the done loading listener on the map, it's no longer in danger of being garbage collected before the code inside the logic completes. So we can remove its member variable assignment on line 46 and instantiate it instead on line 70 

```java
      // create a map with the streets basemap style
      ArcGISMap map = new ArcGISMap(BasemapStyle.ARCGIS_STREETS);
```

Rachael-E (commented):
this looks good @Luiy0 thanks: this keeps the sample nice and straight forward whilst also making use of the compass heading property to populate the text, very nice change. 

Like you say we'll need a new screenshot and updated readme for this before merging, I've also made a comment about a minor code change too.

map_view/map-rotation/src/main/java/com/esri/samples/map_rotation/MapRotationSample.java
```diff
@@ -67,41 +66,26 @@ public void start(Stage stage) {
...
+      // create labels for instructions and map rotation
+      Label instructionsLabel = new Label("Press the A and D keys to rotate the map.");
+      instructionsLabel.setTextFill(Color.WHITE);
+      Label rotationLabel = new Label("Current rotation: ");
+      rotationLabel.setTextFill(Color.WHITE);
+
+      // update the rotation label when the compass rotates
+      rotationLabel.textProperty().bind(Bindings.createStringBinding(()-> {
+        return "Current map rotation: " + Math.round(compass.headingProperty().get()) + "º";
+      }, compass.headingProperty()));
```
Rachael-E: based on IntelliJ suggestion to replace statement lambda with expression lambda: good for readability
```suggestion
      rotationLabel.textProperty().bind(Bindings.createStringBinding(()-> 
        "Current map rotation: " + Math.round(compass.headingProperty().get()) + "º", compass.headingProperty()));
```

Luiy0:
Hi @Rachael-E, it occurred to me that one possible reason for using `setViewpointRotationAsync()` in the sample is to showcase how to change the map rotation angle regardless of it already being done within the `Compass` class. The user might need to see the `Compass` implementation to realize that `setViewpointRotationAsync()` is being used. I updated the readme to somewhat highlight what's happening within `Compass`.

Additionally, binding the `rotationLabel` to the `mapRotationProperty` works well instead of binding to the compass `headingProperty` allowing us to showcase the new property.

map_view/map-rotation/src/main/java/com/esri/samples/map_rotation/MapRotationSample.java
```diff
@@ -67,41 +65,25 @@ public void start(Stage stage) {
...
+      // create labels for instructions and map rotation
+      Label instructionsLabel = new Label("Press the A and D keys to rotate the map.");
+      instructionsLabel.setTextFill(Color.WHITE);
+      Label rotationLabel = new Label("Current rotation: ");
+      rotationLabel.setTextFill(Color.WHITE);
 
-      // clicking the compass sets the map's heading to 0.0 (north) and the compass is hidden
-      // add a listener to reset the slider when this happens
-      compass.setOnMouseClicked(e -> slider.setValue(0.0));
+      // update the rotation label when the compass rotates
```
Rachael-E:
```suggestion
      // update the rotation label when the map rotation property changes
```

map_view/map-rotation/README.md
```diff
@@ -10,12 +10,12 @@ A user may wish to view the map in an orientation other than north-facing.
...
 ## How to use the sample
 
-Use the slider to rotate the map. If the map is not pointed north, the compass will display the current heading. Click the compass to set the map's heading to north.
+Press the A and D keys to rotate the map. If the map is not pointed north, the compass will display the current heading. Click the compass to set the map's heading to north.
 
 ## How it works
 
 1. Create a `ArcGISMap` and set it to a `MapView`.
-2. Use `setViewpointRotationAsync(...)` to set the rotation angle of the map.
+2. Create a `Compass` and pass the `MapView` to set the rotation angle of the map to the compass heading value.
```
Rachael-E: let's break this up a little bit. I thought this sample's main point was to show the compass but I see it's primarily designed to show the map rotation functionality, so let's mention the compass as a last point. 

```suggestion
2. Use `mapView.mapRotationProperty().get()` to get the map's rotation angle.
3. Create a `new Compass(mapView)` to show the direction of north.
```

map_view/map-rotation/README.md
```diff
@@ -10,12 +10,12 @@ A user may wish to view the map in an orientation other than north-facing.
...
 
 ## Relevant API
```
Rachael-E, line 22: under `#Tags` let's add `toolkit` here, same as we do for the Scalebar sample. This would need a metadata update too

map_view/map-rotation/README.md
```diff
@@ -10,12 +10,12 @@ A user may wish to view the map in an orientation other than north-facing.
 
 ## How to use the sample
 
-Use the slider to rotate the map. If the map is not pointed north, the compass will display the current heading. Click the compass to set the map's heading to north.
+Press the A and D keys to rotate the map. If the map is not pointed north, the compass will display the current heading. Click the compass to set the map's heading to north.
```
Rachael-E:
```suggestion
Press the A and D keys to rotate the map. The text will update to display the map's rotation value. If the map is not pointed north, the compass will display the current heading. Click the compass to set the map's heading to north.
```

map_view/map-rotation/README.md
Rachael-E: we should take viewpoint out here too following the sample changes

map_view/map-rotation/src/main/java/com/esri/samples/map_rotation/MapRotationSample.java
```diff
@@ -67,41 +65,25 @@ public void start(Stage stage) {
...
 
       // create a map view and set the map to it
       mapView = new MapView();
-      // disable keyboard navigation to focus on slider/compass map rotation interaction
-      mapView.setEnableKeyboardNavigation(false);
       mapView.setMap(map);
 
-      // create a compass to show the current heading when rotated
-      Compass compass = new Compass(mapView);
+      // create a compass and show the current heading
```
Rachael-E:
```suggestion
      // create a compass to show the direction of north
```

Rachael-E (approved):
thanks again @Luiy0 , really like the direct use of the map rotation property here! Just a couple of clarifications would be useful here for the readme and code comments to tidy this up, then it looks good. Some will require the metadata json update too.

Luiy0:
Hi @jenmerritt, this one's ready for a second review please.

map_view/map-rotation/src/main/java/com/esri/samples/map_rotation/MapRotationSample.java
```diff
@@ -114,34 +96,26 @@ public void start(Stage stage) {
       // set viewpoint to the map view
       mapView.setViewpointAsync(viewpoint);
 
+      // create a vbox to add the labels
+      VBox controlsVBox = new VBox();
```
jenmerritt:
```suggestion
      var controlsVBox = new VBox();
```

map_view/map-rotation/src/main/java/com/esri/samples/map_rotation/MapRotationSample.java
```diff
@@ -67,41 +65,25 @@ public void start(Stage stage) {
...
-      // create a compass to show the current heading when rotated
-      Compass compass = new Compass(mapView);
+      // create a compass to show the direction of north
+      var compass = new Compass(mapView);
+
+      // create labels for instructions and map rotation
+      Label instructionsLabel = new Label("Press the A and D keys to rotate the map.");
+      instructionsLabel.setTextFill(Color.WHITE);
+      Label rotationLabel = new Label("Current rotation: ");
+      rotationLabel.setTextFill(Color.WHITE);
```
jenmerritt:
```suggestion
      var instructionsLabel = new Label("Press the A and D keys to rotate the map.");
      instructionsLabel.setTextFill(Color.WHITE);
      var rotationLabel = new Label("Current rotation: ");
      rotationLabel.setTextFill(Color.WHITE);
```
Let's be consistent and use `var` across the sample where it makes sense, if we've already updated in one place e.g. compass

map_view/map-rotation/README.md
```diff
@@ -10,12 +10,13 @@ A user may wish to view the map in an orientation other than north-facing.
...
 ## How to use the sample
 
-Use the slider to rotate the map. If the map is not pointed north, the compass will display the current heading. Click the compass to set the map's heading to north.
+Press the A and D keys to rotate the map. The text will update to display the map's rotation value. If the map is not pointed north, the compass will display the current heading. Click the compass to set the map's heading to north.
 
 ## How it works
 
 1. Create a `ArcGISMap` and set it to a `MapView`.
-2. Use `setViewpointRotationAsync(...)` to set the rotation angle of the map.
+2. Use `mapView.mapRotationProperty().get()` to get the map's rotation angle.
```
jenmerritt:
```suggestion
2. Use `mapView.getMapRotation()` to get the map's rotation angle.
```
We can use the getter here

map_view/map-rotation/src/main/java/com/esri/samples/map_rotation/MapRotationSample.java
```diff
@@ -67,41 +65,25 @@ public void start(Stage stage) {
...
 
-      // clicking the compass sets the map's heading to 0.0 (north) and the compass is hidden
-      // add a listener to reset the slider when this happens
-      compass.setOnMouseClicked(e -> slider.setValue(0.0));
+      // update the rotation label when the map rotation property changes
+      rotationLabel.textProperty().bind(Bindings.createStringBinding(()->
+        "Current map rotation: " + Math.round(mapView.mapRotationProperty().get()) + "º", mapView.mapRotationProperty()));
```
jenmerritt:
```suggestion
        "Current map rotation: " + Math.round(mapView.getMapRotation()) + "º", mapView.mapRotationProperty()));
```
We can use the getter here

jenmerritt (commented):
Looks good. Just a few comments. One key thing here is that you don't need to use `get()` on the property, you can just use the getter.

Also do a review of where `var` could be used such as setting up UI components like labels. It looks tidier if we're consistent, I've commented on a couple of places.

Luiy0:
Sample should be ready for another look @jenmerritt.

jenmerritt (approved):
Looks fine @Luiy0

PR #762, opened 2022-11-21
Testing loadStatus property

Luiy0:
### Description
The following three samples are tested for use of the `loadStatus` and `loadError` properties where applicable:

- `display-layer-view-state`: Update build.
- `wmts-layer`: Use property listener in replacement of the `addDoneLoadingListener`.
- `wms-layer-url`: Bind `progressIndicator` and use property listener to test the `loadError` property.

Checklist:
- [ ] Request a reviewer
- [ ] Assign a reviewer
- [ ] Tag reviewer in comment


Luiy0:
Ready for a first review @Rachael-E, thanks.

map_view/display-layer-view-state/src/main/java/com/esri/samples/display_layer_view_state/DisplayLayerViewStateSample.java
```diff
@@ -86,48 +87,50 @@ public void start(Stage stage) {
...
-            loadLayerButton.setText("Reload Layer");
-          } else {
-            new Alert(Alert.AlertType.ERROR, "Feature layer failed to load").show();
-          }
-        });
+        // add the feature layer to the map and update its visibility
+        map.getOperationalLayers().add(featureLayer);
+        featureLayer.setVisible(visibilityCheckBox.isSelected());
+
+        // update the button text when the layer load status changes
```
Rachael-E: i think I'd be tempted to keep this wrapped in an `addDoneLoadingListener` block, but check with the team about this pattern. The `addDoneLoadingListener` is such an important part and pattern of our API so let's make sure we get this right. Effectively this would be keeping it as it was before, but instead of doing the `getLoadStatus` check, we're then using the property binding.

map_view/display-layer-view-state/src/main/java/com/esri/samples/display_layer_view_state/DisplayLayerViewStateSample.java
```diff
@@ -86,48 +87,50 @@ public void start(Stage stage) {
...
+        loadLayerButton.textProperty().bind(Bindings.createStringBinding(() -> {
+          LoadStatus loadStatus = featureLayer.loadStatusProperty().get();
+          return switch (loadStatus) {
+            case LOADED -> "Reload layer";
+            case NOT_LOADED -> "Load layer";
+            case LOADING -> "Loading";
+            default -> "Error";
+          };
+        }, featureLayer.loadStatusProperty()));
```
Rachael-E:
```suggestion
        // when the feature layer has loaded, update the button text when the layer load status changes
        
        featureLayer.addDoneLoadingListener(() -> {
        loadLayerButton.textProperty().bind(Bindings.createStringBinding(() -> {
          LoadStatus loadStatus = featureLayer.loadStatusProperty().get();
          return switch (loadStatus) {
            case LOADED -> "Reload layer";
            case NOT_LOADED -> "Load layer";
            case LOADING -> "Loading";
            default -> "Error";
          };
        }, featureLayer.loadStatusProperty()));
        }
```

map_view/display-layer-view-state/src/main/java/com/esri/samples/display_layer_view_state/DisplayLayerViewStateSample.java
```diff
@@ -86,48 +87,50 @@ public void start(Stage stage) {
...
+
+        // update the button text when the layer load status changes
+        loadLayerButton.textProperty().bind(Bindings.createStringBinding(() -> {
+          LoadStatus loadStatus = featureLayer.loadStatusProperty().get();
+          return switch (loadStatus) {
+            case LOADED -> "Reload layer";
+            case NOT_LOADED -> "Load layer";
+            case LOADING -> "Loading";
+            default -> "Error";
+          };
```
Rachael-E: this is Java 17 syntax: we should be sticking with Java 11 for samples for now

ogc/wms-layer-url/src/main/java/com/esri/samples/wms_layer_url/WmsLayerUrlSample.java
```diff
@@ -70,22 +70,24 @@ public void start(Stage stage) {
...
       wmsLayer = new WmsLayer(url, wmsLayerNames);
+
+      // show a progress indicator while the layer loads
+      progressIndicator.visibleProperty().bind(wmsLayer.loadStatusProperty().isEqualTo(LoadStatus.LOADING));
+
       // load the layer and add it as an operational layer
-      wmsLayer.addDoneLoadingListener(() -> {
-        if (wmsLayer.getLoadStatus() != LoadStatus.LOADED) {
-          wmsLayer.getLoadError().printStackTrace();
+      wmsLayer.loadStatusProperty().addListener((observable, oldValue, newValue) -> {
```
Rachael-E, line 79: I'd have the same caution here with removing the `addDoneLoadingListener` but it would be good to get the rest of the team's thoughts here: I'm just not sure if this property listener is the equivalent to the `addDoneLoadingListener` to do a straight swap like this.

ogc/wmts-layer/src/main/java/com/esri/samples/wmts_layer/WmtsLayerSample.java
```diff
@@ -60,15 +60,15 @@ public void start(Stage stage) {
       // create a WMTS service from a URL
       String serviceURL = "https://sampleserver6.arcgisonline.com/arcgis/rest/services/WorldTimeZones/MapServer/WMTS";
       wmtsService = new WmtsService(serviceURL);
-      wmtsService.addDoneLoadingListener(() -> {
-        if (wmtsService.getLoadStatus() == LoadStatus.LOADED) {
+      wmtsService.loadStatusProperty().addListener((observable, oldValue, newValue) -> {
```
Rachael-E, line 63: I'm pretty sure we need the addDoneLoadingListener here - if not, it seems possible the wmtsService load status return could be "not loaded" and trigger the warning on line 71. since it's not on a new thread checking for the load having complete. Similar thoughts to above really!

ogc/wmts-layer/src/main/java/com/esri/samples/wmts_layer/WmtsLayerSample.java
```diff
@@ -60,15 +60,15 @@ public void start(Stage stage) {
...
+        if (newValue == LoadStatus.LOADED) {
           WmtsServiceInfo wmtsServiceInfo = wmtsService.getServiceInfo();
           // get the first layer's ID
           List<WmtsLayerInfo> layerInfos = wmtsServiceInfo.getLayerInfos();
           // create the WMTS layer with the LayerInfo
           WmtsLayer wmtsLayer = new WmtsLayer(layerInfos.get(0));
           map.setBasemap(new Basemap(wmtsLayer));
-        } else {
+        } else if (newValue == LoadStatus.NOT_LOADED || newValue == LoadStatus.FAILED_TO_LOAD) {
```
Rachael-E: tempted to keep this as FAILED_TO_LOAD check. if the service is LOADING then we don't need an alert, but if the service is NOT LOADED that means it hasn't triggered a load which would be a code responsibility elsewhere. This way we can keep the alert notification clear with what the status of the load is. 

```suggestion
        } else if (newValue == LoadStatus.FAILED_TO_LOAD) {
```

Rachael-E (changes requested):
thanks @Luiy0 , this is a really useful first look of where properties can be used. My initial thoughts are that of caution when it comes to removing the addDoneLoadingListener: I think it's worth checking that pattern with the wider team. My understanding is that is a different check to what the property is doing but it would be good to get that confirmed.

ogc/wmts-layer/src/main/java/com/esri/samples/wmts_layer/WmtsLayerSample.java
Luiy0: Fixed.

map_view/display-layer-view-state/src/main/java/com/esri/samples/display_layer_view_state/DisplayLayerViewStateSample.java
Luiy0: Fixed.

map_view/display-layer-view-state/src/main/java/com/esri/samples/display_layer_view_state/DisplayLayerViewStateSample.java
Luiy0: Changed.

ogc/wmts-layer/src/main/java/com/esri/samples/wmts_layer/WmtsLayerSample.java
Luiy0: Alerts now trigger only if load status is `FAILED_TO_LOAD`.

Luiy0: The first status I obtain from the property listener is `LOADING`. I am unsure if the `NOT_LOADED` status refers to something going wrong during loading or because nothing has triggered a load (or both?). If initializing the `wmtsService` variable always triggers the load, then I don't know the difference between `NOT_LOADED` and `FAILED_TO_LOAD`. This is why I used both to display the warning.

ogc/wmts-layer/src/main/java/com/esri/samples/wmts_layer/WmtsLayerSample.java
Luiy0: Changed.

map_view/display-layer-view-state/src/main/java/com/esri/samples/display_layer_view_state/DisplayLayerViewStateSample.java
Luiy0: Changed.

ogc/wmts-layer/src/main/java/com/esri/samples/wmts_layer/WmtsLayerSample.java
Rachael-E: Short answer: `NOT_LOADED` should refer to nothing has triggered a load. The `.loadAsync()` call on the `wmtsService` on line 76 triggers the load (not the variable being initialized), and `FAILED_TO_LOAD` would trigger if something went wrong during loading 😃 

Long answer:
From the API reference: https://developers.arcgis.com/java/api-reference/reference/com/esri/arcgisruntime/loadable/LoadStatus.html#NOT_LOADED

> FAILED_TO_LOAD
> The object failed to load and some functionality may not work.

> NOT_LOADED
> The object is not loaded and some functionality may not work.

A failed to load response would mean there had been an error that had stopped the load from completing successfully: I imagine getting a `NOT_LOADED` response would be the default status if a load operation hadn't been called on the object. So throwing an alert that gives the message "Failed to load" when handling both cases of "failed to load" and "not loaded" isn't appropriate since the two have slightly different reasons for their behaviour. 

You'll get the first status of `LOADING` since there is a `loadAsync` call on line 76. If you removed that for your own testing purposes, you should see the NOT_LOADED status?

map_view/display-layer-view-state/src/main/java/com/esri/samples/display_layer_view_state/DisplayLayerViewStateSample.java
```diff
@@ -86,48 +87,45 @@ public void start(Stage stage) {
...
-            featureLayer.setVisible(visibilityCheckBox.isSelected());
-            loadLayerButton.setText("Reload Layer");
-          } else {
-            new Alert(Alert.AlertType.ERROR, "Feature layer failed to load").show();
-          }
-        });
+        // update the button text when the feature layer has loaded
+        featureLayer.addDoneLoadingListener(() ->
+          loadLayerButton.textProperty().bind(Bindings.createStringBinding(() ->
+            featureLayer.loadStatusProperty().get() == LoadStatus.LOADED ? "Reload layer" : "Load layer",
```
colinanderson: I think "Reload layer" might also be appropriate for a failed layer since the in the API you would call `retryLoadAsync()`. Depends how close to the API you want your button titles to be.

ogc/wms-layer-url/src/main/java/com/esri/samples/wms_layer_url/WmsLayerUrlSample.java
colinanderson: @Rachael-E The property listener should work just as well as the done loading BUT you have to remember it will fire for each state so you might need to deal with intermediate `LOADING` state where as done loading will only happen for `LOADED` or `FAILED` to load.

ogc/wms-layer-url/src/main/java/com/esri/samples/wms_layer_url/WmsLayerUrlSample.java
```diff
@@ -70,22 +70,24 @@ public void start(Stage stage) {
...
+        if (newValue == LoadStatus.FAILED_TO_LOAD) {
           new Alert(Alert.AlertType.ERROR, "Failed to load WMS layer").show();
+          new Alert(Alert.AlertType.ERROR, wmsLayer.loadErrorProperty().get().toString()).show();
```
colinanderson: Don't you get two dialogs popping up here? I think you could combine it into one.

ogc/wmts-layer/src/main/java/com/esri/samples/wmts_layer/WmtsLayerSample.java
colinanderson: `NOT_LOADED` is the status before `loadAsync` has been called. Once you have called `loadAsync` it will go to `LOADING` and then to `LOADED` or `FAILED_TO_LOAD`. Since the property starts in `NOT_LOADED` no change event will happen until it moves to a different state so this code looks ok to me.

colinanderson (approved):
@Luiy0 this looks ok to me (I reviewed the Java code only).

@Rachael-E @Luiy0 in many (most?) cases `addDoneLoadingListener` and `loadStatusProperty.addListener` are interchangeable so I think we should keep a mix in the samples - fine to do these ones but I'm not sure we want to change all samples to use the load status property. 

One thing to be aware of is that done loading listener will fire when you add it if the loadable is already loaded where as the property will only fire its listener if it changes in the future.

Rachael-E (approved):
thanks @Luiy0 , this looks good. Definitely worth a second reviewer after all the chats we've had, just to make sure we've caught everything we discussed!

jenmerritt (approved):

PR #763, opened 2022-11-23
Testing loadStatus property 2

Luiy0:
### Description

- Remove deprecated code related to `getNewLoadStatus()`.
- A second progress indicator was added to the `browse-wfs-layers` sample that uses the `setVisible` method when the wfsFeatureTable is being populated. 

Checklist:
- [ ] Request a reviewer
- [ ] Assign a reviewer
- [ ] Tag reviewer in comment

### DIFF
--- a/ogc/browse-wfs-layers/src/main/java/com/esri/samples/browse_wfs_layers/BrowseWfsLayersSample.java
+++ b/ogc/browse-wfs-layers/src/main/java/com/esri/samples/browse_wfs_layers/BrowseWfsLayersSample.java
@@ -51,7 +51,7 @@ public class BrowseWfsLayersSample extends Application {
 
   private MapView mapView;
   private ArcGISMap map;
-  private ProgressIndicator progressIndicator;
+  private ProgressIndicator wfsServiceIndicator, wfsTableIndicator;
 
   // keep loadables in scope to avoid garbage collection
   private WfsService wfsService;
@@ -76,12 +76,14 @@ public class BrowseWfsLayersSample extends Application {
     String yourAPIKey = System.getProperty("apiKey");
     ArcGISRuntimeEnvironment.setApiKey(yourAPIKey);
 
-    // create a list view to show all of the layers in a WFS service
+    // create a list view to show all the layers in a WFS service
     ListView<WfsLayerInfo> wfsLayerNamesListView = new ListView<>();
     wfsLayerNamesListView.setMaxSize(200, 160);
 
-    // create a progress indicator
-    progressIndicator = new ProgressIndicator();
+    // create progress indicators
+    wfsServiceIndicator = new ProgressIndicator();
+    wfsTableIndicator = new ProgressIndicator();
+    wfsTableIndicator.setVisible(false);
 
     // create a map with the standard imagery basemap style
     map = new ArcGISMap(BasemapStyle.ARCGIS_IMAGERY_STANDARD);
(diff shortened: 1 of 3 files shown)


Luiy0:
Hi @Rachael-E, here's another set of samples ready for review. Thanks!

ogc/browse-wfs-layers/src/main/java/com/esri/samples/browse_wfs_layers/BrowseWfsLayersSample.java
```diff
@@ -157,10 +156,11 @@ private void updateMap(WfsLayerInfo wfsLayerInfo){
...
     // populate the table and then remove progress indicator and set the viewpoint to that of the layer's full extent when done.
-    wfsFeatureTable.populateFromServiceAsync(new QueryParameters(), false, null ).addDoneListener(()->{
-      progressIndicator.setVisible(false);
-      mapView.setViewpointGeometryAsync(wfsFeatureLayer.getFullExtent(), 50);
-    });
+    wfsFeatureTable.populateFromServiceAsync(new QueryParameters(), false, null ).addDoneListener(()->
+      mapView.setViewpointGeometryAsync(wfsFeatureLayer.getFullExtent(), 50));
+
+    // show progress indicator when feature layer is loading
+    progressIndicator.visibleProperty().bind(wfsFeatureTable.loadStatusProperty().isEqualTo(LoadStatus.LOADING));
```
Rachael-E: This new bind is listening for the status of the load status on the wfsFeatureTable. In the original code, the progressIndicator visibility is being set to false when the **`populateFromServiceAsync()` listener** had completed. 

So in this case, this call isn't interchangeable since the listeners are handling two different things. Compare java doc e.g. [populateFromServiceAsync](https://developers.arcgis.com/java/api-reference/reference/com/esri/arcgisruntime/ogc/wfs/WfsFeatureTable.html#populateFromServiceAsync(com.esri.arcgisruntime.data.QueryParameters,boolean,java.lang.Iterable))

> Returns:
a ListenableFuture that represents the result of the query; the results will replace or be appended to this table. The ListenableFuture indicates if the query isDone(); also allows cancellation. Calling get() on the returned future may throw an ExecutionException with its cause set to [IOException](https://docs.oracle.com/javase/10/docs/api/java/io/IOException.html?is-external=true) if a network request fails. 


Let's keep the progress indicator code as it was, since we're listening for the populateFromServiceAsync status return rather than the loadStatus return.

```java

    wfsFeatureTable.populateFromServiceAsync(new QueryParameters(), false, null ).addDoneListener(()->{
      mapView.setViewpointGeometryAsync(wfsFeatureLayer.getFullExtent(), 50);
    });
```


Could we use a property for the code below, where the addDoneLoadingListener is called for the switch statements? Replacing the listener with the property?

Rachael-E: I see with my above suggestion, the progressIndicator won't allow visibility changes due to the below which makes sense (and I think is the error you were running into?)

> ProgressIndicator.visible : A bound value cannot be set.

We have a couple of options here. 

1. Have two separate instances of the progress indicator, one for the loadStatus on the wfsService and another for the doneListener from the populateFromServiceAsync. 
2. Have the vBox disabled and show a message in it when the WFSService is loading. When it's finished loading, enable the vBox for user interaction. That means only then needing one ProgressIndicator to show the populateFromServiceAsync is querying (and disapppearing when done).

Rachael-E (changes requested):
thanks @Luiy0 ! Another interesting case coming up here. I've left a couple of comments for the Browse WFS Layers sample, let me know if you'd like to talk over any of it.

ogc/browse-wfs-layers/src/main/java/com/esri/samples/browse_wfs_layers/BrowseWfsLayersSample.java
Luiy0: @Rachael-E Following from option 1, I've added another progress indicator that uses the `setVisible()` method. I'm not confident that replacing the `addDoneLoadingListener` will improve anything here for the switch. We would have to check if `newValue` is `LOADED` and then do the switch.

Rachael-E (approved):
thanks @Luiy0 looks good to me. ready for 2nd review!

Luiy0:
Hi @jenmerritt, this one's ready for another look. Thanks.

jenmerritt (approved):
LGTM @Luiy0

PR #772, opened 2022-12-14
Update task to use Map

jenmerritt:
### Description
- It is best practice for an `OfflineMapTask` to be created using an `ArcGISMap` rather than a `PortalItem` to avoid a second map being created. Checked all offline samples and just this one to correct.

Checklist:
- [ ] Request a reviewer
- [ ] Assign a reviewer
- [ ] Tag reviewer in comment

### DIFF
--- a/map/download-preplanned-map/README.md
+++ b/map/download-preplanned-map/README.md
@@ -17,7 +17,7 @@ Select a map area from the Preplanned Map Areas list. Click the Download button 
 ## How it works
 
 1. Open the online `ArcGISMap` from a `PortalItem` and display it.
-2. Create an `OfflineMapTask` using the portal item.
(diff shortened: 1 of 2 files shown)


D-R-Smith (approved):
@jenmerritt looks okay to me, and running the sample looked the same before and after

Rachael-E (changes requested):
this change looks good, though we should update the readme too from:

> ## How it works

1. Open the online `ArcGISMap` from a `PortalItem` and display it.
2. Create an `OfflineMapTask` using the portal item.

to 

2. Create an `OfflineMapTask` using the map.

Rachael-E (approved):
thanks @jenmerritt , looks great!

PR #786, opened 2023-04-18
[New Sample] Create and Edit Geometries

01smito01:
### Description

Implementation of the "Create and Edit Geometries" sample. A (very non exhaustive) list of posts I could do with feedback on:
- Definition and loading of the example geometries is less than concise. I'm not particularly enthused about declaring them in the code like this, or converting them into `List<double[]>` to make their conversion into `Point`s easier. Would it be better to move these into a JSON file and (after registering a type adapter) load them into a `List<List<Double>>`? I'm not sure there's any good option here.
- A lot of variables (too many?) are declared at the class level. Would any of these be more sensible to declare within a method body?
- Similarly: each type of geometry being loaded at startup has its own method handling conversion into `Point`s, calling the relevant subclass of `GeometryBuilder`, then putting the result into a graphic and adding it to the graphics overlay. Since there's so much in common here, should these methods be merged together?
- A few buttons (those for the tools in the "Create" section, and the "Delete Selected Element") have multiple bindings that change via a listener to avoid binding to an object which could be null. Is there a way within the Bindings API to manage this fairly common situation?
- Have a play around with the sample. Are there any situations where a tool is unavailable when you wouldn't expect (or vice versa). Are there any situations where a geometry disappears without you intending to discard it?

Checklist:
- [x] Request a reviewer
- [x] Assign a reviewer
- [x] Tag reviewer in comment


01smito01:
@jenmerritt this is ready for review now.

geometry/create-and-edit-geometries/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+Use the control panel to undo or redo changes made to the geometry, delete a selected element, save the geometry, stop the editing session and discard any edits, and remove all geometries from the map.
+
+## How it works
+
+1. Create a `GeometryEditor` and set it to the MapView using `mapView.setGeometryEditor(GeometryEditor)`.
+2. Start the `GeometryEditor` using `GeometryEditor.start(GeometryType)` to create a new geometry or `GeometryEditor.start(Geometry)` to edit an existing geometry.
+    - If using the Geometry Editor to edit an existing geometry, the geometry must be retrieved from the graphics overlay being used to visualize the geometry prior to calling the start method. To do this:
```
jenmerritt: Try replacing these indented bullet points with `*` to satisfy the markdown linter check.

geometry/create-and-edit-geometries/README.md
```diff
@@ -0,0 +1,51 @@
...
+## How to use the sample
+
+To create a new geometry, press the button appropriate for the geometry type you want to create (i.e. points, multipoints, polyline, or polygon) and interactively tap and drag on the map view to create the geometry. To edit an existing geometry, tap the geometry to be edited in the map to select it and then edit the geometry by tapping and dragging elements of the geometry. If creating or editing polyline or polygon geometries, choose the desired creation/editing tool (i.e. `VertexTool` or `FreehandTool`).
+
+Use the control panel to undo or redo changes made to the geometry, delete a selected element, save the geometry, stop the editing session and discard any edits, and remove all geometries from the map.
+
+## How it works
+
+1. Create a `GeometryEditor` and set it to the MapView using `mapView.setGeometryEditor(GeometryEditor)`.
+2. Start the `GeometryEditor` using `GeometryEditor.start(GeometryType)` to create a new geometry or `GeometryEditor.start(Geometry)` to edit an existing geometry.
```
jenmerritt:
```suggestion
2. Start the `GeometryEditor` using `geometryEditor.start(GeometryType)` to create a new geometry or `geometryEditor.start(Geometry)` to edit an existing geometry.
```

geometry/create-and-edit-geometries/README.md
```diff
@@ -0,0 +1,51 @@
...
+    - If using the Geometry Editor to edit an existing geometry, the geometry must be retrieved from the graphics overlay being used to visualize the geometry prior to calling the start method. To do this:
+        - Use `MapView.identifyGraphicsOverlayAsync(...)` to identify graphics at the location of a tap.
+        - Access the `MapView.identifyGraphicsOverlayAsync(...)` results using `MapView.identifyGraphicsOverlayAsync(...).get()`.
+        - Find the desired graphic in the `MapView.identifyGraphicsOverlayAsync(...).get().getGraphics()` list.
+        - Access the geometry associated with the `Graphic` using `Graphic.getGeometry()` - this will be used in the `GeometryEditor.start(Geometry)` method.
+
```
jenmerritt:
```suggestion
        - Use `mapView.identifyGraphicsOverlayAsync(...)` to identify graphics at the location of a tap.
        - Once the identify operation has completed, get the `IdentifyGraphicsOverlayResult` using `listenableFuture.get()`.
        - Find the desired graphic in the `result.getGraphics()` list.
        - Access the geometry associated with the `Graphic` using `graphic.getGeometry()` - this will be used in the `geometryEditor.start(Geometry)` method.

```

geometry/create-and-edit-geometries/README.md
```diff
@@ -0,0 +1,51 @@
...
+3. Create `VertexTool` or `FreehandTool` objects which define how the user interacts with the view to create or edit geometries, using `geometryEditor.setTool(geometryEditorTool)`.
+4. Check to see if undo and redo are possible during an editing session using `GeometryEditor.getCanUndo()` and `GeometryEditor.getCanRedo()`. If it's possible, use `GeometryEditor.undo()` and `GeometryEditor.redo()`.
```
jenmerritt, line 28:
```suggestion
4. Check to see if undo and redo are possible during an editing session using `geometryEditor.getCanUndo()` and `geometryEditor.getCanRedo()`. If it's possible, use `geometryEditor.undo()` and `geometryEditor.redo()`.
```

geometry/create-and-edit-geometries/README.md
```diff
@@ -0,0 +1,51 @@
...
+5. Check whether the currently selected `GeometryEditorElement` can be deleted (`GeometryEditor.getSelectedElement().getCanDelete()`). If the element can be deleted, delete using `GeometryEditor.deleteSelectedElement()`.
```
jenmerritt, line 29:
```suggestion
5. Check whether the currently selected `GeometryEditorElement` can be deleted using the `GeometryEditor.getSelectedElement().canDeleteProperty()`. If the element can be deleted, delete using `geometryEditor.deleteSelectedElement()`.
```

geometry/create-and-edit-geometries/README.md
```diff
@@ -0,0 +1,51 @@
...
+6. Call `GeometryEditor.stop()` to finish the editing session. The `GeometryEditor` does not automatically handle the visualization of a geometry output from an editing session. This must be done manually by propagating the geometry returned by `GeometryEditor.stop()` into a `Graphic` and a `GraphicsOverlay`.
+    - To create a new `Graphic` in the `GraphicsOverlay`:
+        - Using `Graphic(Geometry)`, create a new Graphic with the geometry returned by the `GeometryEditor.stop()` method.
+        - Append the `Graphic` to the `GraphicsOverlay`'s `GraphicListModel` (i.e. `GraphicsOverlay.getGraphics().add(Graphic)`).
+    - To update the geometry underlying an existing `Graphic` in the `GraphicsOverlay`:
+        - Replace the existing `Graphic`'s `Geometry` property with the geometry returned by the `GeometryEditor.stop()` method using `Graphic.setGeometry(Geometry)`.
```
jenmerritt:
```suggestion
6. Call `geometryEditor.stop()` to finish the editing session. The `GeometryEditor` does not automatically handle the visualization of a geometry output from an editing session. This must be done manually by propagating the geometry returned by `geometryEditor.stop()` into a `Graphic` and a `GraphicsOverlay`.
    - To create a new `Graphic` in the `GraphicsOverlay`:
        - Create a new `Graphic` with the `Geometry` returned by the `geometryEditor.stop()` method.
        - Add the `Graphic` to the `GraphicsOverlay` using `graphicsOverlay.getGraphics().add(graphic)`.
    - To update the geometry underlying an existing `Graphic` in the `GraphicsOverlay`:
        - Update the existing `Graphic`'s `Geometry` with the geometry returned by the `geometryEditor.stop()` method using `graphic.setGeometry(geometry)`.
```

geometry/create-and-edit-geometries/README.metadata.json
```diff
@@ -0,0 +1,37 @@
...
+    "vertex",
+    "Geometry",
+    "GeometryEditor",
+    "Graphic",
+    "GraphicsOverlay",
+    "MapView"
+  ],
+  "redirect_from": [
+    "/java/latest/sample-code/create-and-edit-geometries.htm"
+  ],
```
jenmerritt:
```suggestion
  "redirect_from": "",
```
No need for a redirect

jenmerritt (commented):
Thanks @01smito01 ! Some comments on the readme. I'll continue the review on the code next.

geometry/create-and-edit-geometries/src/main/java/module-info.java
```diff
@@ -0,0 +1,28 @@
+/*
+ * Copyright 2022 Esri.
```
jenmerritt:
```suggestion
 * Copyright 2023 Esri.
```

geometry/create-and-edit-geometries/src/main/resources/create_and_edit_geometries/main.fxml
```diff
@@ -0,0 +1,169 @@
...
+                        <Image url="@trash-32.png" requestedHeight="24" preserveRatio="true" smooth="true"/>
+                    </ImageView>
+                </graphic>
+                <tooltip>
+                    <Tooltip text="Delete all geometries"/>
+                </tooltip>
+            </Button>
+        </GridPane>
+    </VBox>
+</StackPane>
```
jenmerritt:
```suggestion
</StackPane>

```

geometry/create-and-edit-geometries/src/main/resources/create_and_edit_geometries/save-32.png
jenmerritt, line 1: I don't think this image is used? If that's correct it can be deleted.

geometry/create-and-edit-geometries/src/main/resources/create_and_edit_geometries/main.fxml
```diff
@@ -0,0 +1,169 @@
...
+  ~ the License.
+  -->
+
+<?import com.esri.arcgisruntime.mapping.view.MapView?>
+<?import javafx.scene.control.*?>
+<?import javafx.scene.layout.*?>
+<?import javafx.geometry.Insets?>
+<?import javafx.scene.paint.Color?>
+<?import javafx.scene.image.ImageView?>
+<?import javafx.scene.image.Image?>
```
jenmerritt:
```suggestion
<?import com.esri.arcgisruntime.mapping.view.MapView?>
<?import javafx.geometry.Insets?>
<?import javafx.scene.control.Button?>
<?import javafx.scene.control.ComboBox?>
<?import javafx.scene.control.Label?>
<?import javafx.scene.control.Separator?>
<?import javafx.scene.control.Tooltip?>
<?import javafx.scene.image.Image?>
<?import javafx.scene.image.ImageView?>
<?import javafx.scene.layout.Background?>
<?import javafx.scene.layout.BackgroundFill?>
<?import javafx.scene.layout.GridPane?>
<?import javafx.scene.layout.StackPane?>
<?import javafx.scene.layout.VBox?>
<?import javafx.scene.paint.Color?>
```
We should expand out the imports so they're all explicit

geometry/create-and-edit-geometries/src/main/resources/create_and_edit_geometries/main.fxml
```diff
@@ -0,0 +1,169 @@
...
+                    </ImageView>
+                </graphic>
+                <tooltip>
+                    <Tooltip text="Point"/>
+                </tooltip>
+            </Button>
+            <Button fx:id="multipointButton" onAction="#handleMultipointButtonClicked" GridPane.rowIndex="0" GridPane.columnIndex="1" text="Multipoint">
+                <graphic>
+                    <ImageView>
+                        <Image url="@multipoint-32.png"/>
```
jenmerritt: Let's put the images in a subfolder "images", so it's easier to access the fxml/css file

jenmerritt (commented):
A few minor comments

geometry/create-and-edit-geometries/src/main/java/com/esri/samples/create_and_edit_geometries/CreateAndEditGeometriesController.java
```diff
@@ -0,0 +1,591 @@
...
+import javafx.util.StringConverter;
+
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class CreateAndEditGeometriesController {
+
+
+  @FXML
+  private MapView mapView;
```
jenmerritt:
```suggestion
  @FXML private MapView mapView;
```
> A lot of variables (too many?) are declared at the class level. Would any of these be more sensible to declare within a method body?

This is ok if they are needed. and can't be declared at local level. And it's always the way with FXML that there's a lot of variables.

One thing you can do is put these on one line to reduce the spacing a bit

geometry/create-and-edit-geometries/src/main/java/com/esri/samples/create_and_edit_geometries/CreateAndEditGeometriesController.java
```diff
@@ -0,0 +1,591 @@
...
+          x.printStackTrace();
+        }
+      });
+    });
+  }
+
+  /**
+   * Defines graphics to be loaded into the sample upon startup.
+   */
+  void createInitialGraphics() {
```
jenmerritt: > Definition and loading of the example geometries is less than concise. I'm not particularly enthused about declaring them in the code like this, or converting them into List<double[]> to make their conversion into Points easier. Would it be better to move these into a JSON file and (after registering a type adapter) load them into a List<List<Double>>? I'm not sure there's any good option here.

Yes go ahead. See Display Device Location with Autopan Modes for an example of us following this approach in another sample. The key here is that the purpose of this sample is to create and edit geometries using the geometry editor, so obfuscating the creation of these initial graphics is ok.

geometry/create-and-edit-geometries/src/main/java/com/esri/samples/create_and_edit_geometries/CreateAndEditGeometriesController.java
```diff
@@ -0,0 +1,591 @@
...
+    pointSymbol = new SimpleMarkerSymbol(SimpleMarkerSymbol.Style.SQUARE, Color.ORANGERED, 10);
+    // yellow circle for multipoints
+    multipointSymbol = new SimpleMarkerSymbol(SimpleMarkerSymbol.Style.CIRCLE, Color.YELLOW, 5);
+    // thin blue line for polylines
+    lineSymbol = new SimpleLineSymbol(SimpleLineSymbol.Style.SOLID, Color.BLUE, 2);
+    // black outline for polygons
+    SimpleLineSymbol polygonLineSymbol = new SimpleLineSymbol(SimpleLineSymbol.Style.DASH, Color.BLACK, 1);
+    // translucent red interior for polygons
+    fillSymbol = new SimpleFillSymbol(SimpleFillSymbol.Style.SOLID, Color.rgb(255, 0, 0, 0.3), polygonLineSymbol);
+  }
```
jenmerritt: Further to the other point about the creation of initial graphics, I would take this part out of a separate method, maybe under where you create the vertex/freehand tools, and add a comment like:
`// create symbols to use for displaying new geometries`

As a user needs to know that they have to configure these symbols for their graphics in order to display the geometries.

geometry/create-and-edit-geometries/src/main/java/com/esri/samples/create_and_edit_geometries/CreateAndEditGeometriesController.java
```diff
@@ -0,0 +1,591 @@
...
+import javafx.scene.control.ComboBox;
+import javafx.scene.paint.Color;
+import javafx.util.StringConverter;
+
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class CreateAndEditGeometriesController {
+
+
```
jenmerritt:
```suggestion
```

geometry/create-and-edit-geometries/src/main/java/com/esri/samples/create_and_edit_geometries/CreateAndEditGeometriesController.java
```diff
@@ -0,0 +1,591 @@
...
+  private VertexTool vertexTool;
+  private FreehandTool freehandTool;
+
+  private GraphicsOverlay graphicsOverlay;
+  private Graphic selectedGraphic;
+
+  private SimpleFillSymbol fillSymbol;
+  private SimpleLineSymbol lineSymbol;
+  private SimpleMarkerSymbol pointSymbol;
+  private SimpleMarkerSymbol multipointSymbol;
```
jenmerritt:
```suggestion
  private GeometryEditor geometryEditor;
  private VertexTool vertexTool;
  private FreehandTool freehandTool;
  private GraphicsOverlay graphicsOverlay;
  private Graphic selectedGraphic;
  private SimpleFillSymbol fillSymbol;
  private SimpleLineSymbol lineSymbol;
  private SimpleMarkerSymbol pointSymbol;
  private SimpleMarkerSymbol multipointSymbol;
```
These don't need to be separated out, though I understand why you did.

geometry/create-and-edit-geometries/src/main/java/com/esri/samples/create_and_edit_geometries/CreateAndEditGeometriesController.java
```diff
@@ -0,0 +1,591 @@
...
+    // set a viewpoint on the map view
+    mapView.setViewpoint(new Viewpoint(53.08230, -9.5920, 5000));
+
+    // create a graphics overlay for the graphics
+    graphicsOverlay = new GraphicsOverlay();
+
+    // add the graphics overlay to the map view
+    mapView.getGraphicsOverlays().add(graphicsOverlay);
+
+    // create a new sketch editor and add it to the map view
```
jenmerritt:
```suggestion
    // create a geometry editor and set it to the map view
```

geometry/create-and-edit-geometries/src/main/java/com/esri/samples/create_and_edit_geometries/CreateAndEditGeometriesController.java
```diff
@@ -0,0 +1,591 @@
...
+    toolComboBox.setConverter(new ComboBoxStringConverter());
+    toolComboBox.getItems().addAll(vertexTool, freehandTool);
+    toolComboBox.getSelectionModel().select(0);
+
+    // add example graphics to the map
+    createGraphicsMarkers();
+    createInitialGraphics();
+
+    // bind button availability to properties of the geometry editor
+    bindButtonsToGeometryEditor();
```
jenmerritt: I think we should put the internals of this method directly here. Totally get why you tidied it away but we should have all the properties here so it can just be followed down and not needing to switch to a method. A sample-ism. We may revise "best practice" here in the future but for now it follows previous patterns.

geometry/create-and-edit-geometries/src/main/java/com/esri/samples/create_and_edit_geometries/CreateAndEditGeometriesController.java
```diff
@@ -0,0 +1,591 @@
...
+
+    // allow user to immediately select and edit graphics
+    selectGraphic();
```
jenmerritt: Again, internals of this method should go in here.

geometry/create-and-edit-geometries/src/main/java/com/esri/samples/create_and_edit_geometries/CreateAndEditGeometriesController.java
```diff
@@ -0,0 +1,591 @@
...
+  }
+
+  /**
+   * Binds button status to geometry editor properties.
+   */
+  private void bindButtonsToGeometryEditor() {
+    // point and multipoint disabled when freehand tool selected
+    pointButton.disableProperty().bind(geometryEditor.toolProperty().isEqualTo(freehandTool));
+    multipointButton.disableProperty().bind(geometryEditor.toolProperty().isEqualTo(freehandTool));
```
jenmerritt:
```suggestion
```
We can actually remove this as the confirmed behavior here is that the combobox is only enabled once a geometry type is selected. So we can disable it by default, and then enable it if polyline/polygon are selected. More on this in another comment.

geometry/create-and-edit-geometries/src/main/java/com/esri/samples/create_and_edit_geometries/CreateAndEditGeometriesController.java
```diff
@@ -0,0 +1,591 @@
...
+        polygonButton.disableProperty().bind(new SimpleBooleanProperty(!geometryEditor.getGeometry().getGeometryType()
+          .equals(GeometryType.POLYGON)));
+      } else {
+        // when editor stopped, re-enable polygon & polyline, enable point/multipoint if vertex tool selected
+        pointButton.disableProperty().bind(geometryEditor.toolProperty().isEqualTo(freehandTool));
+        multipointButton.disableProperty().bind(geometryEditor.toolProperty().isEqualTo(freehandTool));
+        polylineButton.disableProperty().bind(new SimpleBooleanProperty(false));
+        polygonButton.disableProperty().bind(new SimpleBooleanProperty(false));
+      }
+    });
```
jenmerritt:
```suggestion
    geometryEditor.startedProperty().addListener((observable, oldValue, newValue) -> {
      if (newValue) {
        // when the geometry editor is started (and so the geometry will not be null) bind the button state to the geometry type
        pointButton.disableProperty().bind(Bindings.createBooleanBinding(() -> geometryEditor.getGeometry().getGeometryType().equals(GeometryType.POINT)).not());
        multipointButton.disableProperty().bind(Bindings.createBooleanBinding(() -> geometryEditor.getGeometry().getGeometryType().equals(GeometryType.MULTIPOINT)).not());
        polylineButton.disableProperty().bind(Bindings.createBooleanBinding(() -> geometryEditor.getGeometry().getGeometryType().equals(GeometryType.POLYLINE)).not());
        polygonButton.disableProperty().bind(Bindings.createBooleanBinding(() -> geometryEditor.getGeometry().getGeometryType().equals(GeometryType.POLYGON)).not());
        toolComboBox.disableProperty().bind(Bindings.or(
            Bindings.createBooleanBinding(() -> geometryEditor.getGeometry().getGeometryType().equals(GeometryType.MULTIPOINT)),
            Bindings.createBooleanBinding(() -> geometryEditor.getGeometry().getGeometryType().equals(GeometryType.POINT))
        ));
      } else {
        // when the geometry editor is not started (and so the geometry is null) enable all geometry buttons
        pointButton.disableProperty().unbind();
        pointButton.setDisable(false);
        multipointButton.disableProperty().unbind();
        multipointButton.setDisable(false);
        polylineButton.disableProperty().unbind();
        polylineButton.setDisable(false);
        polygonButton.disableProperty().unbind();
        polygonButton.setDisable(false);
        toolComboBox.disableProperty().unbind();
        toolComboBox.setDisable(true);
      }
    });
```
Functionally this suggestion is more or less the same, just using the Bindings API rather than creating properties, and also you can see here the combox controls too. In addition, we can remove the binding and set the state (rather than creating a "mock" property). As I'm sure you encountered, if we don't "unbind" here, then we'd get an error that a "bound value cannot be set".

jenmerritt: Also note that we probably want to revisit some of the documentation/comments, feel free to edit as you see fit.

geometry/create-and-edit-geometries/src/main/java/com/esri/samples/create_and_edit_geometries/CreateAndEditGeometriesController.java
```diff
@@ -0,0 +1,591 @@
...
+
+    // delete selected element enabled only when an element is selected and deletable
+    ChangeListener<GeometryEditorElement> selectedElementChangedListener = (ob, ov, nv) -> {
+      if (nv == null) {
+        deleteSelectedElementButton.disableProperty().bind(new SimpleBooleanProperty(true));
+      } else {
+        deleteSelectedElementButton.disableProperty().bind(geometryEditor.getSelectedElement().canDeleteProperty().not());
+      }
+    };
+    geometryEditor.selectedElementProperty().addListener(selectedElementChangedListener);
```
jenmerritt:
```suggestion
    geometryEditor.selectedElementProperty().addListener(((observable, oldValue, newValue) -> {
      if (newValue == null) {
        deleteSelectedElementButton.disableProperty().unbind();
        deleteSelectedElementButton.setDisable(true);
      } else {
        deleteSelectedElementButton.disableProperty().bind(geometryEditor.getSelectedElement().canDeleteProperty().not());
      }
    }));
```
Similarly, this is functionally the same but we can unbind and set the state. Also we can just set the listener directly as I don't think we will need to manually remove this listener (for example if we wanted to remove it later, we could use the reference in `.removeListener(...)`). I may be proved wrong here in a future review but I think it'll be garbage collected.

geometry/create-and-edit-geometries/src/main/java/com/esri/samples/create_and_edit_geometries/CreateAndEditGeometriesController.java
```diff
@@ -0,0 +1,591 @@
...
+
+    // save/discard buttons enabled when geometry editor has started
+    stopAndSaveButton.disableProperty().bind(geometryEditor.startedProperty().not());
```
jenmerritt:
```suggestion
    stopAndSaveButton.disableProperty().bind(Bindings.or(geometryEditor.startedProperty().not(), geometryEditor.canUndoProperty().not()));
```
Clarification on behavior here is that we only want to stop and save if there are changes to save, hence the additional check here.

geometry/create-and-edit-geometries/src/main/java/com/esri/samples/create_and_edit_geometries/CreateAndEditGeometriesController.java
```diff
@@ -0,0 +1,591 @@
...
+    ChangeListener<GeometryEditorElement> selectedElementChangedListener = (ob, ov, nv) -> {
+      if (nv == null) {
+        deleteSelectedElementButton.disableProperty().bind(new SimpleBooleanProperty(true));
+      } else {
+        deleteSelectedElementButton.disableProperty().bind(geometryEditor.getSelectedElement().canDeleteProperty().not());
+      }
+    };
+    geometryEditor.selectedElementProperty().addListener(selectedElementChangedListener);
+
+    // save/discard buttons enabled when geometry editor has started
```
jenmerritt: I think a good pattern for these comments on the button bindings could be something like: 

```
    // bind the disable property of the stop and discard button to the geometry editor started property
    // disable the button if the geometry editor is not started
```

geometry/create-and-edit-geometries/src/main/java/com/esri/samples/create_and_edit_geometries/CreateAndEditGeometriesController.java
```diff
@@ -0,0 +1,591 @@
...
+    stopAndSaveButton.disableProperty().bind(geometryEditor.startedProperty().not());
+    stopAndDiscardButton.disableProperty().bind(geometryEditor.startedProperty().not());
+
+    // delete all geometries button enabled when geometry editor stopped
+    deleteAllGeometriesButton.disableProperty().bind(geometryEditor.startedProperty());
```
jenmerritt:
```suggestion
    deleteAllGeometriesButton.disableProperty().bind(
        Bindings.or(geometryEditor.startedProperty(), Bindings.createBooleanBinding(() -> graphicsOverlay.getGraphics().isEmpty())));
```
In Qt they check if the graphics overlay is empty so we should do that here too.

geometry/create-and-edit-geometries/src/main/java/com/esri/samples/create_and_edit_geometries/CreateAndEditGeometriesController.java
```diff
@@ -0,0 +1,591 @@
...
+   * Starts the geometry editor with the point geometry type.
+   */
+  @FXML
+  private void handlePointButtonClicked() {
+    if (!geometryEditor.isStarted()) {
+      graphicsOverlay.clearSelection();
+      geometryEditor.start(GeometryType.POINT);
+
+      // disable tool selection when point or multipoint selected
+      toolComboBox.setDisable(true);
```
jenmerritt:
```suggestion
```
We're now binding this so can be removed

geometry/create-and-edit-geometries/src/main/java/com/esri/samples/create_and_edit_geometries/CreateAndEditGeometriesController.java
```diff
@@ -0,0 +1,591 @@
...
+  }
+
+  /**
+   * Starts the geometry editor with the point geometry type.
+   */
+  @FXML
+  private void handlePointButtonClicked() {
+    if (!geometryEditor.isStarted()) {
+      graphicsOverlay.clearSelection();
+      geometryEditor.start(GeometryType.POINT);
```
jenmerritt, line 283:
```suggestion
      geometryEditor.setTool(vertexTool);
      geometryEditor.start(GeometryType.POINT);
```
We should set the vertex tool when we use the point geometry type, follows Qt.

geometry/create-and-edit-geometries/src/main/java/com/esri/samples/create_and_edit_geometries/CreateAndEditGeometriesController.java
```diff
@@ -0,0 +1,591 @@
...
+   * Starts the geometry editor with the multipoint geometry type.
+   */
+  @FXML
+  private void handleMultipointButtonClicked() {
+    if (!geometryEditor.isStarted()) {
+      graphicsOverlay.clearSelection();
+      geometryEditor.start(GeometryType.MULTIPOINT);
+
+      // disable tool selection when point or multipoint selected
+      toolComboBox.setDisable(true);
```
jenmerritt:
```suggestion
```
We're now binding this so can be removed

geometry/create-and-edit-geometries/src/main/java/com/esri/samples/create_and_edit_geometries/CreateAndEditGeometriesController.java
```diff
@@ -0,0 +1,591 @@
...
+  }
+
+  /**
+   * Starts the geometry editor with the multipoint geometry type.
+   */
+  @FXML
+  private void handleMultipointButtonClicked() {
+    if (!geometryEditor.isStarted()) {
+      graphicsOverlay.clearSelection();
+      geometryEditor.start(GeometryType.MULTIPOINT);
```
jenmerritt, line 295:
```suggestion
      geometryEditor.setTool(vertexTool);
      geometryEditor.start(GeometryType.MULTIPOINT);
```
As per point

geometry/create-and-edit-geometries/src/main/java/com/esri/samples/create_and_edit_geometries/CreateAndEditGeometriesController.java
```diff
@@ -0,0 +1,591 @@
...
+  }
+
+  /**
+   * Saves the geometry in the editor to the graphics overlay and sets its symbol to the type relevant for the geometry.
+   */
+  @FXML
+  private void handleStopAndSaveButtonClicked() {
+    if (geometryEditor.getGeometry().isEmpty()) {
+      System.err.println("Geometry is empty. Unable to save geometry.");
+    }
```
jenmerritt:
```suggestion
```
This pattern isn't something we generally do in samples so I suggest adding an Alert as per below comment

geometry/create-and-edit-geometries/src/main/java/com/esri/samples/create_and_edit_geometries/CreateAndEditGeometriesController.java
```diff
@@ -0,0 +1,591 @@
...
+
+    // handle saving depending on if the geometry being edited is new, or already exists and needs updating
+    if (selectedGraphic != null) {
+      updateSelectedGraphic();
+    } else {
+      createNewGraphic();
+    }
```
jenmerritt, line 356:
```suggestion
    if (!geometryEditor.getGeometry().isEmpty()) {
      if (selectedGraphic != null) {
        updateSelectedGraphic();
      } else {
        createNewGraphic();
      }
    } else {
      new Alert(Alert.AlertType.ERROR, "Geometry is empty.\nUnable to save geometry.").show();
    }
```

geometry/create-and-edit-geometries/src/main/java/com/esri/samples/create_and_edit_geometries/CreateAndEditGeometriesController.java
```diff
@@ -0,0 +1,591 @@
...
+
+    // enable tool selection once a graphic is no longer being edited
+    toolComboBox.setDisable(false);
```
jenmerritt:
```suggestion
```
We're binding this now so can remove

geometry/create-and-edit-geometries/src/main/java/com/esri/samples/create_and_edit_geometries/CreateAndEditGeometriesController.java
```diff
@@ -0,0 +1,591 @@
...
+
+  /**
+   * Stops the geometry editor without saving the geometry stored within.
+   */
+  @FXML
+  private void handleStopAndDiscardButtonClicked() {
+    // if editing a pre-existing graphic, make it visible again
+    if (selectedGraphic != null) {
+      selectedGraphic.setVisible(true);
+      selectedGraphic = null;
```
jenmerritt, line 408:
```suggestion
      selectedGraphic.setSelected(false);
      selectedGraphic = null;
```
Also need to deselect as otherwise it stays selected on the map if you don't make any edits.

geometry/create-and-edit-geometries/src/main/java/com/esri/samples/create_and_edit_geometries/CreateAndEditGeometriesController.java
```diff
@@ -0,0 +1,591 @@
...
+  private void handleDeleteAllGeometriesButtonClicked() {
+    graphicsOverlay.getGraphics().clear();
+  }
+
+  /**
+   * Allows the user to select a graphic from the graphics overlay and edit its geometry.
+   */
+  private void selectGraphic() {
+
+    mapView.setOnMouseClicked(e -> {
```
jenmerritt:
```suggestion
    mapView.setOnMouseClicked(e -> {
      if (e.getButton() == MouseButton.PRIMARY && e.isStillSincePress()) {
```
Additional checks to avoid unwanted clicks e.g. currently if you drag over a graphic you select something.

geometry/create-and-edit-geometries/src/main/java/com/esri/samples/create_and_edit_geometries/CreateAndEditGeometriesController.java
```diff
@@ -0,0 +1,591 @@
...
+              selectedGraphic = identifyGraphics.get().getGraphics().get(0);
+              selectedGraphic.setSelected(true);
+
+              // Use vertex tool if selected geometry is a point or multipoint
+              if (selectedGraphic.getGeometry().getGeometryType().equals(GeometryType.POINT) ||
+                selectedGraphic.getGeometry().getGeometryType().equals(GeometryType.MULTIPOINT)) {
+                geometryEditor.setTool(vertexTool);
+                toolComboBox.getSelectionModel().select(0);
+                toolComboBox.setDisable(true);
+              }
```
jenmerritt:
```suggestion
```
Thanks to our bindings and setting the tool when we select point/multipoint, we can remove this block now.

geometry/create-and-edit-geometries/CreateAndEditGeometries.png
jenmerritt, line 1: I think we need to update the image after the changes and as per the common design - ie there are some different states on the button.

geometry/create-and-edit-geometries/src/main/resources/create_and_edit_geometries/polygon_data.json
```diff
@@ -0,0 +1,15 @@
...
+    [-9.59255263, 53.08256769], [-9.59265165, 53.08237906],
+    [-9.59287552, 53.08241478], [-9.59292812, 53.0823012],
+    [-9.5932294, 53.08235022], [-9.59342188, 53.08260009],
+    [-9.59354382, 53.08238728], [-9.59365852, 53.08203535],
+    [-9.59408443, 53.08210446], [-9.59448232, 53.08224456],
+    [-9.5943609, 53.08243697], [-9.59458319, 53.08245939],
+    [-9.59439639, 53.08264619], [-9.59433288, 53.0827975],
+    [-9.59404707, 53.08323649], [-9.59350122, 53.08320723]]],
+  "spatialReference":{"wkid":4326}
+}
```
jenmerritt:
```suggestion
}

```

geometry/create-and-edit-geometries/src/main/resources/create_and_edit_geometries/style.css
```diff
@@ -0,0 +1,37 @@
...
+   -fx-text-fill: white;
+}
+
+.panel-region .radio-button {
+   -fx-text-fill: white;
+}
+
+.color-picker .color-picker-label {
+   -fx-text-fill: black;
+}
```
jenmerritt: Let's remove any unused CSS

geometry/create-and-edit-geometries/src/main/java/com/esri/samples/create_and_edit_geometries/CreateAndEditGeometriesController.java
```diff
@@ -0,0 +1,440 @@
...
+public class CreateAndEditGeometriesController {
+
+  @FXML private MapView mapView;
+  @FXML private Button pointButton;
+  @FXML private Button multipointButton;
+  @FXML private Button polylineButton;
+  @FXML private Button polygonButton;
+  @FXML private Button redoButton;
+  @FXML private Button undoButton;
+  @FXML public Button deleteSelectedElementButton;
```
jenmerritt:
```suggestion
  @FXML private Button deleteSelectedElementButton;
```

geometry/create-and-edit-geometries/src/main/java/com/esri/samples/create_and_edit_geometries/CreateAndEditGeometriesController.java
```diff
@@ -0,0 +1,440 @@
...
+  @FXML private Button stopAndSaveButton;
+  @FXML private Button stopAndDiscardButton;
+  @FXML public Button deleteAllGeometriesButton;
```
jenmerritt:
```suggestion
  @FXML private Button deleteAllGeometriesButton;
```

geometry/create-and-edit-geometries/src/main/java/com/esri/samples/create_and_edit_geometries/CreateAndEditGeometriesController.java
```diff
@@ -0,0 +1,440 @@
...
+    // create a geometry editor and set it to the map view
+    geometryEditor = new GeometryEditor();
+    mapView.setGeometryEditor(geometryEditor);
+
+    // create vertex and freehand tools for the geometry editor and add to combo box - vertex tool selected by default
+    vertexTool = new VertexTool();
+    freehandTool = new FreehandTool();
+    toolComboBox.setConverter(new ComboBoxStringConverter());
+    toolComboBox.getItems().addAll(vertexTool, freehandTool);
+    toolComboBox.getSelectionModel().select(0);
```
jenmerritt:
```suggestion
    // create vertex and freehand tools for the geometry editor and add to combo box
    vertexTool = new VertexTool();
    freehandTool = new FreehandTool();
    toolComboBox.setConverter(new ComboBoxStringConverter());
    toolComboBox.getItems().addAll(vertexTool, freehandTool);
    // bidirectionally bind the geometry editor tool to the tool selected in the combo box
    toolComboBox.valueProperty().bindBidirectional(geometryEditor.toolProperty());
```
We can move that binding up here now and then removes the need for setting a default. The GeometryEditor is initialized with a VertexTool.

geometry/create-and-edit-geometries/src/main/java/com/esri/samples/create_and_edit_geometries/CreateAndEditGeometriesController.java
```diff
@@ -0,0 +1,440 @@
...
+      multipointGraphic.setSymbol(multipointSymbol);
+      polyline1Graphic.setSymbol(lineSymbol);
+      polyline2Graphic.setSymbol(lineSymbol);
+      polygonGraphic.setSymbol(fillSymbol);
+
+      // add example geometry graphics to the graphics overlay
+      graphicsOverlay.getGraphics().addAll(List.of(pointGraphic, multipointGraphic, polyline1Graphic, polyline2Graphic, polygonGraphic));
+    } catch (Exception e) {
+      new Alert(Alert.AlertType.ERROR, "Unable to load example geometries from file").show();
+    }
```
jenmerritt: This looks good. Just a couple of structural things:
- let's put the JSON files into a directory like you did for icons
- Make this a method like `createInitialGraphics()` - as in this case, it's less essential to learning how the Geometry Editor itself works so we can separate this out a bit more.

geometry/create-and-edit-geometries/src/main/java/com/esri/samples/create_and_edit_geometries/CreateAndEditGeometriesController.java
```diff
@@ -0,0 +1,440 @@
...
+        polygonButton.setDisable(false);
+
+        // disable combo box when geometry editor is not started
+        toolComboBox.disableProperty().unbind();
+        toolComboBox.setDisable(true);
+      }
+    });
+
+    // bidirectionally bind the geometry editor tool to the tool selected in the combo box
+    toolComboBox.valueProperty().bindBidirectional(geometryEditor.toolProperty());
```
jenmerritt: As per comment above, we can move this up now.

geometry/create-and-edit-geometries/src/main/java/com/esri/samples/create_and_edit_geometries/CreateAndEditGeometriesController.java
```diff
@@ -0,0 +1,440 @@
...
+
+    // bind the disable property of the stop and discard button to the geometry editor's started property
+    stopAndDiscardButton.disableProperty().bind(geometryEditor.startedProperty().not());
+
+    // bind the disable property of the delete all geometries button to the geometry editor's started property
+    // and to whether there are any graphics in the graphics overlay
+    deleteAllGeometriesButton.disableProperty().bind(Bindings.or(
+      geometryEditor.startedProperty(), Bindings.createBooleanBinding(() -> graphicsOverlay.getGraphics().isEmpty())));
+
+    // allow user to immediately select and edit graphics
```
jenmerritt:
```suggestion
    // configure identification and editing of existing graphics
```

geometry/create-and-edit-geometries/src/main/java/com/esri/samples/create_and_edit_geometries/CreateAndEditGeometriesController.java
```diff
@@ -0,0 +1,440 @@
...
+    mapView.setOnMouseClicked(e -> {
+      if (e.getButton() == MouseButton.PRIMARY && e.isStillSincePress()){
+      graphicsOverlay.clearSelection();
+      Point2D mapViewPoint = new Point2D(e.getX(), e.getY());
```
jenmerritt:
```suggestion
      if (e.getButton() == MouseButton.PRIMARY && e.isStillSincePress()) {
      graphicsOverlay.clearSelection();
      var mapViewPoint = new Point2D(e.getX(), e.getY());
```
Also check the indentation here, I think the if statement body needs indenting

geometry/create-and-edit-geometries/src/main/java/com/esri/samples/create_and_edit_geometries/CreateAndEditGeometriesController.java
```diff
@@ -0,0 +1,440 @@
...
+
+      // get graphics near the clicked location
+      ListenableFuture<IdentifyGraphicsOverlayResult> identifyGraphics;
+      identifyGraphics = mapView.identifyGraphicsOverlayAsync(graphicsOverlay, mapViewPoint, 10, false);
```
jenmerritt:
```suggestion
      ListenableFuture<IdentifyGraphicsOverlayResult> identifyGraphics =
       mapView.identifyGraphicsOverlayAsync(graphicsOverlay, mapViewPoint, 10, false);
```

geometry/create-and-edit-geometries/src/main/java/com/esri/samples/create_and_edit_geometries/CreateAndEditGeometriesController.java
```diff
@@ -0,0 +1,440 @@
...
+
+      identifyGraphics.addDoneListener(() -> {
+        try {
+          // only select graphic if user not already creating/editing a geometry
+          if (!geometryEditor.isStarted()) {
+            if (!identifyGraphics.get().getGraphics().isEmpty()) {
+
+              // store the selected graphic
```
jenmerritt:
```suggestion
              // store the selected graphic so that it can be retained if edits are discarded
```
I think we can make this explicit - as like we've found when implementing ourselves, it's easy to forget this and then ending up deleting everything.

geometry/create-and-edit-geometries/src/main/java/com/esri/samples/create_and_edit_geometries/CreateAndEditGeometriesController.java
```diff
@@ -0,0 +1,440 @@
...
+              // hide the selected graphic & start an editing session with a copy of it
+              geometryEditor.start(selectedGraphic.getGeometry());
+              selectedGraphic.setVisible(false);
+            } else {
+              selectedGraphic = null;
+            }
+          }
+        } catch (Exception x) {
+          // on any error, display the stack trace
+          x.printStackTrace();
```
jenmerritt:
```suggestion
        } catch (Exception ex) {
          // on any error, display the stack trace
          ex.printStackTrace();
```
Just better matches other samples

geometry/create-and-edit-geometries/src/main/java/com/esri/samples/create_and_edit_geometries/CreateAndEditGeometriesController.java
```diff
@@ -0,0 +1,440 @@
...
+  /**
+   * Removes all geometries from the graphics overlay.
+   */
+  @FXML
+  private void handleDeleteAllGeometriesButtonClicked() {
+    graphicsOverlay.getGraphics().clear();
+  }
+
+  /**
+   * Inner class for displaying friendly names in the combo box
```
jenmerritt:
```suggestion
   * Converts the GeometryEditorTool values to Strings to display in the ComboBox.
```

geometry/create-and-edit-geometries/src/main/resources/create_and_edit_geometries/main.fxml
```diff
@@ -0,0 +1,177 @@
...
+                            <Color opacity="0.3"/>
+                        </fill>
+                    </BackgroundFill>
+                </fills>
+            </Background>
+        </background>
+        <GridPane vgap="6" hgap="6" alignment="CENTER">
+            <Label GridPane.rowIndex="0" GridPane.columnIndex="0" text="Create"/>
+        </GridPane>
+        <GridPane vgap="6" hgap="6" alignment="CENTER">
```
jenmerritt: So we can actually put this whole UI into a single GridPane and use a property GridPane.columnSpan="2" where something needs to span right across. I'll send you a DM with a diagram. The main bit is to update the rowIndexes, and add the columSpan values where something needs to span across.

jenmerritt (commented):
Really coming on @01smito01 . Here's some more code comments and then I'll do a review of the comments/doc lastly. I think we'll also want to add some through the FXML file but let's get the code changes done first so it's easier to see the overall picture.

geometry/create-and-edit-geometries/src/main/java/com/esri/samples/create_and_edit_geometries/CreateAndEditGeometriesController.java
```diff
@@ -0,0 +1,451 @@
...
+    stopAndSaveButton.disableProperty().bind(Bindings.or(
+      geometryEditor.startedProperty().not(), geometryEditor.canUndoProperty().not()));
+
+    // bind the disable property of the stop and discard button to the geometry editor's started property
+    stopAndDiscardButton.disableProperty().bind(geometryEditor.startedProperty().not());
+
+    // bind the disable property of the delete all geometries button to the geometry editor's started property
+    // and to whether there are any graphics in the graphics overlay
+    deleteAllGeometriesButton.disableProperty().bind(Bindings.or(
+      geometryEditor.startedProperty(), Bindings.createBooleanBinding(() -> graphicsOverlay.getGraphics().isEmpty())));
```
jenmerritt: This binding doesn't seem to be working effectively, I can only assume it's something to do with the lack of property on the graphics. I experimented with the below which seems to work:
`deleteAllGeometriesButton.disableProperty().bind(Bindings.when(geometryEditor.startedProperty()).then(true).otherwise(graphicsOverlay.getGraphics().isEmpty()));`
See what you think.

geometry/create-and-edit-geometries/src/main/java/com/esri/samples/create_and_edit_geometries/CreateAndEditGeometriesController.java
```diff
@@ -0,0 +1,451 @@
...
+    // handle saving depending on if the geometry being edited is new, or already exists and needs updating
+    if (!geometryEditor.getGeometry().isEmpty()) {
+      if (selectedGraphic != null) {
+        updateSelectedGraphic();
+      } else {
+        createNewGraphic();
+      }
+    } else {
+      new Alert(Alert.AlertType.ERROR, "Geometry is empty. \nUnable to save geometry.").show();
+    }
```
jenmerritt:
```suggestion
      if (selectedGraphic != null) {
        updateSelectedGraphic();
      } else {
        createNewGraphic();
      }
```
Indentation will need adjusting here too.

This check is actually not good, as we need to be able to save when we delete a selected element. Currently this is a bug preventing us from doing so. It's following the implementation details, so I'll need to query it. For now though we should remove it.

geometry/create-and-edit-geometries/src/main/resources/create_and_edit_geometries/main.fxml
```diff
@@ -0,0 +1,167 @@
...
+<?import javafx.scene.layout.Background?>
+<?import javafx.scene.layout.BackgroundFill?>
+<?import javafx.scene.layout.GridPane?>
+<?import javafx.scene.layout.StackPane?>
+<?import javafx.scene.layout.VBox?>
+<?import javafx.scene.paint.Color?>
+<StackPane fx:controller="com.esri.samples.create_and_edit_geometries.CreateAndEditGeometriesController"
+           xmlns:fx="http://javafx.com/fxml"
+           stylesheets="@style.css">
+    <MapView fx:id="mapView"/>
```
jenmerritt, line 37:
```suggestion
    <!--MapView-->
    <MapView fx:id="mapView"/>
```

geometry/create-and-edit-geometries/src/main/resources/create_and_edit_geometries/main.fxml
```diff
@@ -0,0 +1,167 @@
...
+    <VBox StackPane.alignment="TOP_RIGHT" maxWidth="120" maxHeight="110" styleClass="panel-region"
+          alignment="CENTER" spacing="6">
```
jenmerritt:
```suggestion
    <VBox StackPane.alignment="TOP_RIGHT" maxWidth="120" maxHeight="110" styleClass="panel-region"
          alignment="CENTER" spacing="6">
```
We don't need a VBox now we only have 1 Gridpane

geometry/create-and-edit-geometries/src/main/resources/create_and_edit_geometries/main.fxml
```diff
@@ -0,0 +1,167 @@
...
+                <fills>
+                    <BackgroundFill>
+                        <fill>
+                            <Color opacity="0.3"/>
+                        </fill>
+                    </BackgroundFill>
+                </fills>
+            </Background>
+        </background>
+        <GridPane vgap="6" hgap="6" alignment="CENTER">
```
jenmerritt:
```suggestion
        <GridPane vgap="6" hgap="6" alignment="CENTER" StackPane.alignment="TOP_RIGHT" maxWidth="120" maxHeight="110" styleClass="panel-region">
```

jenmerritt:
```suggestion
        <GridPane vgap="6" hgap="6" alignment="CENTER">
                <padding>
            <Insets topRightBottomLeft="10"/>
        </padding>
        <background>
            <Background>
                <fills>
                    <BackgroundFill>
                        <fill>
                            <Color opacity="0.3"/>
                        </fill>
                    </BackgroundFill>
                </fills>
            </Background>
        </background>
```

jenmerritt: Move the padding/background into the gridpane to keep the styling

geometry/create-and-edit-geometries/src/main/resources/create_and_edit_geometries/main.fxml
```diff
@@ -0,0 +1,167 @@
...
+                    <ImageView>
+                        <Image url="@icons/trash-32.png" requestedHeight="24" preserveRatio="true" smooth="true"/>
+                    </ImageView>
+                </graphic>
+                <tooltip>
+                    <Tooltip text="Delete all geometries"/>
+                </tooltip>
+            </Button>
+        </GridPane>
+    </VBox>
```
jenmerritt:
```suggestion
```

geometry/create-and-edit-geometries/src/main/resources/create_and_edit_geometries/main.fxml
jenmerritt:
```suggestion
        <!--GridPane to contain GeometryEditor UI controls-->
        <GridPane vgap="6" hgap="6" alignment="CENTER">
```

geometry/create-and-edit-geometries/src/main/resources/create_and_edit_geometries/main.fxml
```diff
@@ -0,0 +1,167 @@
...
+                        <fill>
+                            <Color opacity="0.3"/>
+                        </fill>
+                    </BackgroundFill>
+                </fills>
+            </Background>
+        </background>
+        <GridPane vgap="6" hgap="6" alignment="CENTER">
+            <Label GridPane.rowIndex="0" GridPane.columnSpan="2" GridPane.halignment="CENTER" text="Create"/>
+            <Button fx:id="pointButton" onAction="#handlePointButtonClicked" GridPane.rowIndex="1" GridPane.columnIndex="0" text="Point">
```
jenmerritt:
```suggestion
            <!--Buttons configured with actions and graphics-->
            <Button fx:id="pointButton" onAction="#handlePointButtonClicked" GridPane.rowIndex="1" GridPane.columnIndex="0" text="Point">
```

geometry/create-and-edit-geometries/src/main/java/com/esri/samples/create_and_edit_geometries/CreateAndEditGeometriesController.java
```diff
@@ -0,0 +1,451 @@
...
+      selectedGraphic.setSelected(false);
+      selectedGraphic = null;
+      toolComboBox.disableProperty().unbind();
+      toolComboBox.setDisable(false);
+    }
+    geometryEditor.stop();
+  }
+
+  /**
+   * Removes all geometries from the graphics overlay.
```
jenmerritt:
```suggestion
   * Removes all graphics from the graphics overlay.
```

geometry/create-and-edit-geometries/src/main/java/com/esri/samples/create_and_edit_geometries/CreateAndEditGeometriesController.java
```diff
@@ -0,0 +1,451 @@
...
+
+  /**
+   * Updates a graphic already in the graphics overlay.
+   */
+  private void updateSelectedGraphic() {
+    // update the geometry of the graphic being edited and make it visible again
+    selectedGraphic.setGeometry(geometryEditor.stop());
+    selectedGraphic.setVisible(true);
+    selectedGraphic.setSelected(false);
+    selectedGraphic = null;
```
jenmerritt, line 396:
```suggestion
    // reset the selected graphic
    selectedGraphic.setSelected(false);
    selectedGraphic = null;
```

geometry/create-and-edit-geometries/src/main/java/com/esri/samples/create_and_edit_geometries/CreateAndEditGeometriesController.java
```diff
@@ -0,0 +1,451 @@
...
+        graphic.setSymbol(fillSymbol);
+        break;
+    }
+
+    // add new graphic to the graphics overlay
+    graphicsOverlay.getGraphics().add(graphic);
+  }
+
+  /**
+   * Updates a graphic already in the graphics overlay.
```
jenmerritt:
```suggestion
   * Updates and then resets the selected graphic.
```

geometry/create-and-edit-geometries/src/main/java/com/esri/samples/create_and_edit_geometries/CreateAndEditGeometriesController.java
```diff
@@ -0,0 +1,451 @@
...
+      } else {
+        createNewGraphic();
+      }
+    } else {
+      new Alert(Alert.AlertType.ERROR, "Geometry is empty. \nUnable to save geometry.").show();
+    }
+  }
+
+  /**
+   * Adds a new graphic to the graphics overlay.
```
jenmerritt:
```suggestion
   * Creates a new graphic based on the geometry type used by the geometry editor, and adds to the graphics overlay.
```

geometry/create-and-edit-geometries/src/main/java/com/esri/samples/create_and_edit_geometries/CreateAndEditGeometriesController.java
```diff
@@ -0,0 +1,451 @@
...
+  /**
+   * Deletes the currently selected element of the geometry.
+   */
+  @FXML
+  private void handleDeleteSelectedElementButtonClicked() {
+    geometryEditor.deleteSelectedElement();
+  }
+
+  /**
+   * Saves the geometry in the editor to the graphics overlay and sets its symbol to the type relevant for the geometry.
```
jenmerritt:
```suggestion
   * Checks whether an existing graphic is being edited and either updates it, or creates a new graphic based on the geometry in the geometry editor.
```

geometry/create-and-edit-geometries/src/main/java/com/esri/samples/create_and_edit_geometries/CreateAndEditGeometriesController.java
```diff
@@ -0,0 +1,451 @@
...
+  /**
+   * Redo the last change made to the geometry while editing is active.
+   */
+  @FXML
+  private void handleRedoButtonClicked() {
+    geometryEditor.redo();
+  }
+
+  /**
+   * Deletes the currently selected element of the geometry.
```
jenmerritt:
```suggestion
   * Deletes the currently selected element of the geometry editor.
```

geometry/create-and-edit-geometries/src/main/java/com/esri/samples/create_and_edit_geometries/CreateAndEditGeometriesController.java
```diff
@@ -0,0 +1,451 @@
...
+          try {
+            // only select graphic if user not already creating/editing a geometry
+            if (!geometryEditor.isStarted()) {
+              if (!identifyGraphics.get().getGraphics().isEmpty()) {
+
+                // store the selected graphic so that it can be retained if edits are discarded
+                selectedGraphic = identifyGraphics.get().getGraphics().get(0);
+                selectedGraphic.setSelected(true);
+
+                // hide the selected graphic & start an editing session with a copy of it
```
jenmerritt:
```suggestion
                // hide the selected graphic and start an editing session with a copy of it
```

geometry/create-and-edit-geometries/src/main/java/com/esri/samples/create_and_edit_geometries/CreateAndEditGeometriesController.java
```diff
@@ -0,0 +1,451 @@
...
+        polygonButton.disableProperty().unbind();
+        polygonButton.setDisable(false);
+
+        // disable combo box when geometry editor is not started
+        toolComboBox.disableProperty().unbind();
+        toolComboBox.setDisable(true);
+      }
+    });
+
+    // bind the disable property of the undo & redo buttons to the geometry editor's can undo & can redo properties
```
jenmerritt:
```suggestion
    // bind the disable property of the undo and redo buttons to the geometry editor's can undo and can redo properties
```

geometry/create-and-edit-geometries/src/main/java/com/esri/samples/create_and_edit_geometries/CreateAndEditGeometriesController.java
```diff
@@ -0,0 +1,451 @@
...
+    // thin blue line for polylines
+    lineSymbol = new SimpleLineSymbol(SimpleLineSymbol.Style.SOLID, Color.BLUE, 2);
+    // black outline for polygons
+    SimpleLineSymbol polygonLineSymbol = new SimpleLineSymbol(SimpleLineSymbol.Style.DASH, Color.BLACK, 1);
+    // translucent red interior for polygons
+    fillSymbol = new SimpleFillSymbol(SimpleFillSymbol.Style.SOLID, Color.rgb(255, 0, 0, 0.3), polygonLineSymbol);
+
+    createInitialGraphics();
+
+    geometryEditor.startedProperty().addListener((observable, oldValue, newValue) -> {
```
jenmerritt, line 128:
```suggestion
    // listen to changes in the geometry editor started property and update the UI
    geometryEditor.startedProperty().addListener((observable, oldValue, newValue) -> {
```

geometry/create-and-edit-geometries/src/main/java/com/esri/samples/create_and_edit_geometries/CreateAndEditGeometriesController.java
```diff
@@ -0,0 +1,451 @@
...
+    // yellow circle for multipoints
+    multipointSymbol = new SimpleMarkerSymbol(SimpleMarkerSymbol.Style.CIRCLE, Color.YELLOW, 5);
+    // thin blue line for polylines
+    lineSymbol = new SimpleLineSymbol(SimpleLineSymbol.Style.SOLID, Color.BLUE, 2);
+    // black outline for polygons
+    SimpleLineSymbol polygonLineSymbol = new SimpleLineSymbol(SimpleLineSymbol.Style.DASH, Color.BLACK, 1);
+    // translucent red interior for polygons
+    fillSymbol = new SimpleFillSymbol(SimpleFillSymbol.Style.SOLID, Color.rgb(255, 0, 0, 0.3), polygonLineSymbol);
+
+    createInitialGraphics();
```
jenmerritt, line 125:
```suggestion
    // add some initial, example graphics to the graphics overlay
    createInitialGraphics();
```

geometry/create-and-edit-geometries/src/main/java/com/esri/samples/create_and_edit_geometries/CreateAndEditGeometriesController.java
```diff
@@ -0,0 +1,451 @@
...
+    // create a geometry editor and set it to the map view
+    geometryEditor = new GeometryEditor();
+    mapView.setGeometryEditor(geometryEditor);
+
+    // create vertex and freehand tools for the geometry editor and add to combo box
+    vertexTool = new VertexTool();
+    freehandTool = new FreehandTool();
+    toolComboBox.setConverter(new ComboBoxStringConverter());
+    toolComboBox.getItems().addAll(vertexTool, freehandTool);
+
```
jenmerritt:
```suggestion
```

geometry/create-and-edit-geometries/src/main/java/com/esri/samples/create_and_edit_geometries/CreateAndEditGeometriesController.java
```diff
@@ -0,0 +1,451 @@
...
+    mapView.setMap(map);
+
+    // set a viewpoint on the map view
+    mapView.setViewpoint(new Viewpoint(53.08230, -9.5920, 5000));
+
+    // create a graphics overlay for the graphics
+    graphicsOverlay = new GraphicsOverlay();
+
+    // add the graphics overlay to the map view
+    mapView.getGraphicsOverlays().add(graphicsOverlay);
```
jenmerritt:
```suggestion
    // create a graphics overlay and add it to the map view
    graphicsOverlay = new GraphicsOverlay();
    mapView.getGraphicsOverlays().add(graphicsOverlay);
```

jenmerritt (commented):
Thanks @01smito01 . There's a couple of bugs from the changes so a couple of suggestions for fixing those. And then I've been through the comments too. Worth a proof read though..

geometry/create-and-edit-geometries/src/main/java/com/esri/samples/create_and_edit_geometries/CreateAndEditGeometriesController.java
01smito01: These both seem to perform the same to me - disabled when the geometry editor is active, enabled when it isn't (regardless of whether there's any graphics in the graphics overlay). What difference in functionality are you seeing?

jenmerritt: When I open the app it's disabled

jenmerritt: Ah - maybe create the initial graphics last thing after you've setup the bindings? Although it should detect them being added. Not sure, it just didn't look correct to me as I thought I couldn't initially clear all the graphics?

jenmerritt: Yeah I can't replicate this again - must have been too early in the day!

jenmerritt (approved):
One more fix - make sure you correct the indentation in the FXML. Key is to try get in the habit of checking imports and indents after refactoring.

But I'm happy for this to go to second review. Post in Slack to say it's ready.

Luiy0:
Hi @01smito01, I'll soon be doing the second review.

geometry/create-and-edit-geometries/src/main/java/com/esri/samples/create_and_edit_geometries/CreateAndEditGeometriesController.java
```diff
@@ -0,0 +1,446 @@
...
+        identifyGraphics.addDoneListener(() -> {
+          try {
+            // only select graphic if user not already creating/editing a geometry
+            if (!geometryEditor.isStarted()) {
+              if (!identifyGraphics.get().getGraphics().isEmpty()) {
+
+                // store the selected graphic so that it can be retained if edits are discarded
+                selectedGraphic = identifyGraphics.get().getGraphics().get(0);
+                selectedGraphic.setSelected(true);
+
```
Luiy0, line 208: Consider re-adding the following to update the tool according to the selected geometry type.

```
// Use vertex tool if selected geometry is a point or multipoint
if (selectedGraphic.getGeometry().getGeometryType().equals(GeometryType.POINT) ||
  selectedGraphic.getGeometry().getGeometryType().equals(GeometryType.MULTIPOINT)) {
  geometryEditor.setTool(vertexTool);
}
```

Luiy0 (commented):
Sample looks good @01smito01! Only one comment on setting the vertex/freehand tool when editing. As we discussed, the code below does the trick. Perhaps there could be an alternate way of achieving the same with bindings @jenmerritt?

Luiy0 (approved):
Looks good @01smito01

PR #788, opened 2023-05-22
CompletableFuture in samples

01smito01:
### Description

Introduced `CompletableFuture` to six samples, which support some of the more interesting methods that `CompletableFuture` provides:
- Edit Feature Attachments
- Generate Offline Map Overrides
- Mobile Map Search and Route
- Perform Valve Isolation Trace
- Read Symbols from Mobile Style File
- Symbol Dictionary

Other than the `build.gradle`s being changed to a daily build, main thing to check is that everything still works and I haven't lost any error messages in the process of converting.

@jenmerritt don't think there's any particular rush for this; I'm sure you can put me right in terms of build versions. 

Checklist:
- [x] Request a reviewer
- [x] Assign a reviewer
- [x] Tag reviewer in comment

### DIFF
--- a/editing/edit-feature-attachments/src/main/java/com/esri/samples/edit_feature_attachments/EditFeatureAttachmentsSample.java
+++ b/editing/edit-feature-attachments/src/main/java/com/esri/samples/edit_feature_attachments/EditFeatureAttachmentsSample.java
@@ -65,8 +65,8 @@ public class EditFeatureAttachmentsSample extends Application {
 
     try {
       // create stack pane and application scene
-      StackPane stackPane = new StackPane();
-      Scene scene = new Scene(stackPane);
+      var stackPane = new StackPane();
+      var scene = new Scene(stackPane);
       scene.getStylesheets().add(getClass().getResource("/edit_feature_attachments/style.css").toExternalForm());
 
       // set title, size, and add scene to stage
@@ -132,7 +132,7 @@ public class EditFeatureAttachmentsSample extends Application {
       featureTable = new ServiceFeatureTable("https://sampleserver6.arcgisonline.com/arcgis/rest/services/DamageAssessment/FeatureServer/0");
 
       // create a feature layer from service feature table
-      FeatureLayer featureLayer = new FeatureLayer(featureTable);
+      var featureLayer = new FeatureLayer(featureTable);
 
       // add the feature layer to the ArcGISMap
       map.getOperationalLayers().add(featureLayer);
@@ -158,6 +158,7 @@ public class EditFeatureAttachmentsSample extends Application {
           mapView.identifyLayerAsync(featureLayer, point, 1, false).toCompletableFuture()
             .whenComplete((layer, ex) -> {
               if (ex == null) {
+                // if the identification operation completes successfully, get the list of identified GeoElements
                 List<GeoElement> identified = layer.getElements();
                 if (identified.size() > 0) {
                   GeoElement element = identified.get(0);
@@ -177,6 +178,7 @@ public class EditFeatureAttachmentsSample extends Application {
                   }
                 }
               } else {
+                // if the identification operation completes with an exception, display an error
                 displayMessage("Exception getting identify result", ex.getCause().getMessage());
               }
             });
@@ -200,9 +202,9 @@ public class EditFeatureAttachmentsSample extends Application {
     feature.fetchAttachmentsAsync().toCompletableFuture().whenComplete(
       (attachmentResults, ex) -> {
         if (ex == null) {
+          // if the attachments were fetched successfully, update the UI attachments list
           attachments = attachmentResults;
 
-          // update UI attachments list
           Platform.runLater(() -> {
             attachmentList.getItems().clear();
             attachments.forEach(attachment -> attachmentList.getItems().add(attachment.getName()));
@@ -273,7 +275,7 @@ public class EditFeatureAttachmentsSample extends Application {
           }
           // update the displayed list of attachments
           fetchAttachments(selected);
-        } else  {
+        } else {
           displayMessage("Error applying edits on server ", ex.getCause().getMessage());
         }
       });
--- a/editing/edit-feature-attachments/src/main/java/com/esri/samples/edit_feature_attachments/EditFeatureAttachmentsSample.java
+++ b/editing/edit-feature-attachments/src/main/java/com/esri/samples/edit_feature_attachments/EditFeatureAttachmentsSample.java
@@ -230,6 +230,7 @@ public class EditFeatureAttachmentsSample extends Application {
   private void addAttachment(byte[] attachment) {
 
     if (selected.canEditAttachments()) {
+      // update feature table and apply update to server when new feature is added
       selected.addAttachmentAsync(attachment, "image/png", "edit_feature_attachments/destroyed.png")
         .toCompletableFuture()
         .thenCompose(addedAttachment -> featureTable.updateFeatureAsync(selected).toCompletableFuture())
--- a/editing/edit-feature-attachments/src/main/java/com/esri/samples/edit_feature_attachments/EditFeatureAttachmentsSample.java
+++ b/editing/edit-feature-attachments/src/main/java/com/esri/samples/edit_feature_attachments/EditFeatureAttachmentsSample.java
@@ -56,7 +56,6 @@ public class EditFeatureAttachmentsSample extends Application {
 
   private ListView<String> attachmentList;
   private Label attachmentsLabel;
-
   private ArcGISFeature selected;
   private List<Attachment> attachments;
   private ServiceFeatureTable featureTable;
@@ -232,7 +231,7 @@ public class EditFeatureAttachmentsSample extends Application {
         .toCompletableFuture();
 
       // update feature table, apply update to server when new feature is added, and update the displayed list of attachments
-      addResult.thenComposeAsync(addedAttachment -> featureTable.updateFeatureAsync(selected).toCompletableFuture())
+      addResult.thenCompose(addedAttachment -> featureTable.updateFeatureAsync(selected).toCompletableFuture())
         .thenRun(() -> applyEdits(featureTable));
     } else {
       displayMessage(null, "Cannot add attachment.");
--- a/editing/edit-feature-attachments/src/main/java/com/esri/samples/edit_feature_attachments/EditFeatureAttachmentsSample.java
+++ b/editing/edit-feature-attachments/src/main/java/com/esri/samples/edit_feature_attachments/EditFeatureAttachmentsSample.java
@@ -264,13 +264,16 @@ public class EditFeatureAttachmentsSample extends Application {
     // apply the changes to the server
     featureTable.applyEditsAsync().toCompletableFuture().whenComplete(
       (editResults, exception) -> {
+        // check if the server edit was successful
         if (exception == null) {
-          // check if the server edit was successful
           if (!editResults.isEmpty()) {
+            // check for any errors. In this case we are only updating 1 feature so just check the first result
             if (!editResults.get(0).hasCompletedWithErrors()) {
-              displayMessage(null, "Edited feature successfully");
+              // if there are no errors, display a success message
+              displayMessage("Success", "Edited feature successfully");
             } else {
               if (editResults.get(0).getError() != null) {
+                // if there is an error, throw
                 throw editResults.get(0).getError();
               }
             }
--- a/editing/edit-feature-attachments/src/main/java/com/esri/samples/edit_feature_attachments/EditFeatureAttachmentsSample.java
+++ b/editing/edit-feature-attachments/src/main/java/com/esri/samples/edit_feature_attachments/EditFeatureAttachmentsSample.java
@@ -18,7 +18,7 @@ package com.esri.samples.edit_feature_attachments;
 
 import org.apache.commons.io.IOUtils;
 import java.util.List;
-import java.util.concurrent.ExecutionException;
+import java.util.concurrent.CompletableFuture;
 
 import javafx.application.Application;
 import javafx.application.Platform;
@@ -41,10 +41,8 @@ import javafx.scene.paint.Paint;
 import javafx.stage.Stage;
 
 import com.esri.arcgisruntime.ArcGISRuntimeEnvironment;
-import com.esri.arcgisruntime.concurrent.ListenableFuture;
 import com.esri.arcgisruntime.data.ArcGISFeature;
 import com.esri.arcgisruntime.data.Attachment;
-import com.esri.arcgisruntime.data.FeatureEditResult;
 import com.esri.arcgisruntime.data.ServiceFeatureTable;
 import com.esri.arcgisruntime.layers.FeatureLayer;
 import com.esri.arcgisruntime.loadable.LoadStatus;
@@ -52,7 +50,6 @@ import com.esri.arcgisruntime.mapping.ArcGISMap;
 import com.esri.arcgisruntime.mapping.BasemapStyle;
 import com.esri.arcgisruntime.mapping.GeoElement;
 import com.esri.arcgisruntime.mapping.Viewpoint;
-import com.esri.arcgisruntime.mapping.view.IdentifyLayerResult;
 import com.esri.arcgisruntime.mapping.view.MapView;
 
 public class EditFeatureAttachmentsSample extends Application {
@@ -88,7 +85,7 @@ public class EditFeatureAttachmentsSample extends Application {
       // create a control panel
       VBox controlsVBox = new VBox(6);
       controlsVBox.setBackground(new Background(new BackgroundFill(Paint.valueOf("rgba(0,0,0,0.3)"), CornerRadii.EMPTY,
-              Insets.EMPTY)));
+        Insets.EMPTY)));
       controlsVBox.setPadding(new Insets(10.0));
       controlsVBox.setMaxSize(180, 250);
       controlsVBox.getStyleClass().add("panel-region");
@@ -160,32 +157,31 @@ public class EditFeatureAttachmentsSample extends Application {
           attachmentList.getItems().clear();
 
           // get the clicked feature
-          ListenableFuture<IdentifyLayerResult> results = mapView.identifyLayerAsync(featureLayer, point, 1, false);
-          results.addDoneListener(() -> {
-            try {
-              IdentifyLayerResult layer = results.get();
-              List<GeoElement> identified = layer.getElements();
-              if (identified.size() > 0) {
-                GeoElement element = identified.get(0);
-                // get selected feature
-                if (element instanceof ArcGISFeature) {
-                  selected = (ArcGISFeature) element;
-                  featureLayer.selectFeature(selected);
-                  selected.loadAsync();
-                  selected.addDoneLoadingListener(() -> {
-                    if (selected.getLoadStatus() == LoadStatus.LOADED) {
-                      fetchAttachments(selected);
-                    } else {
-                      displayMessage("Error", "Element failed to load!");
-                    }
-                  });
-                  addAttachmentButton.setDisable(false);
+          mapView.identifyLayerAsync(featureLayer, point, 1, false).toCompletableFuture()
+            .whenComplete((layer, ex) -> {
+              if (ex == null) {
+                List<GeoElement> identified = layer.getElements();
+                if (identified.size() > 0) {
+                  GeoElement element = identified.get(0);
+                  // get selected feature
+                  if (element instanceof ArcGISFeature) {
+                    selected = (ArcGISFeature) element;
+                    featureLayer.selectFeature(selected);
+                    selected.loadAsync();
+                    selected.addDoneLoadingListener(() -> {
+                      if (selected.getLoadStatus() == LoadStatus.LOADED) {
+                        fetchAttachments(selected);
+                      } else {
+                        displayMessage("Error", "Element failed to load!");
+                      }
+                    });
+                    addAttachmentButton.setDisable(false);
+                  }
                 }
+              } else {
+                displayMessage("Exception getting identify result", ex.getCause().getMessage());
               }
-            } catch (InterruptedException | ExecutionException e) {
-              displayMessage("Exception getting identify result", e.getCause().getMessage());
-            }
-          });
+            });
         }
       });
 
@@ -203,25 +199,25 @@ public class EditFeatureAttachmentsSample extends Application {
    */
   private void fetchAttachments(ArcGISFeature feature) {
 
-    ListenableFuture<List<Attachment>> attachmentResults = feature.fetchAttachmentsAsync();
-    attachmentResults.addDoneListener(() -> {
-      try {
-        attachments = attachmentResults.get();
+    feature.fetchAttachmentsAsync().toCompletableFuture().whenComplete(
+      (attachmentResults, ex) -> {
+        if (ex == null) {
+          attachments = attachmentResults;
 
-        // update UI attachments list
-        Platform.runLater(() -> {
-          attachmentList.getItems().clear();
-          attachments.forEach(attachment -> attachmentList.getItems().add(attachment.getName()));
-          if (!attachments.isEmpty()) {
-            attachmentsLabel.setText("Attachments: ");
-          } else {
-            attachmentsLabel.setText("No Attachments!");
-          }
-        });
-      } catch (InterruptedException | ExecutionException e) {
-        displayMessage("Exception getting feature attachments", e.getCause().getMessage());
-      }
-    });
+          // update UI attachments list
+          Platform.runLater(() -> {
+            attachmentList.getItems().clear();
+            attachments.forEach(attachment -> attachmentList.getItems().add(attachment.getName()));
+            if (!attachments.isEmpty()) {
+              attachmentsLabel.setText("Attachments: ");
+            } else {
+              attachmentsLabel.setText("No Attachments!");
+            }
+          });
+        } else {
+          displayMessage("Exception getting feature attachments", ex.getCause().getMessage());
+        }
+      });
   }
 
   /**
@@ -232,16 +228,12 @@ public class EditFeatureAttachmentsSample extends Application {
   private void addAttachment(byte[] attachment) {
 
     if (selected.canEditAttachments()) {
-      ListenableFuture<Attachment> addResult = selected.addAttachmentAsync(attachment, "image/png",
-              "edit_feature_attachments/destroyed.png");
-      addResult.addDoneListener(() -> {
-        // update feature table
-        ListenableFuture<Void> tableResult = featureTable.updateFeatureAsync(selected);
+      CompletableFuture<Attachment> addResult = selected.addAttachmentAsync(attachment, "image/png", "edit_feature_attachments/destroyed.png")
+        .toCompletableFuture();
 
-        // apply update to server when new feature is added, and update the
-        // displayed list of attachments
-        tableResult.addDoneListener(() -> applyEdits(featureTable));
-      });
+      // update feature table, apply update to server when new feature is added, and update the displayed list of attachments
+      addResult.thenComposeAsync(addedAttachment -> featureTable.updateFeatureAsync(selected).toCompletableFuture())
+        .thenRun(() -> applyEdits(featureTable));
     } else {
       displayMessage(null, "Cannot add attachment.");
     }
@@ -253,13 +245,10 @@ public class EditFeatureAttachmentsSample extends Application {
   private void deleteAttachment(int attachmentIndex) {
 
     if (selected.canEditAttachments()) {
-      ListenableFuture<Void> deleteResult = selected.deleteAttachmentAsync(attachments.get(attachmentIndex));
-      deleteResult.addDoneListener(() -> {
-        // update feature table
-        ListenableFuture<Void> tableResult = featureTable.updateFeatureAsync(selected);
-        // apply update to server when new feature is deleted
-        tableResult.addDoneListener(() -> applyEdits(featureTable));
-      });
+      // update feature table and apply update to server when new feature is deleted
+      selected.deleteAttachmentAsync(attachments.get(attachmentIndex)).toCompletableFuture()
+        .thenCompose(unused -> featureTable.updateFeatureAsync(selected).toCompletableFuture())
+        .thenRun(() -> applyEdits(featureTable));
     } else {
       displayMessage(null, "Cannot delete attachment");
     }
@@ -273,26 +262,25 @@ public class EditFeatureAttachmentsSample extends Application {
   private void applyEdits(ServiceFeatureTable featureTable) {
 
     // apply the changes to the server
-    ListenableFuture<List<FeatureEditResult>> editResult = featureTable.applyEditsAsync();
-    editResult.addDoneListener(() -> {
-      try {
-        List<FeatureEditResult> edits = editResult.get();
-        // check if the server edit was successful
-        if (edits != null && edits.size() > 0) {
-          if (!edits.get(0).hasCompletedWithErrors()) {
-            displayMessage(null, "Edited feature successfully");
-          } else {
-            if (edits.get(0).getError() != null) {
-              throw edits.get(0).getError();
+    featureTable.applyEditsAsync().toCompletableFuture().whenComplete(
+      (edits, ex) -> {
+        if (ex == null) {
+          // check if the server edit was successful
+          if (edits != null && edits.size() > 0) {
+            if (!edits.get(0).hasCompletedWithErrors()) {
+              displayMessage(null, "Edited feature successfully");
+            } else {
+              if (edits.get(0).getError() != null) {
+                throw edits.get(0).getError();
+              }
             }
           }
+          // update the displayed list of attachments
+          fetchAttachments(selected);
+        } else  {
+          displayMessage("Error applying edits on server ", ex.getCause().getMessage());
         }
-        // update the displayed list of attachments
-        fetchAttachments(selected);
-      } catch (InterruptedException | ExecutionException e) {
-        displayMessage("Error applying edits on server ", e.getCause().getMessage());
-      }
-    });
+      });
   }
 
   /**
--- a/editing/edit-feature-attachments/src/main/java/com/esri/samples/edit_feature_attachments/EditFeatureAttachmentsSample.java
+++ b/editing/edit-feature-attachments/src/main/java/com/esri/samples/edit_feature_attachments/EditFeatureAttachmentsSample.java
@@ -155,12 +155,12 @@ public class EditFeatureAttachmentsSample extends Application {
           attachmentList.getItems().clear();
 
           // get the clicked feature
-          mapView.identifyLayerAsync(featureLayer, point, 1, false).toCompletableFuture()
-            .whenComplete((layer, ex) -> {
-              if (ex == null) {
+          mapView.identifyLayerAsync(featureLayer, point, 1, false, 1).toCompletableFuture()
+            .whenComplete((identifyLayerResult, exception) -> {
+              if (exception == null) {
                 // if the identification operation completes successfully, get the list of identified GeoElements
-                List<GeoElement> identified = layer.getElements();
-                if (identified.size() > 0) {
+                List<GeoElement> identified = identifyLayerResult.getElements();
+                if (!identified.isEmpty()) {
                   GeoElement element = identified.get(0);
                   // get selected feature
                   if (element instanceof ArcGISFeature) {
@@ -169,7 +169,7 @@ public class EditFeatureAttachmentsSample extends Application {
                     selected.loadAsync();
                     selected.addDoneLoadingListener(() -> {
                       if (selected.getLoadStatus() == LoadStatus.LOADED) {
-                        fetchAttachments(selected);
+                        fetchAttachmentsAsync(selected);
                       } else {
                         displayMessage("Error", "Element failed to load!");
                       }
@@ -179,7 +179,7 @@ public class EditFeatureAttachmentsSample extends Application {
                 }
               } else {
                 // if the identification operation completes with an exception, display an error
-                displayMessage("Exception getting identify result", ex.getCause().getMessage());
+                displayMessage("Exception getting identify result", exception.getCause().getMessage());
               }
             });
         }
@@ -197,25 +197,25 @@ public class EditFeatureAttachmentsSample extends Application {
   /**
    * Updates the UI with a list of a feature's attachments.
    */
-  private void fetchAttachments(ArcGISFeature feature) {
+  private void fetchAttachmentsAsync(ArcGISFeature feature) {
 
     feature.fetchAttachmentsAsync().toCompletableFuture().whenComplete(
-      (attachmentResults, ex) -> {
-        if (ex == null) {
+      (attachmentResults, exception) -> {
+        if (exception == null) {
           // if the attachments were fetched successfully, update the UI attachments list
           attachments = attachmentResults;
 
           Platform.runLater(() -> {
             attachmentList.getItems().clear();
-            attachments.forEach(attachment -> attachmentList.getItems().add(attachment.getName()));
             if (!attachments.isEmpty()) {
               attachmentsLabel.setText("Attachments: ");
+              attachments.forEach(attachment -> attachmentList.getItems().add(attachment.getName()));
             } else {
               attachmentsLabel.setText("No Attachments!");
             }
           });
         } else {
-          displayMessage("Exception getting feature attachments", ex.getCause().getMessage());
+          displayMessage("Exception getting feature attachments", exception.getCause().getMessage());
         }
       });
   }
@@ -261,22 +261,22 @@ public class EditFeatureAttachmentsSample extends Application {
 
     // apply the changes to the server
     featureTable.applyEditsAsync().toCompletableFuture().whenComplete(
-      (edits, ex) -> {
-        if (ex == null) {
+      (editResults, exception) -> {
+        if (exception == null) {
           // check if the server edit was successful
-          if (edits != null && edits.size() > 0) {
-            if (!edits.get(0).hasCompletedWithErrors()) {
+          if (!editResults.isEmpty()) {
+            if (!editResults.get(0).hasCompletedWithErrors()) {
               displayMessage(null, "Edited feature successfully");
             } else {
-              if (edits.get(0).getError() != null) {
-                throw edits.get(0).getError();
+              if (editResults.get(0).getError() != null) {
+                throw editResults.get(0).getError();
               }
             }
           }
           // update the displayed list of attachments
-          fetchAttachments(selected);
+          fetchAttachmentsAsync(selected);
         } else {
-          displayMessage("Error applying edits on server ", ex.getCause().getMessage());
+          displayMessage("Error applying edits on server ", exception.getCause().getMessage());
         }
       });
   }
--- a/editing/edit-feature-attachments/src/main/java/com/esri/samples/edit_feature_attachments/EditFeatureAttachmentsSample.java
+++ b/editing/edit-feature-attachments/src/main/java/com/esri/samples/edit_feature_attachments/EditFeatureAttachmentsSample.java
@@ -18,7 +18,6 @@ package com.esri.samples.edit_feature_attachments;
 
 import org.apache.commons.io.IOUtils;
 import java.util.List;
-import java.util.concurrent.CompletableFuture;
 
 import javafx.application.Application;
 import javafx.application.Platform;
@@ -227,11 +226,9 @@ public class EditFeatureAttachmentsSample extends Application {
   private void addAttachment(byte[] attachment) {
 
     if (selected.canEditAttachments()) {
-      CompletableFuture<Attachment> addResult = selected.addAttachmentAsync(attachment, "image/png", "edit_feature_attachments/destroyed.png")
-        .toCompletableFuture();
-
-      // update feature table, apply update to server when new feature is added, and update the displayed list of attachments
-      addResult.thenCompose(addedAttachment -> featureTable.updateFeatureAsync(selected).toCompletableFuture())
+      selected.addAttachmentAsync(attachment, "image/png", "edit_feature_attachments/destroyed.png")
+        .toCompletableFuture()
+        .thenCompose(addedAttachment -> featureTable.updateFeatureAsync(selected).toCompletableFuture())
         .thenRun(() -> applyEdits(featureTable));
     } else {
       displayMessage(null, "Cannot add attachment.");
--- a/editing/edit-feature-attachments/src/main/java/com/esri/samples/edit_feature_attachments/EditFeatureAttachmentsSample.java
+++ b/editing/edit-feature-attachments/src/main/java/com/esri/samples/edit_feature_attachments/EditFeatureAttachmentsSample.java
@@ -159,11 +159,13 @@ public class EditFeatureAttachmentsSample extends Application {
             .whenComplete((identifyLayerResult, exception) -> {
               if (exception == null) {
                 // if the identification operation completes successfully, get the list of identified GeoElements
+                // if nothing is identified, an empty list will be returned
                 List<GeoElement> identified = identifyLayerResult.getElements();
                 if (!identified.isEmpty()) {
+                  // only proceed if an element was identified
                   GeoElement element = identified.get(0);
-                  // get selected feature
                   if (element instanceof ArcGISFeature) {
+                    // if the element is an ArcGISFeature, select the feature
                     selected = (ArcGISFeature) element;
                     featureLayer.selectFeature(selected);
                     selected.loadAsync();
--- a/map/generate-offline-map-overrides/src/main/java/com/esri/samples/generate_offline_map_overrides/GenerateOfflineMapOverridesController.java
+++ b/map/generate-offline-map-overrides/src/main/java/com/esri/samples/generate_offline_map_overrides/GenerateOfflineMapOverridesController.java
@@ -119,114 +119,117 @@ public class GenerateOfflineMapOverridesController {
    */
   @FXML
   private void generateOfflineMap() {
-    try {
-      // show the progress bar
-      progressBar.setVisible(true);
+    // show the progress bar
+    progressBar.setVisible(true);
 
-      // create an offline map task with the map
-      var offlineMapTask = new OfflineMapTask(map);
+    // create an offline map task with the map
+    var offlineMapTask = new OfflineMapTask(map);
 
-      // get default offline map parameters for this task given the download area
-      offlineMapTask.createDefaultGenerateOfflineMapParametersAsync(downloadArea.getGeometry())
-        .toCompletableFuture().whenComplete((parameters, ex) -> {
-          if (ex == null) {
-            // get additional offline parameters (overrides) for this task
-            offlineMapTask.createGenerateOfflineMapParameterOverridesAsync(parameters).toCompletableFuture()
-              .thenAccept(overrides -> {
-                  try {
-                    // get the export tile cache parameters for the base layer
-                    OfflineMapParametersKey basemapParamKey = new OfflineMapParametersKey(
-                      mapView.getMap().getBasemap().getBaseLayers().get(0));
-                    ExportTileCacheParameters exportTileCacheParameters =
-                      overrides.getExportTileCacheParameters().get(basemapParamKey);
+    // get default offline map parameters for this task given the download area
+    offlineMapTask.createDefaultGenerateOfflineMapParametersAsync(downloadArea.getGeometry())
+      .toCompletableFuture().whenComplete((parameters, ex) -> {
+        if (ex == null) {
+          // get additional offline parameters (overrides) for this task
+          offlineMapTask.createGenerateOfflineMapParameterOverridesAsync(parameters).toCompletableFuture()
+            .whenComplete((overrides, x) -> {
+              if (x == null) {
+                try {
+                  // get the export tile cache parameters for the base layer
+                  var basemapParamKey = new OfflineMapParametersKey(
+                    mapView.getMap().getBasemap().getBaseLayers().get(0));
+                  ExportTileCacheParameters exportTileCacheParameters =
+                    overrides.getExportTileCacheParameters().get(basemapParamKey);
 
-                    // create a new sublist of level IDs in the range requested by the user
-                    exportTileCacheParameters.getLevelIDs().clear();
-                    for (int i = minScaleLevelSpinner.getValue(); i < maxScaleLevelSpinner.getValue(); i++) {
-                      exportTileCacheParameters.getLevelIDs().add(i);
-                    }
-                    // set the area of interest to the original download area plus a buffer
-                    exportTileCacheParameters.setAreaOfInterest(GeometryEngine.buffer(downloadArea.getGeometry(),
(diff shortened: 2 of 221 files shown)


jenmerritt:
@01smito01 RE: the build versions, if you can revert any samples that you haven't changed as part of the completable futures work to 200.1.0, and then only the samples that have changed will show up in the diff. You're correct to update to a daily build in those samples where you're testing those changes. Thanks!

editing/edit-feature-attachments/src/main/java/com/esri/samples/edit_feature_attachments/EditFeatureAttachmentsSample.java
```diff
@@ -160,32 +155,31 @@ public void start(Stage stage) {
...
-                      fetchAttachments(selected);
-                    } else {
-                      displayMessage("Error", "Element failed to load!");
-                    }
-                  });
-                  addAttachmentButton.setDisable(false);
+          mapView.identifyLayerAsync(featureLayer, point, 1, false).toCompletableFuture()
+            .whenComplete((layer, ex) -> {
+              if (ex == null) {
+                List<GeoElement> identified = layer.getElements();
```
jenmerritt: I think we can add a comment within like
```suggestion
               // if the operation completes successfully, get the result
                List<GeoElement> identified = layer.getElements();
```

editing/edit-feature-attachments/src/main/java/com/esri/samples/edit_feature_attachments/EditFeatureAttachmentsSample.java
```diff
@@ -160,32 +155,31 @@ public void start(Stage stage) {
...
+                        fetchAttachments(selected);
+                      } else {
+                        displayMessage("Error", "Element failed to load!");
+                      }
+                    });
+                    addAttachmentButton.setDisable(false);
+                  }
                 }
+              } else {
+                displayMessage("Exception getting identify result", ex.getCause().getMessage());
```
jenmerritt:
```suggestion
               // if the operation completes with an exception, display an error message
                displayMessage("Exception getting identify result", ex.getCause().getMessage());
```

editing/edit-feature-attachments/src/main/java/com/esri/samples/edit_feature_attachments/EditFeatureAttachmentsSample.java
jenmerritt: I won't add this in each sample, but I think it's good doc to add where relevant across them all

map/generate-offline-map-overrides/src/main/java/com/esri/samples/generate_offline_map_overrides/GenerateOfflineMapOverridesController.java
```diff
@@ -131,111 +127,103 @@ private void generateOfflineMap() {
...
-        .createDefaultGenerateOfflineMapParametersAsync(downloadArea.getGeometry());
+      offlineMapTask.createDefaultGenerateOfflineMapParametersAsync(downloadArea.getGeometry())
+        .toCompletableFuture().whenComplete((parameters, ex) -> {
+          if (ex == null) {
+            // get additional offline parameters (overrides) for this task
+            offlineMapTask.createGenerateOfflineMapParameterOverridesAsync(parameters).toCompletableFuture()
+              .thenAccept(overrides -> {
+                  try {
+                    // get the export tile cache parameters for the base layer
+                    OfflineMapParametersKey basemapParamKey = new OfflineMapParametersKey(
```
jenmerritt:
```suggestion
                    var basemapParamKey = new OfflineMapParametersKey(
```
Where we have `new ClassName` I think we can use more `var`s to just reduce some words.

symbology/read-symbols-from-mobile-style-file/src/main/java/com/esri/samples/read_symbols_from_mobile_style_file/ReadSymbolsFromMobileStyleFileController.java
```diff
@@ -116,20 +114,12 @@ private void loadSymbolsFromStyleFile() {
...
-          // use the default parameters to perform the search, getting all the available symbols within the file
-          ListenableFuture<List<SymbolStyleSearchResult>> symbolStyleSearchResultFuture = emojiStyle.searchSymbolsAsync(defaultSearchParameters);
-          symbolStyleSearchResultFuture.addDoneListener(() -> {
-            try {
-
-              // loop through the results and add each item to a list view according to category
-              List<SymbolStyleSearchResult> symbolStyleSearchResults = symbolStyleSearchResultFuture.get();
-              symbolStyleSearchResults.forEach(symbolStyleSearchResult -> {
-
+      emojiStyle.getDefaultSearchParametersAsync().toCompletableFuture()
```
jenmerritt:
```suggestion
     // insert doc here
      emojiStyle.getDefaultSearchParametersAsync().toCompletableFuture()
```
I think we need some more explicit doc here

jenmerritt (commented):
@01smito01 your additions of the completable futures look fine to me and choice of samples seems good. My main comments relate to use needing to add some additional doc - particularly where we have the `if / else` for successful/unsuccessful completions as otherwise the `else` gets quite lost by the time youve scrolled down.

That said, I think it's probably sensible to get another experienced set of eyes over these at this point, and I'd be happy for you to wait until after that to add comments for the benefit of the sample, at which point I'm also happy to re-review.

editing/edit-feature-attachments/src/main/java/com/esri/samples/edit_feature_attachments/EditFeatureAttachmentsSample.java
01smito01: Not sure this is quite right - the layer _is_ the result. Not like before where we'd have a `IdentifyLayerResult` (or similar) which we'd need to call `get().getElements()` on (and thus surround with try/catch). Would something like "if the identification operation completes successfully, get the list of identified geoElements" be better?

jenmerritt: Happy for you to devise the comments as appropriate yep, it was more to demonstrate that a comment either side of the added if/else would be good and easier to pick them out.

01smito01:
@colinanderson this is ready for second review when you're free.

editing/edit-feature-attachments/src/main/java/com/esri/samples/edit_feature_attachments/EditFeatureAttachmentsSample.java
```diff
@@ -160,32 +155,33 @@ public void start(Stage stage) {
...
-                  selected.addDoneLoadingListener(() -> {
-                    if (selected.getLoadStatus() == LoadStatus.LOADED) {
-                      fetchAttachments(selected);
-                    } else {
-                      displayMessage("Error", "Element failed to load!");
-                    }
-                  });
-                  addAttachmentButton.setDisable(false);
+          mapView.identifyLayerAsync(featureLayer, point, 1, false).toCompletableFuture()
+            .whenComplete((layer, ex) -> {
```
colinanderson: >Not sure this is quite right - the layer is the result. Not like before where we'd have a IdentifyLayerResult (or similar) which we'd need to call get().getElements() on (and thus surround with try/catch). 

The result is still an `IdentifyLayerResult` so instead of calling it `layer` it should be `identifyLayerResult` or maybe just `result`.

colinanderson: Probably better to call it `exception` rather than `ex`. Possibly it should be `throwable` instead of `exception`.

editing/edit-feature-attachments/src/main/java/com/esri/samples/edit_feature_attachments/EditFeatureAttachmentsSample.java
```diff
@@ -160,32 +155,33 @@ public void start(Stage stage) {
...
+              if (ex == null) {
+                // if the identification operation completes successfully, get the list of identified GeoElements
+                List<GeoElement> identified = layer.getElements();
+                if (identified.size() > 0) {
```
colinanderson: I think there is an overload of the identify where you can specify the max number of results so you could specify one and then do an `isEmpty()`. Aren't you missing checking for nothing identified at the moment?

editing/edit-feature-attachments/src/main/java/com/esri/samples/edit_feature_attachments/EditFeatureAttachmentsSample.java
```diff
@@ -160,32 +155,33 @@ public void start(Stage stage) {
...
+                  GeoElement element = identified.get(0);
+                  // get selected feature
+                  if (element instanceof ArcGISFeature) {
+                    selected = (ArcGISFeature) element;
+                    featureLayer.selectFeature(selected);
+                    selected.loadAsync();
+                    selected.addDoneLoadingListener(() -> {
+                      if (selected.getLoadStatus() == LoadStatus.LOADED) {
+                        fetchAttachments(selected);
```
colinanderson: This looks like a non-async call but the method is actually doing async stuff so maybe call it `fetchAttachmentsAsync`.

editing/edit-feature-attachments/src/main/java/com/esri/samples/edit_feature_attachments/EditFeatureAttachmentsSample.java
```diff
@@ -203,25 +199,25 @@ public void start(Stage stage) {
...
-    });
+    feature.fetchAttachmentsAsync().toCompletableFuture().whenComplete(
+      (attachmentResults, ex) -> {
+        if (ex == null) {
+          // if the attachments were fetched successfully, update the UI attachments list
+          attachments = attachmentResults;
+
+          Platform.runLater(() -> {
+            attachmentList.getItems().clear();
+            attachments.forEach(attachment -> attachmentList.getItems().add(attachment.getName()));
```
colinanderson: Wouldn't it make more sense to do this otherwise you are iterating the empty list which works but seems pointless if you are going to test for empty on the next line.
```
if (!attachments.isEmpty()) {
  attachmentsLabel.setText("Attachments: ");
  attachments.forEach(attachment -> attachmentList.getItems().add(attachment.getName()));
} else {
  attachmentsLabel.setText("No Attachments!");
}

File: editing/edit-feature-attachments/src/main/java/com/esri/samples/edit_feature_attachments/EditFeatureAttachmentsSample.java [hunk 9]
```diff
@@ -273,26 +260,25 @@ private void deleteAttachment(int attachmentIndex) {
...
-          if (!edits.get(0).hasCompletedWithErrors()) {
-            displayMessage(null, "Edited feature successfully");
-          } else {
-            if (edits.get(0).getError() != null) {
-              throw edits.get(0).getError();
+    featureTable.applyEditsAsync().toCompletableFuture().whenComplete(
+      (edits, ex) -> {
+        if (ex == null) {
+          // check if the server edit was successful
+          if (edits != null && edits.size() > 0) {
```
colinanderson: Can `edits` actually be `null` here - probably it is an empty list if there are no results.

Shouldn't it be `editResults` rather than `edits`?

editing/edit-feature-attachments/src/main/java/com/esri/samples/edit_feature_attachments/EditFeatureAttachmentsSample.java
```diff
@@ -273,26 +260,25 @@ private void deleteAttachment(int attachmentIndex) {
...
+            if (!edits.get(0).hasCompletedWithErrors()) {
```
colinanderson: We're assuming only one edit here - is this good sample code? Might be better to show how you would handle multiple results.

editing/edit-feature-attachments/src/main/java/com/esri/samples/edit_feature_attachments/EditFeatureAttachmentsSample.java
```diff
@@ -273,26 +260,25 @@ private void deleteAttachment(int attachmentIndex) {
...
+              displayMessage(null, "Edited feature successfully");
+            } else {
+              if (edits.get(0).getError() != null) {
+                throw edits.get(0).getError();
+              }
             }
           }
+          // update the displayed list of attachments
+          fetchAttachments(selected);
```
colinanderson: See comment above - looks like this is a non-blocking call so should probably be `fetchAttachmentsAsync(selected)`.

map/generate-offline-map-overrides/src/main/java/com/esri/samples/generate_offline_map_overrides/GenerateOfflineMapOverridesController.java
```diff
@@ -123,122 +119,117 @@ private void initialize() {
...
-
-      generateOfflineMapParametersFuture.addDoneListener(() -> {
-        try {
-          final GenerateOfflineMapParameters parameters = generateOfflineMapParametersFuture.get();
+    // create an offline map task with the map
+    var offlineMapTask = new OfflineMapTask(map);
 
+    // get default offline map parameters for this task given the download area
+    offlineMapTask.createDefaultGenerateOfflineMapParametersAsync(downloadArea.getGeometry())
+      .toCompletableFuture().whenComplete((parameters, ex) -> {
```
colinanderson: It might be possible to do
`.toCompletableFuture().exceptionally(...).theAccept(...)` //or combine etc. as a way to handle exceptions.

map/generate-offline-map-overrides/src/main/java/com/esri/samples/generate_offline_map_overrides/GenerateOfflineMapOverridesController.java
```diff
@@ -123,122 +119,117 @@ private void initialize() {
...
+        if (ex == null) {
           // get additional offline parameters (overrides) for this task
-          ListenableFuture<GenerateOfflineMapParameterOverrides> parameterOverridesFuture = offlineMapTask
-            .createGenerateOfflineMapParameterOverridesAsync(parameters);
-
-          parameterOverridesFuture.addDoneListener(() -> {
-            try {
-              GenerateOfflineMapParameterOverrides overrides = parameterOverridesFuture.get();
+          offlineMapTask.createGenerateOfflineMapParameterOverridesAsync(parameters).toCompletableFuture()
+            .whenComplete((overrides, x) -> {
```
colinanderson: I think we should use better names than `ex` and `x`.

map/generate-offline-map-overrides/src/main/java/com/esri/samples/generate_offline_map_overrides/GenerateOfflineMapOverridesController.java
```diff
@@ -123,122 +119,117 @@ private void initialize() {
...
-                    while (layerOptionsIterator.hasNext()) {
-                      GenerateLayerOption layerOption = layerOptionsIterator.next();
-                      if (layerOption.getLayerId() == layerId) {
-                        switch (layer.getName()) {
-                          // remove the System Valve layer from the layer options if it should not be included
-                          case "System Valve":
-                            if (!systemValvesCheckBox.isSelected()) {
-                              layerOptionsIterator.remove();
+                  // configure layer option parameters for each layer depending on the options selected in the UI
+                  for (Layer layer : map.getOperationalLayers()) {
```
colinanderson: Can't you do something more modern like
```
map.getOperationalLayers().stream()
  .filter(l -> l instanceof FeatureLayer)
  .map(l -> (FeatureLayer) l)...
```

map/generate-offline-map-overrides/src/main/java/com/esri/samples/generate_offline_map_overrides/GenerateOfflineMapOverridesController.java
```diff
@@ -123,122 +119,117 @@ private void initialize() {
...
+                      var featureLayer = (FeatureLayer) layer;
+                      ServiceFeatureTable featureTable = (ServiceFeatureTable) featureLayer.getFeatureTable();
+                      long layerId = featureTable.getLayerInfo().getServiceLayerId();
+                      // get the layer option parameters specifically for this layer
+                      var offlineMapParametersKey = new OfflineMapParametersKey(layer);
+                      GenerateGeodatabaseParameters generateGeodatabaseParameters = overrides.getGenerateGeodatabaseParameters()
+                        .get(offlineMapParametersKey);
+                      List<GenerateLayerOption> layerOptions = generateGeodatabaseParameters.getLayerOptions();
+                      // use an iterator so we can remove layer options while looping over them
+                      Iterator<GenerateLayerOption> layerOptionsIterator = layerOptions.iterator();
```
colinanderson, line 165: Do you actually need to use `Iterator` here? Maybe you can just `foreach` or `stream` it?

map/generate-offline-map-overrides/src/main/java/com/esri/samples/generate_offline_map_overrides/GenerateOfflineMapOverridesController.java
```diff
@@ -123,122 +119,117 @@ private void initialize() {
...
+                      if (!layerOptions.isEmpty()) {
+                        while (layerOptionsIterator.hasNext()) {
+                          GenerateLayerOption layerOption = layerOptionsIterator.next();
+                          if (layerOption.getLayerId() == layerId) {
```
colinanderson, line 169: If you can stream it then you should be able to use `filter` here e.g. 
```
.filter(layerOption -> layerOption.getLayerId() == layerId)
```

colinanderson: Maybe you need to use an iterator because you are changing the list... might be worth seeing if there is a better way to do this.

map/mobile-map-search-and-route/src/main/java/com/esri/samples/mobile_map_search_and_route/MobileMapSearchAndRouteSample.java
```diff
@@ -90,7 +87,7 @@ public void start(Stage stage) {
       mapPackageListView.setMaxSize(170, 100);
 
       // create a file chooser to find local mmpk files
-      FileChooser fileChooser = new FileChooser();
+      var fileChooser = new FileChooser();
       FileChooser.ExtensionFilter mmpkFilter = new FileChooser.ExtensionFilter("Map Packages (*.mmpk)", "*.mmpk");
       fileChooser.getExtensionFilters().add(mmpkFilter);
       fileChooser.setInitialDirectory(new File(System.getProperty("data.dir"), "./samples-data/mmpk"));
```
colinanderson: It would probably be better to use `Path.resolve` etc. here because you'e hardcoded the path separator and I'm sure in other places it uses something to avoid doing that (https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/io/File.html#separatorChar maybe).

map/mobile-map-search-and-route/src/main/java/com/esri/samples/mobile_map_search_and_route/MobileMapSearchAndRouteSample.java
```diff
@@ -213,77 +210,69 @@ protected void updateItem(ArcGISMap map, boolean bln) {
...
-                      RouteParameters routeParameters = defaultParameters.get();
-                      routeParameters.setStops(stops);
-                      // solve the route and display the result graphic
-                      ListenableFuture<RouteResult> routeResults = routeTask.solveRouteAsync(routeParameters);
-                      routeResults.addDoneListener(() -> {
-                        try {
-                          RouteResult result = routeResults.get();
-                          Route route = result.getRoutes().get(0);
+          mobileMapPackage.getLocatorTask().reverseGeocodeAsync(mapPoint, reverseGeocodeParameters).toCompletableFuture()
+            .whenComplete((geocodes, ex) -> {
```
colinanderson: How about `geocodeResults` instead of `geocodes`?

map/mobile-map-search-and-route/src/main/java/com/esri/samples/mobile_map_search_and_route/MobileMapSearchAndRouteSample.java
```diff
@@ -213,77 +210,69 @@ protected void updateItem(ArcGISMap map, boolean bln) {
...
+              if (ex == null) {
+                // show a pin graphic and a callout with the geocode's address
+                if (geocodes.size() > 0) {
```
colinanderson: As commented for other sample - what if there is more than one result? How about commenting that we are only going to consider the first result in this sample?

map/mobile-map-search-and-route/src/main/java/com/esri/samples/mobile_map_search_and_route/MobileMapSearchAndRouteSample.java
```diff
@@ -213,77 +210,69 @@ protected void updateItem(ArcGISMap map, boolean bln) {
...
+                  GeocodeResult geocode = geocodes.get(0);
+                  Point location = geocode.getDisplayLocation();
+
+                  // get attributes from the result for the callout
+                  String address = geocode.getAttributes().get("Match_addr").toString();
+                  HashMap<String, Object> attributes = new HashMap<>();
```
colinanderson: I'd call this `graphicAttributes` because it is a little confusing that it called `geocode.getAttributes()` on the line above.

map/mobile-map-search-and-route/src/main/java/com/esri/samples/mobile_map_search_and_route/MobileMapSearchAndRouteSample.java
```diff
@@ -213,77 +210,69 @@ protected void updateItem(ArcGISMap map, boolean bln) {
...
+                  if (!mapView.getMap().getTransportationNetworks().isEmpty() && locationsGraphicsOverlay.getGraphics().size() > 1) {
+                    // use the first transportation network dataset in the map to create a route task
+                    TransportationNetworkDataset networkDataset = mapView.getMap().getTransportationNetworks().get(0);
+                    var routeTask = new RouteTask(networkDataset);
+                    // create route parameters with the last two locations' stops
+                    List<Graphic> locationGraphics = locationsGraphicsOverlay.getGraphics();
+                    List<Stop> stops = locationGraphics.subList(Math.max(locationGraphics.size() - 2, 0), locationGraphics.size())
+                      .stream()
+                      .map(g -> new Stop((Point) g.getGeometry()))
+                      .collect(Collectors.toList());
```
colinanderson: I think with Java17 you can just do `.toList()`.

map/mobile-map-search-and-route/src/main/java/com/esri/samples/mobile_map_search_and_route/MobileMapSearchAndRouteSample.java
```diff
@@ -213,77 +210,69 @@ protected void updateItem(ArcGISMap map, boolean bln) {
...
+                    routeTask.createDefaultParametersAsync().toCompletableFuture()
+                      .thenCompose(routeParameters -> {
+                        routeParameters.setStops(stops);
+                        // solve the route and display the result graphic
+                        return routeTask.solveRouteAsync(routeParameters).toCompletableFuture();
+                      }).whenComplete((routeResult, x) -> {
```
colinanderson: `x` is not a great name.

symbology/read-symbols-from-mobile-style-file/src/main/java/com/esri/samples/read_symbols_from_mobile_style_file/ReadSymbolsFromMobileStyleFileController.java
```diff
@@ -104,7 +102,7 @@ public void initialize() {
    */
   private void loadSymbolsFromStyleFile() {
     // create a SymbolStyle with the .stylx file
-    File stylxFile = new File(System.getProperty("data.dir"), "./samples-data/stylx/emoji-mobile.stylx");
+    var stylxFile = new File(System.getProperty("data.dir"), "./samples-data/stylx/emoji-mobile.stylx");
```
colinanderson: As commented above try using `resolve` (https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/nio/file/Path.html#resolve(java.lang.String)) as a nicer way to make paths and handle separators.

symbology/read-symbols-from-mobile-style-file/src/main/java/com/esri/samples/read_symbols_from_mobile_style_file/ReadSymbolsFromMobileStyleFileController.java
```diff
@@ -173,33 +159,33 @@ private void buildCompositeSymbol() {
...
 
     // get the symbol from the SymbolStyle
-    ListenableFuture<Symbol> symbolFuture = emojiStyle.getSymbolAsync(symbolKeys);
-    symbolFuture.addDoneListener(() -> {
-      try {
-        faceSymbol = (MultilayerPointSymbol) symbolFuture.get();
-        if (faceSymbol == null) {
-          return;
+    emojiStyle.getSymbolAsync(symbolKeys).toCompletableFuture().whenComplete(
+      (symbol, ex)  -> {
```
colinanderson: I would split the lines with the `whenComplete(` on the same line as `(symbol, ex)  -> {` e.g.
```
.whenComplete((symbol, ex)  -> {
```

symbology/read-symbols-from-mobile-style-file/src/main/java/com/esri/samples/read_symbols_from_mobile_style_file/SymbolLayerInfoListCell.java
```diff
@@ -33,29 +27,28 @@ protected void updateItem(SymbolStyleSearchResult item, boolean empty) {
...
-      });
+      item.getSymbolAsync().toCompletableFuture().whenComplete(
+        (symbol, ex) -> {
+          if (ex == null) {
+            // create a bitmap swatch from the symbol
+            symbol.createSwatchAsync(Color.TRANSPARENT, 1).toCompletableFuture()
+              .whenComplete((symbolImage, e) -> {
+                if (e == null) {
+                  // update the image view with the symbol swatch
+                  symbolImageView.setImage(symbolImage);
```
colinanderson: Is the flickering fixed or was that a different sample?

symbology/symbol-dictionary/src/main/java/com/esri/samples/symbol_dictionary/SymbolDictionaryController.java
```diff
@@ -51,7 +51,7 @@ public class SymbolDictionaryController {
 
   public void initialize() {
     // loads a specification for the symbol dictionary
-    File stylxFile = new File(System.getProperty("data.dir"), "./samples-data/stylx/mil2525d.stylx");
+    var stylxFile = new File(System.getProperty("data.dir"), "./samples-data/stylx/mil2525d.stylx");
```
colinanderson: See previous comments about using `resolve` etc.

symbology/symbol-dictionary/src/main/java/com/esri/samples/symbol_dictionary/SymbolView.java
```diff
@@ -78,26 +74,19 @@ public void initialize(URL location, ResourceBundle resources) {
...
-        } catch (InterruptedException | ExecutionException e) {
-          new Alert(Alert.AlertType.ERROR, "Error getting symbol" + e.getMessage()).show();
-        }
-      });
-
+      // get the symbol and create a swatch from it
+      styleSymbolSearchResult.getSymbolAsync().toCompletableFuture()
+        .thenCompose(symbol ->
+          symbol.createSwatchAsync(40, 40, Color.color(1.0, 1.0, 1.0, 0.0), new Point(0, 0, 0)).toCompletableFuture())
+        .whenComplete((image, ex) -> {
```
colinanderson, line 81: indentation looks off here.

utility_network/perform-valve-isolation-trace/src/main/java/com/esri/samples/perform_valve_isolation_trace/PerformValveIsolationTraceController.java
```diff
@@ -240,64 +237,59 @@ private void handleTraceClick() {
...
 
-            if (!utilityElementTraceResult.getElements().isEmpty()) {
+              if (!utilityElementTraceResult.getElements().isEmpty()) {
 
-              // iterate through the map's feature layers
-              mapView.getMap().getOperationalLayers().forEach(layer -> {
-                if (layer instanceof FeatureLayer) {
+                // iterate through the map's feature layers
+                mapView.getMap().getOperationalLayers().forEach(layer -> {
+                  if (layer instanceof FeatureLayer) {
```
colinanderson: Use `filter(layer -> layer instanceof FeatureLayer).map(layer -> (FeatureLayer) layer)`?

utility_network/perform-valve-isolation-trace/src/main/java/com/esri/samples/perform_valve_isolation_trace/PerformValveIsolationTraceController.java
```diff
@@ -240,64 +237,59 @@ private void handleTraceClick() {
...
-                      queryParameters.getObjectIds().add(utilityElement.getObjectId());
-                    }
-                  });
+                      String networkSourceName = utilityElement.getNetworkSource().getName();
+                      String featureTableName = ((FeatureLayer) layer).getFeatureTable().getTableName();
 
-                  // select features that match the query
-                  ListenableFuture<FeatureQueryResult> featureQueryResultListenableFuture =
-                          ((FeatureLayer) layer).selectFeaturesAsync(queryParameters, FeatureLayer.SelectionMode.NEW);
+                      if (networkSourceName.equals(featureTableName)) {
```
colinanderson: You can maybe filter and map here.

utility_network/perform-valve-isolation-trace/src/main/java/com/esri/samples/perform_valve_isolation_trace/PerformValveIsolationTraceController.java
```diff
@@ -240,64 +237,59 @@ private void handleTraceClick() {
...
+                        queryParameters.getObjectIds().add(utilityElement.getObjectId());
+                      }
+                    });
+
+                    // select features that match the query
+                    ((FeatureLayer) layer).selectFeaturesAsync(queryParameters, FeatureLayer.SelectionMode.NEW)
+                      .toCompletableFuture().thenRun(() -> {
```
colinanderson: I'm not sure about this one - seems like since `selectFeaturesAsync` has a result you should check the result with a `whenComplete` or `excptionally`.

utility_network/perform-valve-isolation-trace/src/main/java/com/esri/samples/perform_valve_isolation_trace/PerformValveIsolationTraceController.java
```diff
@@ -332,97 +324,91 @@ private void handleMapViewClicked(MouseEvent e) {
...
+
+                // create utility element for each feature and store it in a list
+                List<UtilityElement> utilityElementList = new ArrayList<>();
+                listOfFeatures.forEach(feature -> utilityElementList.add(utilityNetwork.createElement(feature)));
+
+                // check the utility network source type of every utility element
+                utilityElementList.forEach(utilityElement -> {
+
+                  // check if the network source is a junction or an edge
+                  if (utilityElement.getNetworkSource().getSourceType() == UtilityNetworkSource.Type.JUNCTION) {
```
colinanderson: Can you do a `filter` for this?

colinanderson (commented):
@01smito01 a few comments to look at. Some relate to the completable futures and others are just general stuff in these samples - modernisation etc.

editing/edit-feature-attachments/src/main/java/com/esri/samples/edit_feature_attachments/EditFeatureAttachmentsSample.java
01smito01: The only methods that call `applyEdits` are attaching or deleting a single attachment and then immediately calling `applyEdits`, so the method never has to deal with more than one edit AFAIK. Method name is a bit unfortunate since it's really just "apply edit".

editing/edit-feature-attachments/src/main/java/com/esri/samples/edit_feature_attachments/EditFeatureAttachmentsSample.java
01smito01: `layer.getElements()` returns an empty list if nothing was identified, so the check for `identified.size` (have replaced with `!identified.isEmpty()`) works fine as is. If there aren't any identified results, then the user didn't click near an icon, and we don't need to do anything in that situation.

map/generate-offline-map-overrides/src/main/java/com/esri/samples/generate_offline_map_overrides/GenerateOfflineMapOverridesController.java
01smito01: Tried `foreach` here but it didn't work (UI toggles became nonfunctional). No obvious/easy way to do it with streams either as we're not removing options based on any attribute of the layer option.

Honestly this bit is really weird and I don't like it, but not worth spending any more time on for now - testing is quite slow as I have to wait for the offline map to generate before I can see if my changes did anything.

map/mobile-map-search-and-route/src/main/java/com/esri/samples/mobile_map_search_and_route/MobileMapSearchAndRouteSample.java
01smito01: The implementation of `reverseGeocodeAsync` only returns one location:
image

Could change in future, but I don't think that's worth commenting on. I'll add a comment that mentions that `geocodes` is a singleton list.

map/generate-offline-map-overrides/src/main/java/com/esri/samples/generate_offline_map_overrides/GenerateOfflineMapOverridesController.java
jenmerritt: @01smito01 feel free to create a separate issue for this in the Java repo if you think it's worth exploring further and separately from completable futures. We can get the CF in and then revisit some of the sample improvements separately where they are maybe less simple changes. If you do, just make a note in the thread here and then we know where we are for reviewing. 👍

symbology/read-symbols-from-mobile-style-file/src/main/java/com/esri/samples/read_symbols_from_mobile_style_file/SymbolLayerInfoListCell.java
01smito01: Unfortunately not fixed - I've left the offending method (`ReadSymbolsFromMobileStyleFileController.updateSymbolPreview`) using a ListenableFuture for now, which avoids the issues but isn't a particularly satisfying conclusion.

symbology/read-symbols-from-mobile-style-file/src/main/java/com/esri/samples/read_symbols_from_mobile_style_file/ReadSymbolsFromMobileStyleFileController.java
01smito01: My understanding is we're using the two-argument `File(String, String)` constructor as this will behave like `File(String)` if the first argument is null (which it will be unless the user has specified a different datadir themselves). I don't think `Path` resolves null so cleanly but could be wrong.

In the cases where we have file separators in the second argument I've been using `Path.of("path", "to", "the", "file").toString()` instead of `resolve`, which IMO is neater than chaining `resolve` calls. If there's a good reason to prefer `resolve` I'll switch to that.

symbology/symbol-dictionary/src/main/java/com/esri/samples/symbol_dictionary/SymbolView.java
01smito01: Purely in terms of the indenting on each line it looks fine to me - ran the autoindenter and that didn't make any changes either. Unless you mean in terms of having the right hand side of the lambda on its own line?

utility_network/perform-valve-isolation-trace/src/main/java/com/esri/samples/perform_valve_isolation_trace/PerformValveIsolationTraceController.java
01smito01: The original `ListenableFuture` here didn't catch any exceptions either - probably because launching the sample and pressing "trace" without changing any other setting throws a `JsonEmbeddedException`. Doesn't crash the application at all, but it's not clear to me what's causing it or how to fix it (maybe that's why it was never caught in the first place?).

symbology/read-symbols-from-mobile-style-file/src/main/java/com/esri/samples/read_symbols_from_mobile_style_file/ReadSymbolsFromMobileStyleFileController.java
colinanderson: Handling of null is easily fixed with `System.getProperty("data.dir", "")`.

utility_network/perform-valve-isolation-trace/src/main/java/com/esri/samples/perform_valve_isolation_trace/PerformValveIsolationTraceController.java
01smito01: We've got a decent bit of if/else-if logic so unless I'm missing something this only really works as two separate stream operations - tried that and it's not better than what we already have to my eyes.

symbology/symbol-dictionary/src/main/java/com/esri/samples/symbol_dictionary/SymbolView.java
colinanderson: Yeah, the split line made it look off.

editing/edit-feature-attachments/src/main/java/com/esri/samples/edit_feature_attachments/EditFeatureAttachmentsSample.java
jenmerritt: Edit: this is done ✅ 

Let's make the comments more explicit here. As you say we don't want to do anything if there is nothing identified in this app, but let's make that absolutely clear so it's easier for someone to pick out if they did want an action if nothing was identified. I'm not tied to it word for word but easier to write a suggestion.

```          // identify a clicked feature by performing an identify layer operation on the MapView, returning a maximum of 1 result
          mapView.identifyLayerAsync(featureLayer, point, 1, false, 1).toCompletableFuture()
            .whenComplete((identifyLayerResult, exception) -> {
              if (exception == null) {
                // if the identification operation completes successfully, get the list of identified GeoElements
                // if nothing is identified, an empty list will be returned
                List<GeoElement> identified = identifyLayerResult.getElements();
                if (!identified.isEmpty()) {
                  // only proceed if an element was identified
                  GeoElement element = identified.get(0);
```

editing/edit-feature-attachments/src/main/java/com/esri/samples/edit_feature_attachments/EditFeatureAttachmentsSample.java
jenmerritt: Edit: below comment update is done ✅ 

In the case of this sample we are applying a single feature edit to the server each time and applying straight away, rather than like an "apply edits" button or something. Reading the doc it sounds like there'd be a result per feature, "This returns edit results for each feature, each result includes attachment results for that feature.". So that'll be where the assumption came from in the original implementation.

It looks like other instances of `applyEdits` have been implemented in this same way in the samples. So if this is an improvement here to ensure we capture any possible results, let's take this out of this PR and I'll create a separate issue to apply in all places (and avoid it becoming disjointed advice).

In the meantime for this PR, let's add some doc here:
```
        // check if the server edit was successful
        if (exception == null) {
            // check for any errors. In this case we are only updating 1 feature so just check the first result.
            if (!editResults.get(0).hasCompletedWithErrors()) {
              displayMessage(null, "Edited feature successfully");
```

jenmerritt: I'm also noting feedback elsewhere about methods involving an async operation and not being noted as async in the method name. I'll add this to the issue to look more generally at applying edits in samples.

map/generate-offline-map-overrides/src/main/java/com/esri/samples/generate_offline_map_overrides/GenerateOfflineMapOverridesController.java
```diff
@@ -123,122 +119,117 @@ private void initialize() {
...
-              // show the job's progress with the progress bar
-              job.addProgressChangedListener(() -> progressBar.setProgress(job.getProgress() / 100.0));
-            } catch (ExecutionException | InterruptedException | IOException ex) {
-              new Alert(Alert.AlertType.ERROR, "Error configuring the offline map task").show();
-            }
-          });
-        } catch (Exception ex) {
-          new Alert(Alert.AlertType.ERROR, "Error creating override parameters").show();
+              } else {
+                // if the generation of offline map parameter overrides completed exceltipnally, display an error
```
jenmerritt:
```suggestion
                // if the generation of offline map parameter overrides completed exceptionally, display an error
```

map/generate-offline-map-overrides/src/main/java/com/esri/samples/generate_offline_map_overrides/GenerateOfflineMapOverridesController.java
jenmerritt: I can see the benefits here where we've got more bracket indentations going on, of being able to see both the exception actions and the successful actions without needing to find the right bracket further down. Assuming I've got the approach right here:
```java
offlineMapTask.createGenerateOfflineMapParameterOverridesAsync(parameters).toCompletableFuture().exceptionally(ex -> {
             // if the operation completes with an exception, display an alert
              new Alert(Alert.AlertType.ERROR, "Error creating override parameters: " + e.getMessage()).show();
              return null;
            }).thenAccept(overrides -> {
             // if the operation completes successfully
              try {
                // get the export tile cache parameters for the base layer
                var basemapParamKey = new OfflineMapParametersKey(
                    mapView.getMap().getBasemap().getBaseLayers().get(0));
                ExportTileCacheParameters exportTileCacheParameters =
                    overrides.getExportTileCacheParameters().get(basemapParamKey);
// etc etc etc
```

Returning null feels a bit weird but as far as I can tell it's what you'd do? Any additional thoughts on benefits of the approach above (if it's valid) vs using `whenComplete` here @colinanderson ? Is it purely readability side I noted, or are there other potential benefits?

map/generate-offline-map-overrides/src/main/java/com/esri/samples/generate_offline_map_overrides/GenerateOfflineMapOverridesController.java
jenmerritt: I'll create a separate issue to look at this per above comment.

Also linking to this comment to reference: https://github.com/Esri/arcgis-maps-sdk-java-samples/pull/788#discussion_r1206917324

map/mobile-map-search-and-route/src/main/java/com/esri/samples/mobile_map_search_and_route/MobileMapSearchAndRouteSample.java
jenmerritt: Edit: below change is done ✅ 
I think the comment addition is good here and lines up the approach with https://github.com/Esri/arcgis-maps-sdk-java-samples/pull/788#discussion_r1218005017. 
But let's change it to `!geocodeResults.isEmpty()` to again match what's been done in the above.

symbology/read-symbols-from-mobile-style-file/src/main/java/com/esri/samples/read_symbols_from_mobile_style_file/ReadSymbolsFromMobileStyleFileController.java
jenmerritt: Looks like we should improve how we're handling Files in Samples and use `resolve`. I'll create a separate issue for this, rather than just update in this sample. Thanks for highlighting this improvement.

symbology/read-symbols-from-mobile-style-file/src/main/java/com/esri/samples/read_symbols_from_mobile_style_file/SymbolLayerInfoListCell.java
jenmerritt: I suggest taking this sample out of this PR and I'll create a separate issue for it as it sounds like we are better off not patching it with completable futures and then leaving that bit with a ListenableFuture if it wasn't quite right in the first place with needing to use that `get`. Let's explore it separately and see if completable futures can work after a bit of a rewrite.

utility_network/perform-valve-isolation-trace/src/main/java/com/esri/samples/perform_valve_isolation_trace/PerformValveIsolationTraceController.java
jenmerritt: Hmm. So previously, we're updating/resetting the UI regardless of what happens - which I think is right, and so `thenRun` is good to ensure we do that regardless. But I think it was a weakness before to not be checking, as if someone tinkered with the code and ended up passing null queryParameters for example, this could throw. So, dependent on @colinanderson 's feedback on this approach above... https://github.com/Esri/arcgis-maps-sdk-java-samples/pull/788#discussion_r1218193848

...maybe we can alter it to something like:
```java
                    layer.selectFeaturesAsync(queryParameters, FeatureLayer.SelectionMode.NEW)
                      .toCompletableFuture().exceptionally(exception -> {
                        new Alert(Alert.AlertType.ERROR, "Error selecting features: " + exception.getMessage());
                        return null;
                        }).thenRun(() -> {
                        // update the status text, enable the buttons and hide the progress indicator
                        statusLabel.setText("Isolation trace completed.");
                        enableUI(true);
                      });
```

utility_network/perform-valve-isolation-trace/src/main/java/com/esri/samples/perform_valve_isolation_trace/PerformValveIsolationTraceController.java
jenmerritt: Yeah I think a filter and a map works quite nicely here: 
```java
                        // populate the query parameters with the object IDs of any utility element in the trace result whose network source is equal to the layer's feature table
                       queryParameters.getObjectIds().addAll(utilityElementTraceResult.getElements().stream()
                          .filter(utilityElement ->
                              utilityElement.getNetworkSource().getName().equals(layer.getFeatureTable().getTableName()))
                          .map(UtilityElement::getObjectId).toList());
```

utility_network/perform-valve-isolation-trace/src/main/java/com/esri/samples/perform_valve_isolation_trace/PerformValveIsolationTraceController.java
jenmerritt: This is what I tried:
```java
                 // handle junctions
                  utilityElementList.stream().filter(utilityElement ->
                          utilityElement.getNetworkSource().getSourceType() == UtilityNetworkSource.Type.JUNCTION).toList()
                      .stream().filter(junction -> junction.getAssetType().getTerminalConfiguration() != null && junction.getAssetType().getTerminalConfiguration().getTerminals().size() > 1).toList()
                      .forEach(junction -> {
                        // prompt the user to select a terminal for this feature
                        Optional<UtilityTerminal> userSelectedTerminal = promptForTerminalSelection(junction.getAssetType().getTerminalConfiguration().getTerminals());

                        // apply the selected terminal
                        if (userSelectedTerminal.isPresent()) {
                          UtilityTerminal terminal = userSelectedTerminal.get();
                          junction.setTerminal(terminal);
                          // show the terminals name in the status label
                          String terminalName = terminal.getName() != null ? terminal.getName() : "default";
                          statusLabel.setText("Feature added at terminal: " + terminalName);

                          // don't create the element if no terminal was selected
                        } else {
                          statusLabel.setText("No terminal selected - no feature added");
                        }
                      });

                  // handle edges
                  utilityElementList.stream().filter(utilityElement -> utilityElement.getNetworkSource().getSourceType() == UtilityNetworkSource.Type.EDGE).toList()
                      .forEach(edge -> {
                        // get the geometry of the identified feature as a polyline, and remove the z component
                        Polyline polyline = (Polyline) GeometryEngine.removeZ(listOfFeatures.get(0).getGeometry());

                        // compute how far the clicked location is along the edge feature
                        double fractionAlongEdge = GeometryEngine.fractionAlong(polyline, mapPoint, -1);
                        if (Double.isNaN(fractionAlongEdge)) {
                          new Alert(Alert.AlertType.ERROR, "Cannot add starting location / barrier here.");
                          return;
                        }

                        // set the fraction along edge
                        edge.setFractionAlongEdge(fractionAlongEdge);
                        // update the status label text
                        statusLabel.setText("Fraction along edge: " + Math.round(edge.getFractionAlongEdge() * 1000d) / 1000d);
                      });
```
I think I agree in terms of legibility preferring an if/else, however if we are better making heavier use of Streams like this then I think we should. We can work on improving the comments to better match what's going on to accompany the chanining. 

Thoughts @colinanderson on the above approach and if we should a) go ahead and implement for this sample and b) make better/more use of streams throughout the samples?

jenmerritt:
@colinanderson This PR is ready for you to come back to thanks - the open conversations above have a few queries to run by you. Some of the feedback I've pulled out to form issues to look at across the samples and avoid bloating the primary goal of the completable futures here. But there's a few improvements it would be good to make - dependent on your feedback on the code snippets within the various comment threads. Thanks!

map/generate-offline-map-overrides/src/main/java/com/esri/samples/generate_offline_map_overrides/GenerateOfflineMapOverridesController.java
colinanderson: Probably `exceptionally` works better when you can provide a default value and `null` is probably not a good value. `whenComplete` is ok here. If the future could return a empty list for example then exceptionally might be cleaner looking because it could return an empty list and the `thenApply` would work as normal - assuming it already handled empty list which it should.

utility_network/perform-valve-isolation-trace/src/main/java/com/esri/samples/perform_valve_isolation_trace/PerformValveIsolationTraceController.java
colinanderson: If I was doing it I'd probably have used the filter just because I prefer that to looping and branching but it does mean sometimes you end up streaming twice (or more). Either way is fine.

utility_network/perform-valve-isolation-trace/src/main/java/com/esri/samples/perform_valve_isolation_trace/PerformValveIsolationTraceController.java
colinanderson: Yeah, @jenmerritt we want to make sure to handle any errors and your code looks like it does that. Even if we don't expect errors we should show handling them because users will copy this code and may never realise they are encountering an error. An alternative is to do something like
```
whenComplete((exception, result) -> {
  if (exception != null) {
    // alert
  }
  // update UI
})
```
There is a little weirdness that the UI will say "Isolation trace completed." even though it actually failed but that's probably beyond the scope of your sample.

01smito01: Done a bit of poking around on this: we only seem to get the exception when the combobox is set to "Subnetwork Controller", the checkbox is unselected, and the user hasn't placed any filter barriers on the map. The `JsonEmbeddedException` is fairly unhelpful: message is "Unable to complete operation." and details "Unable to perform query operation." I'm wondering if it's best to split off this problem as a new issue.

Incidentally, any other combobox selection is fine without any filter barriers - we just run into the message we've defined for if there are no results.

jenmerritt: Can also replicate using the Isolating option too. I'll create an issue as this isn't specific to completable futures. Replace with the below to better match our other implementations of CF:
```
                    // select features that match the query
                    layer.selectFeaturesAsync(queryParameters, FeatureLayer.SelectionMode.NEW)
                      .toCompletableFuture().whenComplete((result, exception) -> {
                        // update the status text, enable the buttons and hide the progress indicator
                        statusLabel.setText("Isolation trace completed.");
                        enableUI(true);
                      });
```
And we will resolve the exception handling separately since there isn't currently any either.

jenmerritt:
Is this ready for re-review @01smito01 ? When it is, assign Colin first and then I'll review comments / doc again last.

Note to self: this should wait to merge once website is pointing at `main` again.

01smito01:
Yeah I think this is ready for re-review. @colinanderson can you take a look when available?

editing/edit-feature-attachments/src/main/java/com/esri/samples/edit_feature_attachments/EditFeatureAttachmentsSample.java
```diff
@@ -232,16 +230,10 @@ private void fetchAttachments(ArcGISFeature feature) {
...
-              "edit_feature_attachments/destroyed.png");
-      addResult.addDoneListener(() -> {
-        // update feature table
-        ListenableFuture<Void> tableResult = featureTable.updateFeatureAsync(selected);
-
-        // apply update to server when new feature is added, and update the
-        // displayed list of attachments
-        tableResult.addDoneListener(() -> applyEdits(featureTable));
-      });
+      selected.addAttachmentAsync(attachment, "image/png", "edit_feature_attachments/destroyed.png")
```
jenmerritt, line 234:
```suggestion
      // update feature table and apply update to server when new feature is added
      selected.addAttachmentAsync(attachment, "image/png", "edit_feature_attachments/destroyed.png")
```

jenmerritt (commented):
Just 1 very minor doc addition and also please remove the rest of the changes to `read symbols from mobile style file` - we've removed the changes to Controller, but also `build.gradle` and `SymbolLayerInfoListCell` to revert. Then I will be good to approve.

jenmerritt (approved):
Thanks @01smito01 , approving, and I've retargeted at `v.next` now.

colinanderson (approved):
Looks good.

PR #794, opened 2023-07-05
ManageBookmarksSample: bind the ListView to the map's list of Bookmarks

alan-edi:
### Description

Modifies ManageBookmarksSample to bind the ListView to the map's list of Bookmarks. This makes use of the BookmarkList now being an ObservableList.

I've also removed the check on bookmark names being unique. This seemed an unnecessary complication of the sample, especially since doing such a check now the ListView contains Bookmark objects is more complex that before. Not that it can't be done - this is how it would look:
```java
 && bookmarkListView.getItems().stream().noneMatch(b -> b.getName().equals(text)
```

Checklist:
- [x] Request a reviewer
- [x] Assign a reviewer
- [x] Tag reviewer in comment

### DIFF
--- a/map/manage-bookmarks/src/main/java/com/esri/samples/manage_bookmarks/ManageBookmarksSample.java
+++ b/map/manage-bookmarks/src/main/java/com/esri/samples/manage_bookmarks/ManageBookmarksSample.java
@@ -89,8 +89,8 @@ public class ManageBookmarksSample extends Application {
       bookmarkListView.setCellFactory(p -> new BookmarkListCell());
 
       // when user clicks on a bookmark change to that location
-      bookmarkListView.getSelectionModel().selectedItemProperty().addListener((ov, old_val, new_val) -> {
-        mapView.setBookmarkAsync(new_val);
+      bookmarkListView.getSelectionModel().selectedItemProperty().addListener((observable, oldValue, newValue) -> {
+        mapView.setBookmarkAsync(newValue);
       });
 
       // create button to add a bookmark
--- a/map/manage-bookmarks/src/main/java/com/esri/samples/manage_bookmarks/ManageBookmarksSample.java
+++ b/map/manage-bookmarks/src/main/java/com/esri/samples/manage_bookmarks/ManageBookmarksSample.java
@@ -89,9 +89,8 @@ public class ManageBookmarksSample extends Application {
       bookmarkListView.setCellFactory(p -> new BookmarkListCell());
 
       // when user clicks on a bookmark change to that location
-      bookmarkListView.getSelectionModel().selectedItemProperty().addListener((observable, oldValue, newValue) -> {
-        mapView.setBookmarkAsync(newValue);
-      });
+      bookmarkListView.getSelectionModel().selectedItemProperty().addListener(
+          (observable, oldValue, newValue) -> mapView.setBookmarkAsync(newValue));
 
       // create button to add a bookmark
       Button addBookmarkButton = new Button("Add Bookmark");
--- a/map/manage-bookmarks/src/main/java/com/esri/samples/manage_bookmarks/ManageBookmarksSample.java
+++ b/map/manage-bookmarks/src/main/java/com/esri/samples/manage_bookmarks/ManageBookmarksSample.java
@@ -26,6 +26,7 @@ import javafx.scene.control.Alert;
 import javafx.scene.control.Alert.AlertType;
 import javafx.scene.control.Button;
 import javafx.scene.control.Label;
+import javafx.scene.control.ListCell;
 import javafx.scene.control.ListView;
 import javafx.scene.control.TextInputDialog;
 import javafx.scene.layout.Background;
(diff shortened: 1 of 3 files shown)


alan-edi:
Assigning to @jenmerritt for review please. Note it requires `arcgisVersion` to be set to a 200.2.0 build. I have this in my local `build.gradle`:
```
arcgisVersion = '200.2.0-3924'
```
but I assume we wouldn't want to check that in at the moment. What do you normally do?

map/manage-bookmarks/src/main/java/com/esri/samples/manage_bookmarks/ManageBookmarksSample.java
```diff
@@ -82,14 +83,14 @@ public void start(Stage stage) {
...
+      ListView<Bookmark> bookmarkListView = new ListView<>();
+      bookmarkListView.setMaxHeight(190);
+      bookmarkListView.setCellFactory(p -> new BookmarkListCell());
 
       // when user clicks on a bookmark change to that location
-      bookmarkNames.getSelectionModel().selectedItemProperty().addListener((ov, old_val, new_val) -> {
-        int index = bookmarkNames.getSelectionModel().getSelectedIndex();
-        mapView.setBookmarkAsync(bookmarkList.get(index));
+      bookmarkListView.getSelectionModel().selectedItemProperty().addListener((ov, old_val, new_val) -> {
+        mapView.setBookmarkAsync(new_val);
```
jenmerritt:
```suggestion
      bookmarkListView.getSelectionModel().selectedItemProperty().addListener((observable, oldValue, newValue) -> {
        mapView.setBookmarkAsync(newValue);
```
In other listener usages we've used this format for these parameters

jenmerritt (commented):
The changes look good @alan-edi . The only thing that jumps out is an IntelliJ warning on line 136. It doesn't stop the sample working, but wanted to raise it - do you also see this?
intellij-warning
If it's an IntelliJ warning we should ignore, we may need to add a comment or it could be confusing to users.

I agree with the change regarding checking for duplicate bookmarks. I just reviewed the readme and there is a line that says `Enter a unique name for the bookmark`, but I don't think we need to change that particularly, as it's still more useful if you enter a unique name... However I did then spot on line 20 `Add the new bookmark to the book mark list` - could you tidy "book mark" -> "bookmark" while we're here!

Also, you can go ahead and commit the `arcgisVersion = '200.2.0-3924'` line in build.gradle. It's fine in this branch and they all get overwritten/updated at release time to the correct final version when merged into `main`.

alan-edi:
@jenmerritt 

1. I don't get the IntelliJ warning on line 136 that you're seeing. Also I don't understand it, since `BookmarkList` implements `ListenableList<Bookmark>`, and `ListenableList<E>` now extends `ObservableList<E>`.
2. I fixed the "book mark" typo and committed the `arcgisVersion` change.
3. Yes these param names are better, so I've changed them.
4. I see one of the PR checks is failing. It seems to be caused by the `utility_network/perform-valve-isolation-trace` sample failing to build. Any ideas what I should do about that?

jenmerritt (approved):
Approving.

The warning regarding the BookmarkList must just be a linter I'm using in IntelliJ if you don't see it..

I should have added in my previous comment, you can ignore the gradle build Github check in this PR - it will fail for any samples that have been worked on in `v.next` with a daily build because it can't fetch the daily builds. So it's ok.

alan-edi:
Assigning to @01smito01 for second review please

map/manage-bookmarks/src/main/java/com/esri/samples/manage_bookmarks/ManageBookmarksSample.java
```diff
@@ -82,14 +83,14 @@ public void start(Stage stage) {
...
+      // create a ListView to display the bookmarks
+      ListView<Bookmark> bookmarkListView = new ListView<>();
+      bookmarkListView.setMaxHeight(190);
+      bookmarkListView.setCellFactory(p -> new BookmarkListCell());
 
       // when user clicks on a bookmark change to that location
-      bookmarkNames.getSelectionModel().selectedItemProperty().addListener((ov, old_val, new_val) -> {
-        int index = bookmarkNames.getSelectionModel().getSelectedIndex();
-        mapView.setBookmarkAsync(bookmarkList.get(index));
+      bookmarkListView.getSelectionModel().selectedItemProperty().addListener((observable, oldValue, newValue) -> {
```
01smito01: Think we can get rid of the curly brackets here now?

01smito01 (approved):
LGTM @alan-edi, very little to add. FWIW I'm getting the complaint about the cast (or lack of it) at line 136 as well, but I think it's just IntelliJ being silly about where it's sourcing its definitions of each class from. Stepping into `BookmarkList` it's getting `ListenableList` from a JAR for 200.1.0 even though I specified 200.2.0-3931 in the `build.gradle` for manage bookmarks:
image

The sample compiles and runs fine so I'm assuming this is IntelliJ's not dealing with having multiple versions of the runtime properly.

One general comment on the functionality of the sample is that it would be nice if clicking on the currently-selected bookmark recentred the view on that bookmark, as at the moment you need to click on a second bookmark and back onto the first. The version currently in `v.next` exhibits the same behaviour as with these changes, so it's beyond the scope of this PR.

alan-edi:
Thanks @01smito01:
- Good bit of digging re the 200.1.0 JAR.
- I've removed those curly brackets.
- I agree with you re clicking on the currently-selected bookmark, but I think it's best to resist complicating the sample just to achieve that.

PR #797, opened 2023-07-20
Update create and edit geometries sample with scale/rotate/shapes

JonLavi:
Hi @jenmerritt @01smito01 ,

This is the update for the CreateAndEditGeometries sample with the new functionality for the upcoming release.
Apologies if I've missed anything, it's been a while since I did any samples work.

Assigning you both as I'm unsure who needs to review this, please feel free to unassign yourself.

Let me know your thoughts and if there's anything you want improved.


geometry/create-and-edit-geometries/README.md
```diff
@@ -10,7 +10,13 @@ A field worker can mark features of interest on a map using an appropriate geome
...
 
-To create a new geometry, press the button appropriate for the geometry type you want to create (i.e. points, multipoints, polyline, or polygon) and interactively tap and drag on the map view to create the geometry. To edit an existing geometry, tap the geometry to be edited in the map to select it and then edit the geometry by tapping and dragging elements of the geometry. If creating or editing polyline or polygon geometries, choose the desired creation/editing tool (i.e. `VertexTool` or `FreehandTool`).
+To create a new geometry, press the button appropriate for the geometry type you want to create (i.e. points, multipoints, polyline, or polygon) and interactively tap and drag on the map view to create the geometry.
+
+To edit an existing geometry, tap the geometry to be edited in the map to select it and then edit the geometry by tapping and dragging elements of the geometry.
+
+When the whole geometry is selected, you can use the control handles to scale and rotate the geometry.
+
+If creating or editing polyline or polygon geometries, choose the desired creation/editing tool (i.e. `VertexTool`, `FreehandTool`, or one of the available `ShapeTool`s).
```
01smito01: Now that we have six tools available rather than just two, it might be better not to enumerate them and have "choose the desired creation/editing tool from the combo box" or similar.

geometry/create-and-edit-geometries/src/main/java/com/esri/samples/create_and_edit_geometries/CreateAndEditGeometriesController.java
```diff
@@ -101,14 +110,51 @@ public void initialize() {
...
+
+    rectangleShapeTool.getConfiguration().scaleModeProperty().bind(
+            Bindings.when(scaleModeUniformCheckBox.selectedProperty())
+                    .then(GeometryEditorScaleMode.UNIFORM)
+                    .otherwise(GeometryEditorScaleMode.STRETCH));
+
+    triangleShapeTool.getConfiguration().scaleModeProperty().bind(
+            Bindings.when(scaleModeUniformCheckBox.selectedProperty())
+                    .then(GeometryEditorScaleMode.UNIFORM)
+                    .otherwise(GeometryEditorScaleMode.STRETCH));
```
01smito01: Was going to suggest using an iterator to do this, except `getConfiguration` only exists at the subclass level so we have to do something silly like:
```java
    toolComboBox.getItems().forEach(tool -> {
      if (tool instanceof VertexTool){
        ((VertexTool) tool).getConfiguration().scaleModeProperty().bind(
          Bindings.when(scaleModeUniformCheckBox.selectedProperty())
            .then(GeometryEditorScaleMode.UNIFORM)
            .otherwise(GeometryEditorScaleMode.STRETCH));
      } else if (tool instanceof FreehandTool) {
        ((FreehandTool) tool).getConfiguration().scaleModeProperty().bind(
          Bindings.when(scaleModeUniformCheckBox.selectedProperty())
            .then(GeometryEditorScaleMode.UNIFORM)
            .otherwise(GeometryEditorScaleMode.STRETCH));
      } else if (tool instanceof ShapeTool) {
        ((ShapeTool) tool).getConfiguration().scaleModeProperty().bind(
          Bindings.when(scaleModeUniformCheckBox.selectedProperty())
            .then(GeometryEditorScaleMode.UNIFORM)
            .otherwise(GeometryEditorScaleMode.STRETCH));
      }
    });
```

You might prefer that over what we've got at present, or keep as is; I'm not particularly enthused by one over the other (they're both pretty horrid). Java 21's [pattern matching for switch](https://openjdk.org/jeps/441) might make this less bad, but we don't get to use that for another few years.

geometry/create-and-edit-geometries/src/main/java/com/esri/samples/create_and_edit_geometries/CreateAndEditGeometriesController.java
```diff
@@ -101,14 +110,51 @@ public void initialize() {
...
     freehandTool = new FreehandTool();
+    arrowShapeTool = ShapeTool.create(ShapeToolType.ARROW);
+    ellipseShapeTool = ShapeTool.create(ShapeToolType.ELLIPSE);
+    rectangleShapeTool = ShapeTool.create(ShapeToolType.RECTANGLE);
+    triangleShapeTool = ShapeTool.create(ShapeToolType.TRIANGLE);
     toolComboBox.setConverter(new ComboBoxStringConverter());
-    toolComboBox.getItems().addAll(vertexTool, freehandTool);
+    toolComboBox.getItems().addAll(vertexTool, freehandTool, arrowShapeTool, ellipseShapeTool, rectangleShapeTool, triangleShapeTool);
+    // initially configure the geometry editor with the vertex tool
+    geometryEditor.setTool(vertexTool);
```
01smito01, line 123: Vertex tool is selected by default, so this configuration isn't necessary unless you think it's important to be explicit here.

01smito01 (approved):
Looks pretty good @JonLavi, only a few things for your consideration. As this change is going into `v.next` you can update the arcgis version in `build.gradle` to a daily build - at the moment it's set to 200.1.0 which causes issues as we're using methods that don't exist in that release.

(You can ignore the CI checks failing, that just happens on `v.next` as the CI server can't fetch the daily builds itself)

geometry/create-and-edit-geometries/src/main/java/com/esri/samples/create_and_edit_geometries/CreateAndEditGeometriesController.java
JonLavi: This is actually a 'workaround' needed due to how GeometryEditor.start uses a default VertexTool:

When we `geometryEditor.start()`, the GeometryEditor starts with a built-in instance of VertexTool.
In this sample, we [bind the resize mode](https://github.com/Esri/arcgis-maps-sdk-java-samples/pull/797/files#diff-4c91b223790a750050f688a52854c221cb0c82b5fbf6eb1ae91db63ce09f4eb0R128-R131) to the instance of the vertexTool we create [here](https://github.com/Esri/arcgis-maps-sdk-java-samples/pull/797/files#diff-4c91b223790a750050f688a52854c221cb0c82b5fbf6eb1ae91db63ce09f4eb0R114).

This means that if we don't explicitly `setTool(vertexTool)` in the line you highlight above, we'll actually be using a different instance of `VertexTool` (the default one), and the binding we do [here](https://github.com/Esri/arcgis-maps-sdk-java-samples/pull/797/files#diff-4c91b223790a750050f688a52854c221cb0c82b5fbf6eb1ae91db63ce09f4eb0R128-R131) does not seem to take effect, until we switch into a different tool using the [combobox](https://github.com/Esri/arcgis-maps-sdk-java-samples/pull/797/files#diff-4c91b223790a750050f688a52854c221cb0c82b5fbf6eb1ae91db63ce09f4eb0R121) and finally into the vertexTool that is set into the combo box (that has all the bindings).

I hope this makes sense - that's the reason for commit https://github.com/Esri/arcgis-maps-sdk-java-samples/pull/797/commits/a06c7cf425b903c1615b2406cbf17f21a5f597c9.

If you have any other ideas about this one, I'm all ears.

geometry/create-and-edit-geometries/src/main/java/com/esri/samples/create_and_edit_geometries/CreateAndEditGeometriesController.java
JonLavi: Yes, @jenmerritt and I noticed the same annoyance. The fact that we can't call `getConfiguration()` on the base class `GeometryEditorTool` forces us to write a fair bit of clunky code. Your suggestion is slightly shorter, so I'm happy to take that on. 

On the topic of Java21 syntax: let me know if the changes I've made to some of the switch statements [here](https://github.com/Esri/arcgis-maps-sdk-java-samples/pull/797/files#diff-4c91b223790a750050f688a52854c221cb0c82b5fbf6eb1ae91db63ce09f4eb0R421-R424) and [here](https://github.com/Esri/arcgis-maps-sdk-java-samples/pull/797/files#diff-4c91b223790a750050f688a52854c221cb0c82b5fbf6eb1ae91db63ce09f4eb0R478-R482) conform to the current code standard of the ArcGIS Maps SDK for Java samples?

JonLavi:
Thanks @01smito01, I've made a couple of updates as per your suggestions - let me know what you think?

geometry/create-and-edit-geometries/src/main/java/com/esri/samples/create_and_edit_geometries/CreateAndEditGeometriesController.java
01smito01: The switch expressions/enhanced switch are fine - I wanted to use them when creating the sample but couldn't as we were making it available for 200.1.0, which supported Java 11 as well as 17 (switch expressions being a Java 14 feature). We're dropping Java 11 support with 200.2.0 so these are perfectly fine to include and look a lot nicer to boot.

01smito01 (approved):
Happy with that @JonLavi. Could merge now, or hold off a couple weeks and get a review from Jen, or get someone else from the Java team to give it a look over. Up to you, really.

jenmerritt:
@01smito01 @JonLavi Looks like this didn't get a different second reviewer - is the intention this goes in for this release?

JonLavi:
> @01smito01 @JonLavi Looks like this didn't get a different second reviewer - is the intention this goes in for this release?

Hi @jenmerritt , I left this open to give you a chance to review again.
If we can get it in for this release still, then that would be great.
However, we might still add a small change to the data (to have both MapView and GeometryEditor's geometries in the same Spatial Reference). I'll reach out separately

jenmerritt (commented):
Changes look fine to me @JonLavi . I'll await approving until there's confirmation on the data SR.

JonLavi:
@jenmerritt I've pushed a commit to describe the geometries that are loaded at sample startup in WebMercator. Shouldn't notice any difference visually, this is just so that the geometries have the same SpatialReference as the MapView

jenmerritt (approved):
LGTM @JonLavi

PR #799, opened 2023-08-04
Add sample for custom dynamic entity data source

D-R-Smith:
### Description
Add initial implementation of a sample for a custom dynamic entity data source.

Note that because the Java callout does not track an entity item's location that part of the test design is not implemented.

Checklist:
- [x] Request a reviewer
- [x] Assign a reviewer
- [x] Tag reviewer in comment


D-R-Smith:
Assigning to @01smito01 for a first review please.

real_time/add-custom-dynamic-entity-data-source/src/main/java/com/esri/samples/add_custom_dynamic_entity_data_source/AddCustomDynamicEntityDataSourceSample.java
```diff
@@ -0,0 +1,151 @@
...
+  private MapView mapView = new MapView();
+
+  private SimulatedDataSource dynamicEntityDataSource;
+
+  @Override
+  public void start(Stage stage) {
+    StackPane stackPane = new StackPane();
+    Scene scene = new Scene(stackPane);
+
+    // Set title, size, and add scene to stage
```
01smito01: In samples our style is that inline comments use neither capital letters nor full stops. Will need to be changed throughout this file and `SimulatedDataSource`.

real_time/add-custom-dynamic-entity-data-source/src/main/java/com/esri/samples/add_custom_dynamic_entity_data_source/AddCustomDynamicEntityDataSourceSample.java
```diff
@@ -0,0 +1,151 @@
...
+
+import javafx.application.Application;
+import javafx.scene.Scene;
+import javafx.scene.layout.StackPane;
+import javafx.scene.paint.Color;
+import javafx.stage.Stage;
+
+public class AddCustomDynamicEntityDataSourceSample extends Application {
+
+  private MapView mapView = new MapView();
```
01smito01: We're initialising `mapView` within the start method so I think we can leave it declared but uninitialised here. Would be in line with other samples.

real_time/add-custom-dynamic-entity-data-source/src/main/java/com/esri/samples/add_custom_dynamic_entity_data_source/AddCustomDynamicEntityDataSourceSample.java
```diff
@@ -0,0 +1,151 @@
...
+    mapView.setViewpoint(new Viewpoint(new Point(-123.657, 47.984, SpatialReferences.getWgs84()), 3e6));
+
+    // Create a custom data source implementation of a DynamicEntityDataSource.
+    // This takes the path to the simulation file, field name that will be used as the entity id, and the delay between
+    // each observation that is processed.
+    // In this example we are using a json file (stored as a resource) as our custom data source.
+    // This field value should be a unique identifier for each entity.
+    // Adjusting the value for the delay will change the speed at which the entities and their observations are displayed.
+    var resource = getClass().getResource("/add_custom_dynamic_entity_data_source/AIS_MarineCadastre_SelectedVessels_CustomDataSource.json");
+    dynamicEntityDataSource = new SimulatedDataSource(resource.getPath(), "MMSI", 10);
```
01smito01: This is quite a lot of comment - feels like we're repeating the Javadoc for `SimulatedDataSource`'s constructor a bit. Think it could do with being reduced and split up a bit, something like:
```suggestion
    // JSON file (stored as a resource) for use as custom data source
    var resource = getClass().getResource("/add_custom_dynamic_entity_data_source/AIS_MarineCadastre_SelectedVessels_CustomDataSource.json");
    // create a custom data source implementation of a DynamicEntityDataSource with a data source, entity id field name, and update delay
    dynamicEntityDataSource = new SimulatedDataSource(resource.getPath(), "MMSI", 10);
```

I'm not saying to write this specifically but see what you can do.

real_time/add-custom-dynamic-entity-data-source/src/main/java/com/esri/samples/add_custom_dynamic_entity_data_source/AddCustomDynamicEntityDataSourceSample.java
```diff
@@ -0,0 +1,151 @@
...
+    // Add the dynamic entity layer to the map.
+    map.getOperationalLayers().add(dynamicEntityLayer);
+
+    // Add the map view to the stack pane
+    stackPane.getChildren().addAll(mapView);
+  }
+
+  /**
+   *  Set up the track display properties, these properties will be used to configure the appearance of the track line
+   *  and previous observations.
```
01smito01: Javadoc here has an extra space after the `*`

real_time/add-custom-dynamic-entity-data-source/src/main/java/com/esri/samples/add_custom_dynamic_entity_data_source/AddCustomDynamicEntityDataSourceSample.java
01smito01: Also wondering if it's better to have the call to `getPath` on the line assigning `resource`. It's a bit of an ambiguous  `var` there, not one I think we'd have in other places (I could be wrong!)

real_time/add-custom-dynamic-entity-data-source/src/main/java/com/esri/samples/add_custom_dynamic_entity_data_source/SimulatedDataSource.java
```diff
@@ -0,0 +1,168 @@
...
+import com.esri.arcgisruntime.geometry.Point;
+import com.esri.arcgisruntime.geometry.SpatialReferences;
+import com.esri.arcgisruntime.realtime.DynamicEntityDataSource;
+import com.esri.arcgisruntime.realtime.DynamicEntityDataSourceInfo;
+
+/**
+ * A custom DynamicEntityDataSource for processing observations read from a given json file.
+ */
+class SimulatedDataSource extends DynamicEntityDataSource {
+  private final String fileName;
```
01smito01: Should probably have a line break between these.

real_time/add-custom-dynamic-entity-data-source/src/main/java/com/esri/samples/add_custom_dynamic_entity_data_source/SimulatedDataSource.java
```diff
@@ -0,0 +1,168 @@
...
+   */
+  private void processNextObservation() {
+    int lineNumber = currentLineNumber.getAndAdd(1);
+    if (lineNumber >= allJsonLines.size()) {
+      // Finished all the lines
+      observationProcessing.cancel(false);
+    } else {
+      // Process the line by parsing the json, and creating the observation, and adding to the data source.
+      Observation observation = gson.fromJson(allJsonLines.get(lineNumber), Observation.class);
+      Point point = new Point(observation.geometry.x, observation.geometry.y, SpatialReferences.getWgs84());
```
01smito01: Could mebbe use `var` here, but it doesn't matter very much.

real_time/add-custom-dynamic-entity-data-source/src/main/java/com/esri/samples/add_custom_dynamic_entity_data_source/SimulatedDataSource.java
```diff
@@ -0,0 +1,168 @@
...
+  }
+
+  @Override
+  protected CompletableFuture<Void> onConnectAsync() {
+    // Store all lines from file in a list
+    try (BufferedReader bufferedReader = new BufferedReader(new FileReader(fileName))) {
+      String line;
+      while ((line = bufferedReader.readLine()) != null) {
+        allJsonLines.add(line);
+      }
```
01smito01: Nothing wrong with what we have at the moment; would something like this be preferred?
```suggestion
      allJsonLines.addAll(bufferedReader.lines().filter(Objects::nonNull).toList());
```

01smito01 (approved):
This looks pretty good @D-R-Smith, I don't have too many comments to add. Don't have anything to say pro or contra using an `Executor` - not something I'm knowledgeable enough about to have an opinion on.

real_time/add-custom-dynamic-entity-data-source/src/main/java/com/esri/samples/add_custom_dynamic_entity_data_source/AddCustomDynamicEntityDataSourceSample.java
D-R-Smith: Fixed.

real_time/add-custom-dynamic-entity-data-source/src/main/java/com/esri/samples/add_custom_dynamic_entity_data_source/AddCustomDynamicEntityDataSourceSample.java
D-R-Smith: Updated.

real_time/add-custom-dynamic-entity-data-source/src/main/java/com/esri/samples/add_custom_dynamic_entity_data_source/AddCustomDynamicEntityDataSourceSample.java
D-R-Smith: Updated it to be more like what you suggest.

real_time/add-custom-dynamic-entity-data-source/src/main/java/com/esri/samples/add_custom_dynamic_entity_data_source/AddCustomDynamicEntityDataSourceSample.java
D-R-Smith: Fixed

real_time/add-custom-dynamic-entity-data-source/src/main/java/com/esri/samples/add_custom_dynamic_entity_data_source/SimulatedDataSource.java
D-R-Smith: Updated it.

real_time/add-custom-dynamic-entity-data-source/src/main/java/com/esri/samples/add_custom_dynamic_entity_data_source/SimulatedDataSource.java
D-R-Smith: Much neater - I've changed to use your suggestion.

D-R-Smith:
Assigning to @jenmerritt for a second review please. I also plan on getting Colin to review the use of the data source code.

real_time/add-custom-dynamic-entity-data-source/src/main/java/com/esri/samples/add_custom_dynamic_entity_data_source/AddCustomDynamicEntityDataSourceSample.java
```diff
@@ -0,0 +1,149 @@
...
+    stage.setHeight(700);
+    stage.setScene(scene);
+    stage.show();
+
+    // authentication with an API key or named user is required to access basemaps and other location services
+    String yourAPIKey = System.getProperty("apiKey");
+    ArcGISRuntimeEnvironment.setApiKey(yourAPIKey);
+
+    // create a new map with the navigation basemap style
+    ArcGISMap map = new ArcGISMap(BasemapStyle.ARCGIS_OCEANS);
```
jenmerritt: This is as per the design but I'm a bit confused if this should be the navigation basemap here (the design has the navigation basemap in the screengrab, but oceans in implementation). Has there been any chat on this that I've missed? If not I can clarify in the samples channel.

jenmerritt (commented):
Just one initial question. I think it's good but will do another look over tomorrow and likely approve then.

real_time/add-custom-dynamic-entity-data-source/src/main/java/com/esri/samples/add_custom_dynamic_entity_data_source/AddCustomDynamicEntityDataSourceSample.java
D-R-Smith: @jenmerritt I'm not aware of any chat, so it would be good to get this clarified.

jenmerritt: Thanks, will confirm 👍

jenmerritt: (I have asked)

real_time/add-custom-dynamic-entity-data-source/src/main/java/com/esri/samples/add_custom_dynamic_entity_data_source/AddCustomDynamicEntityDataSourceSample.java
```diff
@@ -0,0 +1,149 @@
...
+    stage.setWidth(800);
+    stage.setHeight(700);
+    stage.setScene(scene);
+    stage.show();
+
+    // authentication with an API key or named user is required to access basemaps and other location services
+    String yourAPIKey = System.getProperty("apiKey");
+    ArcGISRuntimeEnvironment.setApiKey(yourAPIKey);
+
+    // create a new map with the navigation basemap style
```
jenmerritt:
```suggestion
    // create a new map with the oceans basemap style
```

jenmerritt: Oceans is correct so we can adjust the comment here

D-R-Smith: Thanks - fixed the comment.

jenmerritt (approved):
LGTM @D-R-Smith

D-R-Smith:
Assigning to @colinanderson for an extra review since it is the first time I've used custom data source.

colinanderson (commented):
Did you reference another implementation for this?

Just from a quick look I don’t think the intention of the `onXxxAsnc()` is to do something and then return a completed future. I think the intention is to return a future that does something and will eventually complete. The who does the async (API or user) question was never resolved to my satisfaction so it is currently up to the user to make it async e.g. use completeAsync(). What you have will work but it isn’t a good example for users - what if connect takes 30 seconds? The UI will probably be blocked.

Here is what the updated test does for a delayed connect
```
return future.completeAsync(() -> {
        try {
               Thread.sleep(10000);
               return null;
             } catch (InterruptedException e) {
               throw new RuntimeException(e);
             }
           }
       );
```
You can add a whenDone or similar to this to start observation processing since that should happen after the connection is made. So you end up with something like
```
return future.completeAsync(() -> {
  // make the connection
}).whenDone(unused, throwable -> {
  // handle exception
  // or start pushing values
});
```
Using `completeAsync` does raise questions about using the fork join pool but there is an overload where you provide your own executor. I've not looked in to the fork join pool stuff - I think it was a question of if it was intended for IO stuff or non-IO stuff.

D-R-Smith:
@colinanderson 
> Did you reference another implementation for this?

Initially I looked at the sample code here: https://devtopia.esri.com/runtime/common-samples/blob/main/designs/AddCustomDynamicEntityDataSource/implementation-details.md

colinanderson:
@D-R-Smith based on .Net I wonder if you should move 
```
    try (BufferedReader bufferedReader = new BufferedReader(new FileReader(fileName))) {
      allJsonLines.addAll(bufferedReader.lines().filter(Objects::nonNull).toList());
    } catch (IOException e) {
      System.out.println(e.getMessage());
    }
```
into the observation processing. Reading all the lines into an array isn't really "connecting" in my mind. If you did that it would be ok to say connect is done immediately so you end up with something like
```
return CompleteableFuture.completedFuture(null).whenDone((u, t) -> {
  // if no exception start pushing observations
});
```

D-R-Smith:
>  Reading all the lines into an array isn't really "connecting" in my mind

This made it easy to have the file closing after the read. Using a static collection of observations is pretty artificial anyway, I felt. I was thinking of having the reading of the file done in the completable future since that hopefully hints to users that they put some code in there for their real life connection.

D-R-Smith:
@colinanderson I've pushed some changes to the connect / completable future code.

real_time/add-custom-dynamic-entity-data-source/src/main/java/com/esri/samples/add_custom_dynamic_entity_data_source/AddCustomDynamicEntityDataSourceSample.java
```diff
@@ -0,0 +1,149 @@
...
+  }
+
+  /**
+   * Stops and releases all resources used in application.
+   */
+  @Override
+  public void stop() {
+    // notify the observations timer thread to stop
+    if (dynamicEntityDataSource != null) {
+      dynamicEntityDataSource.stopObservations();
```
colinanderson: Shouldn't you be disconnecting which would presumably stop the observations?

real_time/add-custom-dynamic-entity-data-source/src/main/java/com/esri/samples/add_custom_dynamic_entity_data_source/SimulatedDataSource.java
```diff
@@ -0,0 +1,180 @@
...
+    this.fileName = fileName;
+    this.entityIdFieldName = entityIdFieldName;
+    this.delay = delay;
+  }
+
+  @Override
+  protected CompletableFuture<Void> onConnectAsync() {
+    return CompletableFuture.completedFuture((Void)null).whenComplete((unused, throwable) -> {
+      if (throwable != null) {
+        System.out.println("Failed to connect custom data source: " + throwable.getMessage());
```
colinanderson: Shouldn't errors got to `System.err`? Also below.

real_time/add-custom-dynamic-entity-data-source/src/main/java/com/esri/samples/add_custom_dynamic_entity_data_source/SimulatedDataSource.java
```diff
@@ -0,0 +1,180 @@
...
+  /**
+   * Read a file of observation data and start processing.
+   */
+  private void startProcessingObservations() {
+    // store all lines from file in a list
+    try (BufferedReader bufferedReader = new BufferedReader(new FileReader(fileName))) {
+      allJsonLines.addAll(bufferedReader.lines().filter(Objects::nonNull).toList());
+    } catch (IOException e) {
+      System.out.println("Failed to start processing observations: " + e.getMessage());
+      throw new CompletionException(e);
```
colinanderson: It is worth checking that this exception actually comes out and is not swallowed in the `whenComplete`. The doc makes it look like you might need to be looking at the stage returned by `whenComplete` to see the exception.
https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/concurrent/CompletableFuture.html#whenComplete(java.util.function.BiConsumer)
>Unlike method [handle](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/concurrent/CompletionStage.html#handle(java.util.function.BiFunction)), this method is not designed to translate completion outcomes, so the supplied action should not throw an exception. However, if it does, the following rules apply: if this stage completed normally but the supplied action throws an exception, then the returned stage completes exceptionally with the supplied action's exception. Or, if this stage completed exceptionally and the supplied action throws an exception, then the returned stage completes exceptionally with this stage's exception.

colinanderson (commented):
@D-R-Smith looks like it should work. Just a few questions.

real_time/add-custom-dynamic-entity-data-source/src/main/java/com/esri/samples/add_custom_dynamic_entity_data_source/AddCustomDynamicEntityDataSourceSample.java
D-R-Smith: That makes sense - I'll remove the extra cancel method and call disconnect.

real_time/add-custom-dynamic-entity-data-source/src/main/java/com/esri/samples/add_custom_dynamic_entity_data_source/SimulatedDataSource.java
D-R-Smith: I'll change them.

real_time/add-custom-dynamic-entity-data-source/src/main/java/com/esri/samples/add_custom_dynamic_entity_data_source/SimulatedDataSource.java
D-R-Smith: I've updated this and I've checked the error causes a retry in DynamicEntityDataSource. Not sure if that is enough.

D-R-Smith:
@colinanderson Thanks for the guidance so far - I've pushed the changes from the latest comments.

real_time/add-custom-dynamic-entity-data-source/src/main/java/com/esri/samples/add_custom_dynamic_entity_data_source/SimulatedDataSource.java
```diff
@@ -0,0 +1,177 @@
...
+    this.fileName = fileName;
+    this.entityIdFieldName = entityIdFieldName;
+    this.delay = delay;
+  }
+
+  @Override
+  protected CompletableFuture<Void> onConnectAsync() {
+    return CompletableFuture.completedFuture((Void)null).whenComplete((unused, throwable) -> {
+      if (throwable != null) {
+        System.err.println("Failed to connect custom data source: " + throwable.getMessage());
```
colinanderson: I think the designed way to handle this would be to look at the connection status and error when `connectAsync` completes - probably use the properties on `DynamicEntityDataSource`.

Here is where you might check if a cancellation exception occurred because that would mean disconnect had been called during connecting and you _might_ need to do some clean up etc. because of that.

colinanderson: I think we should use the `completeAsync` variant here (see https://devtopia.esri.com/runtime/java/blob/main/arcgis-java-testFX/src/test/java/com/esri/arcgisruntime/testfx/realtime/CustomDynamicEntityDataSourceTest.java#L99) so that the loading up of the data is async.

colinanderson (commented):
@D-R-Smith looks ok. Just a couple of comments.

D-R-Smith:
@colinanderson  Updated to use the `DynamicEntityDataSource` properties to check for FAILED and to use `completeAsync`.

real_time/add-custom-dynamic-entity-data-source/src/main/java/com/esri/samples/add_custom_dynamic_entity_data_source/SimulatedDataSource.java
```diff
@@ -0,0 +1,183 @@
...
+  @Override
+  protected CompletableFuture<Void> onConnectAsync() {
+    return new CompletableFuture<Void>().completeAsync(() -> {
+      try {
+        startProcessingObservations();
+        return null;
+      } catch (IOException e) {
+        throw new CompletionException(e);
+      }
+    }).whenComplete((unused, throwable) -> {
```
colinanderson: No, the app itself should be looking at if the connect worked - you don't know if the error etc. has even been set at the end of `onConnectAsync` as that is just part of the whole connect workflow.

I guess the sample design didn't consider this so maybe we should do nothing. The app code is relying on auto loading and connection via the dynamic entity layer. You could listen to the properties somewhere in https://github.com/Esri/arcgis-maps-sdk-java-samples/pull/799/files#diff-8c9721433ff4577e659f28286d5b9ca2d286c1cf1ded8e6732ff25e2a105f026R48 an report if something goes wrong.

colinanderson (commented):
Hopefully clarified!

real_time/add-custom-dynamic-entity-data-source/src/main/java/com/esri/samples/add_custom_dynamic_entity_data_source/SimulatedDataSource.java
D-R-Smith: @colinanderson ahh, yes, of course. I've added a couple of property listeners to hint to users they need to be aware of problems with connections, and removed the duff part on the future.

colinanderson (approved):
Looks ok.

real_time/add-custom-dynamic-entity-data-source/src/main/java/com/esri/samples/add_custom_dynamic_entity_data_source/SimulatedDataSource.java
```diff
@@ -0,0 +1,175 @@
...
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
+
+import com.google.gson.Gson;
+
+import com.esri.arcgisruntime.data.Field;
+import com.esri.arcgisruntime.geometry.Point;
+import com.esri.arcgisruntime.geometry.SpatialReferences;
+import com.esri.arcgisruntime.realtime.ConnectionStatus;
```
jenmerritt, line 39: Just an unused import @D-R-Smith

jenmerritt:
```suggestion
```
Unused import

PR #803, opened 2023-08-21
Mark/java doc download

mbcoder:
### Description

Adding logic for making the JavaDoc work automatically.

Also switching to latest LTS version of JavaFX and using new Gradle version

Checklist:
- [ ] Request a reviewer
- [ ] Assign a reviewer
- [ ] Tag reviewer in comment

### DIFF
--- a/analysis/analyze-hotspots/build.gradle
+++ b/analysis/analyze-hotspots/build.gradle
@@ -90,6 +90,6 @@ task productionZip(type: Zip) {
 if (rootProject == project) {
     // the wrapper task is only used if this project is the root project i.e. not a subproject
     wrapper {
-        gradleVersion = '8.1.1'
+        gradleVersion = '8.3'
     }
 }
--- a/analysis/analyze-hotspots/build.gradle
+++ b/analysis/analyze-hotspots/build.gradle
@@ -1,7 +1,7 @@
 plugins {
     id 'application'
     id 'idea'
-    id 'org.openjfx.javafxplugin' version '0.0.14'
+    id 'org.openjfx.javafxplugin' // version is defined in settings.gradle
 }
 
 idea {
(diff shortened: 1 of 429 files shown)


analysis/analyze-hotspots/build.gradle
```diff
@@ -1,6 +1,13 @@
 plugins {
     id 'application'
-    id 'org.openjfx.javafxplugin' // version is defined in settings.gradle
+    id 'idea'
+    id 'org.openjfx.javafxplugin' version '0.0.14'
```
D-R-Smith: If we are specifying the plugin version here should we not remove  the version from the `settings.gradle` file to avoid future confusion?
Also, does this cause any issue when the samples are included in the sample viewer, which sets a version in its own top-level gradle file and we may need to avoid any conflict with the versions?
Applies to other samples as well.

D-R-Smith (approved):
@mbcoder I'm approving this since the changes look okay to me, although there are two things to consider:
* I've left a comment about whether the javafx plugin version change affects the sample viewer
* I think the `idea` plugin is missing from `feature_layers/time-based-query/build.gradle` file

analysis/analyze-hotspots/build.gradle
steverozic: I thought the purpose of the version in `settings.gradle` was so the version could be stipulated once and removed from `build.gradle` files for each sample in addition the point David makes about the discrepancy. I could be wrong.

mbcoder: @D-R-Smith sample viewer specifies its own plugin version and will be updated too.

The gradle version is set in different places so you can use this project as a massive multi-project, or as individual projects.

mbcoder:
I've added the missing plugin to `feature_layers/time-based-query/build.gradle`

analysis/analyze-hotspots/build.gradle
mbcoder: @D-R-Smith fair point - I've put that back

mfeigl (approved):
@mbcoder great work making these changes!

PR #823, opened 2024-01-22
Implementation of Feature Clustering sample

mfeigl:
### Description
Implementation of the Feature Clustering Sample.

The sample works as outlined in its readme. 

Checklist:
- [x] Request a reviewer
- [x] Assign a reviewer
- [x] Tag reviewer in comment


mfeigl:
@D-R-Smith , could you please review this PR?

feature_layers/add-feature-clustering/src/main/java/com/esri/samples/add_feature_clustering/AddFeatureClusteringSample.java
```diff
@@ -0,0 +1,316 @@
...
+import javafx.scene.layout.CornerRadii;
+import javafx.scene.layout.HBox;
+import javafx.scene.layout.StackPane;
+import javafx.scene.layout.VBox;
+import javafx.scene.paint.Color;
+import javafx.scene.paint.Paint;
+import javafx.scene.text.Font;
+import javafx.stage.Stage;
+
+import com.esri.arcgisruntime.ArcGISRuntimeEnvironment;
```
D-R-Smith: It would maybe be neater if all `com.esri.arcgisruntime` imports were grouped together.

feature_layers/add-feature-clustering/src/main/java/com/esri/samples/add_feature_clustering/AddFeatureClusteringSample.java
```diff
@@ -0,0 +1,316 @@
...
+          // add the map view to the stack pane
+          stackPane.getChildren().addAll(mapView, vBox);
+          StackPane.setAlignment(vBox, Pos.TOP_LEFT);
+          StackPane.setMargin(vBox, new Insets(10, 0, 0, 10));
+
+          // Add a class break for each intended value range and define a symbol to display for features in that range.
+          // In this case, the average building height ranges from 0 to 8 stories.
+          // For each cluster of features with a given average building height, a symbol is defined with a specified
+          // color.
+          List<ClassBreaksRenderer.ClassBreak> classBreaks = new ArrayList<ClassBreaksRenderer.ClassBreak>();
```
D-R-Smith: Can use `new ArrayList<>();` since the type argument can be inferred.

feature_layers/add-feature-clustering/src/main/java/com/esri/samples/add_feature_clustering/AddFeatureClusteringSample.java
```diff
@@ -0,0 +1,316 @@
...
+  private ArcGISMap mMap;
+  private CheckBox mDisplayLabelsCheckbox;
+  private Label mClusterRadiusLabel;
+  private Slider mClusterRadiusSlider;
+  private Label mMaxScaleLabel;
+  private Slider mMaxScaleSlider;
+  private Label mCurrentMapScaleLabel;
+  private FeatureLayer mLayer;
+
+  ClusteringFeatureReduction mClusteringFeatureReduction;
```
D-R-Smith: This could be `private`.

feature_layers/add-feature-clustering/src/main/java/com/esri/samples/add_feature_clustering/AddFeatureClusteringSample.java
```diff
@@ -0,0 +1,316 @@
...
+import javafx.scene.text.Font;
+import javafx.stage.Stage;
+
+import com.esri.arcgisruntime.ArcGISRuntimeEnvironment;
+import com.esri.arcgisruntime.mapping.ArcGISMap;
+import com.esri.arcgisruntime.mapping.view.MapView;
+
+public class AddFeatureClusteringSample extends Application {
+
+  private ArcGISMap mMap;
```
D-R-Smith: I think the convention in samples is not to use the `m` prefix so I'd suggest removing it from all the members.

feature_layers/add-feature-clustering/src/main/java/com/esri/samples/add_feature_clustering/AddFeatureClusteringSample.java
```diff
@@ -0,0 +1,316 @@
...
+    mClusterRadiusLabel.setTextFill(Color.WHITE);
+    mClusterRadiusSlider = new Slider();
+    mClusterRadiusSlider.setValue(DEFAULT_CLUSTER_R);
+
+    // create a slider for adjusting cluster radius
+    mClusterRadiusSlider = new Slider(30, 85, DEFAULT_CLUSTER_R);
+    mClusterRadiusSlider.setMaxWidth(MAX_WIDTH);
+    // add a listener to the slider's value property to set the cluster radius
+    mClusterRadiusSlider.setOnMouseReleased(event -> {
+      mClusterRadiusLabel.setText(String.valueOf("Cluster radius: " + (int) mClusterRadiusSlider.getValue()));
```
D-R-Smith: I think the `String.valueOf` is unnecessary. Similarly on line 269.

feature_layers/add-feature-clustering/src/main/java/com/esri/samples/add_feature_clustering/AddFeatureClusteringSample.java
```diff
@@ -0,0 +1,316 @@
...
+  private Slider mMaxScaleSlider;
+  private Label mCurrentMapScaleLabel;
+  private FeatureLayer mLayer;
+
+  ClusteringFeatureReduction mClusteringFeatureReduction;
+
+  private MapView mapView;
+  private PortalItem portalItem; // keep loadable in scope to avoid garbage collection
+
+  private final int DEFAULT_CLUSTER_R = 60;
```
D-R-Smith: I'd consider spelling out `RADIUS` in full in the name - we don't seem to abbreviate it elsewhere like this.

feature_layers/add-feature-clustering/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+![Image of display map](AddFeatureClustering.png)
+
+## Use case
+
+Feature clustering can be used to dynamically aggregate groups of points that are within proximity of each other in order to represent each group with a single symbol. Such grouping allows you to see patterns in the data that are difficult to visualize when a layer contains hundreds or thousands of points that overlap and cover each other. Users can add feature clustering to point feature layers. This is useful when the layer does not have the feature reduction defined or when the existing feature reduction properties need to be overridden.
+
+## How to use the sample
+
+Interact with the controls to customize clustering feature reduction properties. Tap on any clustered aggregate geoelement to see the cluster feature count and aggregate fields in the popup.
```
D-R-Smith: > Tap on any clustered aggregate geoelement to see the cluster feature count

I didn't see any popups appearing when I clicked on things.

feature_layers/add-feature-clustering/src/main/java/com/esri/samples/add_feature_clustering/AddFeatureClusteringSample.java
```diff
@@ -0,0 +1,316 @@
...
+    final int MIN_WIDTH = 100;
+    final int MAX_WIDTH = 300;
+    final int SPACING = 6;
+
+    HBox displayHBox = new HBox(SPACING);
+    Label label1 = new Label("Display Labels");
+    label1.setMinWidth(MIN_WIDTH);
+    label1.setTextFill(Color.WHITE);
+    // create checkbox to toggle the showing of cluster count
+    mDisplayLabelsCheckbox = new CheckBox();
```
D-R-Smith: The samples are not completely consistent but a quick search suggests the more common way is to put the text (_"Display Labels"_) in the `CheckBox` constructor which takes a `String` rather than a separate label, but there may be a reason to do it like this.

feature_layers/add-feature-clustering/src/main/java/com/esri/samples/add_feature_clustering/AddFeatureClusteringSample.java
```diff
@@ -0,0 +1,316 @@
...
+    mDisplayLabelsCheckbox.setSelected(false);
+    mDisplayLabelsCheckbox.setTextFill(Color.WHITE);
+    mDisplayLabelsCheckbox.selectedProperty().addListener(o -> {
+      if (mDisplayLabelsCheckbox.isSelected()) {
+        SimpleLabelExpression simpleLabelExpression = new SimpleLabelExpression("[cluster_count]");
```
D-R-Smith: I wonder if it would be neater to just construct the `labelDefinition` once outside of the handler and then the handler would look simpler as it would just be adding or removing the labelDefinition?

feature_layers/add-feature-clustering/src/main/java/com/esri/samples/add_feature_clustering/AddFeatureClusteringSample.java
```diff
@@ -0,0 +1,316 @@
...
+    mClusterRadiusLabel = new Label("Cluster radius: " + DEFAULT_CLUSTER_R);
+    mClusterRadiusLabel.setMinWidth(MIN_WIDTH);
+    mClusterRadiusLabel.setTextFill(Color.WHITE);
+    mClusterRadiusSlider = new Slider();
+    mClusterRadiusSlider.setValue(DEFAULT_CLUSTER_R);
+
+    // create a slider for adjusting cluster radius
+    mClusterRadiusSlider = new Slider(30, 85, DEFAULT_CLUSTER_R);
+    mClusterRadiusSlider.setMaxWidth(MAX_WIDTH);
+    // add a listener to the slider's value property to set the cluster radius
```
D-R-Smith: I think this used to listen to the property but that was what caused the flicker - the comment needs updated (similarly for the comment for the max scale).

feature_layers/add-feature-clustering/src/main/java/com/esri/samples/add_feature_clustering/AddFeatureClusteringSample.java
```diff
@@ -0,0 +1,316 @@
...
+      mClusteringFeatureReduction.setMaxScale(mMaxScaleSlider.getValue());
+    });
+    maxScaleHBox.getChildren().addAll(mMaxScaleLabel, mMaxScaleSlider);
+
+    // show current map scale in a label within the control panel
+    mCurrentMapScaleLabel = new Label("Scale: 1:" + mapView.getMapScale());
+    mCurrentMapScaleLabel.setTextFill(Color.WHITE);
+
+    // listen for map scale changes and update the label
+    mapView.mapScaleProperty().addListener((observable, oldValue, newValue) ->
```
D-R-Smith: The change listener could be done as a binding like this which seems neater to me but is subjective:

```java
 mCurrentMapScaleLabel.textProperty().bind(mapView.mapScaleProperty().map(i -> "Scale: 1:" + Math.round(i.doubleValue())));
```

feature_layers/add-feature-clustering/src/main/java/com/esri/samples/add_feature_clustering/AddFeatureClusteringSample.java
```diff
@@ -0,0 +1,316 @@
...
+      mapView.setMap(mMap);
+
+    } catch (Exception e) {
+      // on any error, display the stack trace.
+      e.printStackTrace();
+    }
+  }
+
+  /**
+   * Creates a UI with three buttons and a label.
```
D-R-Smith: Comment appears incorrect - the UI has mostly sliders not buttons.

feature_layers/add-feature-clustering/src/main/java/com/esri/samples/add_feature_clustering/AddFeatureClusteringSample.java
```diff
@@ -0,0 +1,316 @@
...
+          mLayer.setFeatureReduction(mClusteringFeatureReduction);
+
+          // Set initial slider values.
+          // Note that the default value for cluster radius is 60.
+          // Increasing the cluster radius increases the number of features that are grouped together into a cluster.
+          mClusteringFeatureReduction.setRadius(DEFAULT_CLUSTER_R);
+
+          // Note that the default value for max scale is 0.
+          // The max scale value is the maximum scale at which clustering is applied.
+          mClusteringFeatureReduction.setMaxScale(0.0);
```
D-R-Smith: If the default is zero, is there a benefit in setting it?

feature_layers/add-feature-clustering/src/main/java/com/esri/samples/add_feature_clustering/AddFeatureClusteringSample.java
```diff
@@ -0,0 +1,316 @@
...
+
+      mMap = new ArcGISMap(portalItem);
+
+      // Get the Zurich buildings feature layer once the map has finished loading
+      mMap.addDoneLoadingListener(() -> {
+        if (mMap.getLoadStatus() == LoadStatus.LOADED) {
+          LayerList l = mMap.getOperationalLayers();
+          if (l.isEmpty())
+            return;
+          mLayer = (FeatureLayer) mMap.getOperationalLayers().get(0);
```
D-R-Smith: Could use `l` instead of calling `mMap.getOperationalLayers()` again, and it might be worth giving `l` a more descriptive name.

D-R-Smith (approved):
@mfeigl I've made a few comments for consideration, but looks okay to me.

feature_layers/add-feature-clustering/README.md
mfeigl: Thank you for noticing, I missed them - added now.

feature_layers/add-feature-clustering/src/main/java/com/esri/samples/add_feature_clustering/AddFeatureClusteringSample.java
mfeigl: Yes I do agree, that's what I prefer as well - in this case, however, the UI would become inconsistent, as the text is on the right side of the checkbox. Therefore, I'd rather leave it as is.

feature_layers/add-feature-clustering/src/main/java/com/esri/samples/add_feature_clustering/AddFeatureClusteringSample.java
mfeigl: Yes, I do like it - thank you for the suggestion!

feature_layers/add-feature-clustering/src/main/java/com/esri/samples/add_feature_clustering/AddFeatureClusteringSample.java
mfeigl: You are right, but I shall leave it so to showcase this functionality as well.

feature_layers/add-feature-clustering/src/main/java/com/esri/samples/add_feature_clustering/AddFeatureClusteringSample.java
mfeigl: Yes, very true!

mfeigl:
@D-R-Smith , could you do another review, please?

feature_layers/add-feature-clustering/src/main/java/com/esri/samples/add_feature_clustering/AddFeatureClusteringSample.java
```diff
@@ -101,29 +110,69 @@ public void start(Stage stage) {
...
+                popupContent = "";
+                IdentifyLayerResult layer = identifiedLayerResults.get();
+
+                for (Popup popup : layer.getPopups()) {
+                  final Map<String, Object> attributes = popup.getGeoElement().getAttributes();
+                  for (final String name : attributes.keySet()) {
+                    popupContent += name + ": " + attributes.get(name).toString() + "\n";
+                  }
+                }
+                System.out.println("###: " + popupContent);
```
D-R-Smith: Can remove the print.

feature_layers/add-feature-clustering/src/main/java/com/esri/samples/add_feature_clustering/AddFeatureClusteringSample.java
```diff
@@ -56,27 +66,26 @@
...
-  private Slider mMaxScaleSlider;
-  private Label mCurrentMapScaleLabel;
-  private FeatureLayer mLayer;
-
-  ClusteringFeatureReduction mClusteringFeatureReduction;
-
+  private ArcGISMap map;
+  private CheckBox displayLabelsCheckbox;
+  private Label clusterRadiusLabel;
+  private ComboBox clusterRadiusPicker;
```
D-R-Smith: IntelliJ highlighted that these were unparameterized uses of ComboBox, and could be changed to `ComboBox<Integer>`. I think this would get rid of the warning about unchecked or unsafe  operations as well, since the compiler can then check the values match the ComboBox expectations.

You would then initialise them later with the `<>` syntax, like this
```java
clusterRadiusPicker = new ComboBox<>();
```

feature_layers/add-feature-clustering/src/main/java/com/esri/samples/add_feature_clustering/AddFeatureClusteringSample.java
```diff
@@ -101,29 +110,69 @@ public void start(Stage stage) {
...
+            ListenableFuture<IdentifyLayerResult> identifiedLayerResults = mapView.identifyLayerAsync(layer, point, 3.0, true);
+            identifiedLayerResults.addDoneListener(() -> {
+              try {
+                // clear the list of popup content
+                popupContent = "";
+                IdentifyLayerResult layer = identifiedLayerResults.get();
+
+                for (Popup popup : layer.getPopups()) {
+                  final Map<String, Object> attributes = popup.getGeoElement().getAttributes();
+                  for (final String name : attributes.keySet()) {
```
D-R-Smith: I feel using a stream would be the more usual way of collecting the strings these days, but you may feel it is better with the explicit loop. I'm thinking along the lines of

```java
popupContent = attributes
                      .entrySet()
                      .stream()
                      .map(a -> a.getKey() + ": " + a.getValue())
                      .collect(Collectors.joining(System.lineSeparator()));
```

feature_layers/add-feature-clustering/src/main/java/com/esri/samples/add_feature_clustering/AddFeatureClusteringSample.java
```diff
@@ -101,29 +110,69 @@ public void start(Stage stage) {
...
+                }
+                System.out.println("###: " + popupContent);
+                Platform.runLater(() ->{
+                  popupContentLabel.setText(popupContent);
+                  popupContentLabel.setTextFill(Color.WHITE);
+                  vBoxPopupInformation.setVisible(!popupContentLabel.getText().isEmpty());
+                });
+
+              } catch (ExecutionException | InterruptedException e) {
+                throw new RuntimeException(e);
```
D-R-Smith: Since this is inside a handler it might just get swallowed up by the calling thread, so maybe a `System.err.println` of the `getMessage` would be better.

feature_layers/add-feature-clustering/src/main/java/com/esri/samples/add_feature_clustering/AddFeatureClusteringSample.java
```diff
@@ -200,9 +249,22 @@ public void start(Stage stage) {
   }
 
   /**
-   * Creates a UI with three buttons and a label.
+   * Gets a VBox and sets up its styling.
```
D-R-Smith: I'd consider rewording more like `Sets styling for the given VBox.`

feature_layers/add-feature-clustering/src/main/java/com/esri/samples/add_feature_clustering/AddFeatureClusteringSample.java
```diff
@@ -211,83 +273,71 @@ private VBox controlsVBox() {
...
+    maxScaleLabel.setMinWidth(MIN_WIDTH);
+    maxScaleLabel.setTextFill(Color.WHITE);
+    // create a ComboBox for adjusting max scale
+    maxScalePicker = new ComboBox();
+    maxScalePicker.getItems().addAll(0, 1000, 5000, 10000, 50000, 100000, 500000);
+    maxScalePicker.setValue(0);
+
+    // add a listener to the ComboBox's value property to set the max scale
+    maxScalePicker.setOnAction(event -> {
+      clusteringFeatureReduction.setMaxScale((int) maxScalePicker.getValue());
```
D-R-Smith: If you parameterise the declaration of the ComboBox you can remove the cast to `int`.

D-R-Smith (approved):
@mfeigl A few comments to consider but looks okay

feature_layers/add-feature-clustering/src/main/java/com/esri/samples/add_feature_clustering/AddFeatureClusteringSample.java
mfeigl: Sounds great, thank you!

feature_layers/add-feature-clustering/src/main/java/com/esri/samples/add_feature_clustering/AddFeatureClusteringSample.java
mfeigl: Yes absolutely!

mfeigl:
Assigning to @mbcoder for second review.

feature_layers/add-feature-clustering/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+## Relevant API
+
+* AggregateGeoElement
+* ClassBreaksRenderer
+* FeatureLayer
+* FeatureReduction
+* GeoElement
+* IdentifyLayerResult
+* PopupViewer
```
mbcoder: Why is this relevant API?

mbcoder (changes requested):
Mostly looks good, but I had one question about the PopupViewer

feature_layers/add-feature-clustering/README.md
mfeigl: I accidentally left it in, thank you for noticing - removed it now

mbcoder (approved):
Looks good now,

